PR #171, opened 2014-05-12
Plbvh raycast2

rtrius:
The increase in memory consumption from splitting the m_raytracePairsKernel (8 -> 56 bytes per pair)
is rather high. Since the rigid body index is already known for each ray-rigid pair and it is possible to interpolate the ray's m_hitPoint from its m_hitFraction, we can just store the normal and put the
m_hitFraction in the normal's w component. This would reduce the memory usage to 24 bytes per pair.

Does this sound like a good idea? If so, I'll make another commit to implement it.


erwincoumans:
There is no real active development in the OpenCL acceleration and due to refactoring of the examples (old demos) the pull request is in conflict. Unless you have time to make it work with latest trunk, lets close the pull request. Thanks a lot for the contributions anyhow!

PR #234, opened 2014-09-04
memset() header added

r-lyeh:
File header was missing :)
Failed to compile if using my own build scripts in debug mode.

Keep up the good work
- r-lyeh


erwincoumans:
the memset is commented out, so we can close the issue

PR #236, opened 2014-09-04
Replaced the collision states defines

ljyanesm:
ACTIVE_TAG = 1,
ISLAND_SLEEPING = 2,
WANTS_DEACTIVATION = 3,
DISABLE_DEACTIVATION = 4,
DISABLE_SIMULATION = 5

With an enum, thus validating at compile time that

setActivationState(), forceActivationState() calls always receive valid arguments.
Also, changed the demo application and btRigidBody to reflect the change.


erwincoumans:
Need to figure out a way to re-trigger travis on this pull request

erwincoumans:
At this stage I prefer not to break the API and people having to fix their code bases when they upgrade to a newer version of Bullet 2.x. I'll make sure to fix it in Bullet 3.x.

PR #253, opened 2014-10-02
Install pkg-config file on MinGW systems as well

TimothyGu:


scpeters:
+1

erwincoumans:
Sure, it seems harmless change. Thanks for the contribution!

PR #259, opened 2014-10-07
Fix for use of uninitialized variables.

AlexanderPolyakov:
m_implicitShapeDimensions was used in setMargin, called from setSafeMargin
from constructors of btBoxShape, btCylinerShape and btBox2dShape,
effectively using uninitialized variable and leading to floating point
exceptions.
It was working correctly only because in the same constructor
m_implicitShapeDimensions was reinitialized after setSafeMargin.
Properly initialize variable before use by first setting it up and only
then using it in setSafeMargin.


erwincoumans:
Somehow I was processing only the most recent pull requests, and older ones got snowed under. Let's apply the patch.

PR #301, opened 2014-12-21
add DebugDraw for shapes that are wrapped with a btUniformScalingShape

VirusFree:


erwincoumans:
Why is the issue closed? I haven't had time yet to look at it.

VirusFree:
Actually, I changed the physics engine i use to one that is more .NET (c#) friendly and i deleted the clone on my pc and was going(maybe) to delete my github fork too.
Also i would not have been able to test/change stuff anymore so i thought i should close this pull request.
But I could leave it as it ,or reopen it if you like :)

PR #302, opened 2014-12-22
Solves a source of warnings that come up when using Clang

psi29a:
This will not only speed up compilation by removing a source of warnings that spam your console, but makes it easier to read see what is going on.

### DIFF
--- a/src/Bullet3Common/b3Vector3.h
+++ b/src/Bullet3Common/b3Vector3.h
@@ -1077,25 +1077,13 @@ public:
 	{
 		int maxIndex = -1;
 		b3Scalar maxVal = b3Scalar(-B3_LARGE_FLOAT);
-		if (m_floats[0] > maxVal)
+		for (unsigned char i = 0; i < 4; i++)
 		{
-			maxIndex = 0;
-			maxVal = m_floats[0];
-		}
-		if (m_floats[1] > maxVal)
-		{
-			maxIndex = 1;
-			maxVal = m_floats[1];
-		}
-		if (m_floats[2] > maxVal)
-		{
-			maxIndex = 2;
-			maxVal =m_floats[2];
-		}
-		if (m_floats[3] > maxVal)
-		{
-			maxIndex = 3;
-			maxVal = m_floats[3];
+			if (m_floats[i] > maxVal)
+			{
+				maxIndex = i;
+				maxVal = m_floats[i];
+			}
 		}
 
 		return maxIndex;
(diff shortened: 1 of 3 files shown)


psi29a:
Added an additional patch that fixes the same problem, but also in the bullet3 side of things.

src/Bullet3Common/b3Vector3.h
```diff
@@ -1077,25 +1077,13 @@ class b3Vector4 : public b3Vector3
 	{
 		int maxIndex = -1;
 		b3Scalar maxVal = b3Scalar(-B3_LARGE_FLOAT);
-		if (m_floats[0] > maxVal)
+		for (unsigned char i = 0; i++; i < 4)
```
scpeters: Shouldn't the `i++` and `i < 4` be reversed?

scpeters:
It looks like the loops were manually unrolled. I'm not sure if there are performance implications of this change.

psi29a:
Oei! Thanks for catching that.

Manually unrolling is unnecessary in most cases when you can pass -funroll-loops in both gcc and clang and is also done in msvc when producing optimized builds.

erwincoumans:
Processing / cleaning up some old pull requests. What were those clang warnings exactly?

psi29a:
This happens quite a bit, still does. :)

> In file included from /home/bcurtis/Workspace/Private/Bullet/bullet3/src/LinearMath/btConvexHull.cpp:18:
> In file included from /home/bcurtis/Workspace/Private/Bullet/bullet3/src/LinearMath/btConvexHull.h:22:
> /home/bcurtis/Workspace/Private/Bullet/bullet3/src/LinearMath/btVector3.h:1150:4: warning: Value stored to 'maxVal' is never read
>                         maxVal = m_floats[3];
>                         ^        ~~~~~~~~~~~
> /home/bcurtis/Workspace/Private/Bullet/bullet3/src/LinearMath/btVector3.h:1179:4: warning: Value stored to 'minVal' is never read
>                         minVal = m_floats[3];
>                         ^        ~~~~~~~~~~~
> In file included from /home/bcurtis/Workspace/Private/Bullet/bullet3/src/LinearMath/btConvexHull.cpp:18:
> In file included from /home/bcurtis/Workspace/Private/Bullet/bullet3/src/LinearMath/btConvexHull.h:23:
> /home/bcurtis/Workspace/Private/Bullet/bullet3/src/LinearMath/btAlignedObjectArray.h:179:4: warning: Returning null reference
>                         return m_data[n];
>                         ^~~~~~~~~~~~~~~~
> /home/bcurtis/Workspace/Private/Bullet/bullet3/src/LinearMath/btConvexHull.cpp:476:10: warning: Access to field 'rise' results in a dereference of a null pointer (loaded from variable 't')
>         return (t->rise >epsilon)?t:NULL ;
>                 ^~~~~~~

psi29a:
They are annoying because the files are used everywhere so it is just conscientiously spammed in build logs from bullet itself and to applications that depends on the headers.

psi29a:
Just did a rebase to catch up to current master, no conflicts. :)

erwincoumans:
I fixed the warning manually, without a loop, thanks for the report/pull request!

psi29a:
Super, thank you. ;)

Issue #258, opened 2014-10-05
btCollisionWorld::rayTest - thread safe?


varoudis:
Hello,

Is rayTest thread safe? 
Can you do a lot parallel rayTests on different thread?
During the tests NO updates/modifications will happen to the world and NO world step! (if that make a difference)

Thanks

erwincoumans:
Yes, you can do lots of parallel ray tests in different threads (PyBullet does this)

Issue #254, opened 2014-10-02
avoid memory allocation during rayTest


erwincoumans:
See also http://bulletphysics.org/Bullet/phpBB3/viewtopic.php?f=9&t=10142

erwincoumans:
closing out-of-date issues.

Issue #126, opened 2014-03-30
Implement CPU multithreaded version (pthreads/TBB/OpenCL/MPI etc)


erwincoumans:
Right now Bullet 2 is single threaded and we discarded the obsolete BulletMultiThreaded. Bullet 3 OpenCL is mainly designed for GPU and OpenCL drivers are unreliable. It would be good to create a new multithreaded version of Bullet keeping in mind Bullet 2 and Bullet 3 OpenCL.

erwincoumans:
This was contributed now, CPU multithreading can be enabled. See Bullet/examples/BenchmarkDemo for an example. Thanks @lunkhound !

georgwuensch:
Dear Erwin, Dear Lunkhound, great that this is finally integrated. Thank you!

One question from your experience, what differences do you see between using openMP, ThreadSupport and TBB? Will there be any differences? 

I am working on a little logger to do more testing on different laptops and workstations with the different options. 

Maybe I can share some results on this if you like.

Yours Georg.

georgwuensch:
HI, While looking into it I found (I am still compiling with Studio 2010...) that it uses concurrency:: as namespace for the ms ppl, while it only compiles with Concurrency:: AM I missing a settings there?

erwincoumans:
You may have to fix things to get it the various multi-threading working.

lunkhound:
Hi,

The PPL option should compile for MSVC 2013 and 2015 as I recall. I never tried it with VC 2010, so it sounds like it only works with newer versions of PPL. I don't really recommend using PPL. Its mainly there as an example of how to implement a task scheduler for Bullet. Performance wasn't great compared to the other options.

The ThreadSupport option is probably the best choice for most cases. It's built-in (to Bullet) so there's no extra setup, and it has the best performance (on the Windows platform at least, according to my testing).

I found IntelTBB to also have good performance, but it's a bit of a hassle to setup since it's an external library.

OpenMP on windows with MSVC works, and is very easy to setup since it is built into the compiler but doesn't perform that great. With other compilers hopefully the performance is better.

Please feel free to share your results with CPU multithreading on this [Forum thread](https://pybullet.org/Bullet/phpBB3/viewtopic.php?f=9&t=10232)

PR #303, opened 2014-12-23
threading for Bullet2

lunkhound:
MultiThreadedDemo uses either TBB or OpenMP (issue #126)
Also makes Dvbt rayTest threadsafe (issue #254, #258)
Discussion (here)[http://www.bulletphysics.org/Bullet/phpBB3/viewtopic.php?f=9&t=10232]

Areas that run in parallel:
 - narrowphase (dispatchAllPairs)
 - discreteDynamicsWorld solve islands (each island is a separate task)
 - discreteDynamicsWorld predictUnconstraintMotion
 - discreteDynamicsWorld createPredictiveContacts
 - discreteDynamicsWorld integrateTransforms

There were some significant changes to how the btSimulationIslandManager works, so that islands could be dispatched in parallel to different solvers.


erwincoumans:
Thanks for sharing your multithreading code! 
I'll give some feedback once I get the time to review it properly.

lunkhound:
Thanks!  I welcome any feedback or questions.

erwincoumans:
I cannot accept the patch right now, because I am focussing on Bullet 3.x and rethinking the whole API and such. I'll take a look at your patch anyway, and see what will be useful for future refactoring.
I'll leave the pull request open anyway, hopefully someone can use the patch.

lunkhound:
I am currently working on rebasing the work in this patch to the current version of bullet. I've collapsed it from 50+ commits to about 12 or so. I'm stripping out various dead-end lines of development and failed experiments, so the resulting patch should be much easier to follow. All the changes to the bullet core are already applied, all that remains is to port the demo to the new example browser framework. When it is ready I'll create a new pull request.

lunkhound:
Closing this in favor of the newer PR.

PR #305, opened 2014-12-28
Overcome the 'GetTickCount64' issue on XP using C++11 chrono library.

iSLC:
For those in need of such option. I haven't tested the implementation (*although it should be fairly straight forward*). The only thing I've tested was to see if it compiles. The used compiler was [MinGW-w64 4.9.2 x32](http://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/4.9.2/threads-posix/dwarf/).


erwincoumans:
This issue is fixed in the latest repo, by this commit:
https://github.com/bulletphysics/bullet3/commit/a159fbac698520b7fcca585d6129eaa79bcdd69b#diff-bbc3160e91a1636299055c622964e0f1
Thanks for the report!

PR #307, opened 2015-01-02
Fix the shadows in the 'Gyroscopic' demo

a0121536:
Hello, it's just a small thing to straighten up the shadows in the Gyro demo.
- Svetlin Nikolov


erwincoumans:
thanks for the contribution, all bits help big or small :+1:

PR #313, opened 2015-01-21
Fixed self-referencial bug in btVector3's setInterpolate3 (NEON)

jackoalan:
When compiling for ARM NEON, the intrinsics in `btVector3::setInterpolate3` may encounter a situation where `v0` is a self-reference (and setting `mVec128` multiple times corrupts its own state). Multiple such situations occur in `btKinematicCharacterController.cpp`. SSE and Scalar variants are immune.

### DIFF
--- a/src/Bullet3Common/b3Vector3.h
+++ b/src/Bullet3Common/b3Vector3.h
@@ -467,10 +467,10 @@ public:
 		__m128 tmp3 = _mm_add_ps(r0,r1);
 		mVec128 = tmp3;
 #elif defined(B3_USE_NEON)
-		mVec128 = vsubq_f32(v1.mVec128, v0.mVec128);
-		mVec128 = vmulq_n_f32(mVec128, rt);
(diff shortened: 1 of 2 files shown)


erwincoumans:
thanks for the report and fix. Did you run the iOS tests by any chance (under tests/Bullet2) ? I will have to setup my iOS build environment to double-check. Thanks!

jackoalan:
I did not build the tests (I'm ignorant of the premake build system)

However, I'm debugging against a full application (with an SSE desktop target too) and was wondering why the kinematic character controller was flinging my character out of the map on NEON with the same code. 

This edit fixed the character controller behavior like the SSE build

sinkingsugar:
I can confirm that the issue is there (under iOS, tested). In my case I just disabled neon code because my team was in a rush for release and also because I was actually using 2.82 r2704. I will soon migrate to this github version tho and would be nice if it was fixed.

erwincoumans:
ok, I confirmed the issue and that your patch fixes it, thanks for the contribution!

PR #315, opened 2015-01-23
Add option to create btTriangleMesh with vertices + indices, not just triangles.

quiasmo:
Des no harm, could come in handy for someone else.


erwincoumans:
The 'Gwen' OpenGL target_link_libraries is only needed when building shared libs, this is fixed in latest trunk, so I need to skip that part. I wonder if we can 'cherry-pick' the addTriangle commit ( 0d4adc6) using the web interface.

erwincoumans:
The btTriangleMesh::addTriangleIndices call is added
https://github.com/bulletphysics/bullet3/commit/c7671779c238d9761c9d14de25898d60f44ff3fa
Thanks for the contribution!

PR #317, opened 2015-02-01
Fix for 64bit builds btSqrt and USE_APPROXIMATION

namar0x0309:
Fix issue with btSqrt with macro USE_APPROXIMATION turned on.

In 64bit systems, type 'long' is 8 bytes and not 4 which the btSqrt USE_APPROXIMATION code relies on to get to upper 4 bytes. Changed algorithm for arm64 to something more stable. 

Source of implementation: http://www.hackersdelight.org/hdcodetxt/rsqrt.c.txt
Note, author is unknown.


erwincoumans:
Thanks for the contribution. Where did you get the double-precision number from? http://www.azillionmonkeys.com/qed/sqroot.html ?

namar0x0309:
Xcode forces 64bit now, that's how the function 'btSqrt' stopped working. I'm expecting this to hit more people that use the 'USE_APPROXIMATION' macro.

PR #319, opened 2015-02-04
Use absolute paths in pkg-config file

scpeters:
Otherwise compilation can fail depending on where the build folder is located.

### DIFF
--- a/bullet.pc.cmake
+++ b/bullet.pc.cmake
@@ -2,5 +2,5 @@ Name: bullet
 Description: Bullet Continuous Collision Detection and Physics Library
 Requires:
 Version: @BULLET_VERSION@
-Libs: -L@LIB_DESTINATION@ -lBulletSoftBody -lBulletDynamics -lBulletCollision -lLinearMath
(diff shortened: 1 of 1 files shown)


erwincoumans:
I never use system-wide install myself, but others do. I just hope this change doesn't break things for others. I'll apply the fix once it passes the builds and we'll hear if people complain :-)

scpeters:
It used to work with 2.82 with the absolute path, but something has changed since then.

erwincoumans:
It is odd, because this CMAKE_INSTALL_PREFIX has never been used from the history:
https://code.google.com/p/bullet/source/browse/trunk/bullet.pc.cmake
Would installation in a relative path location still work with this CMAKE_INSTALL_PREFIX?

scpeters:
I think the difference is in the variable definitions in CMakeLists.txt; in [bullet 2.82](https://code.google.com/p/bullet/source/browse/trunk/CMakeLists.txt#401), the `LIB_DESTINATION` and other variables had `CMAKE_INSTALL_PREFIX` at the beginning. Now, however, these variables [use relative paths](https://github.com/bulletphysics/bullet3/blob/master/CMakeLists.txt#L287-L290).

PR #322, opened 2015-02-12
Add extra -I flag to bullet.pc to allow namespaced includes

### LINKED ISSUE #321: Allow #include <bullet/LinearMath/...>
Currently the bullet header files are installed to `${CMAKE_INSTALL_PREFIX}/include/bullet`, which is good, but the exported include path points to this exact folder, so that users wishing to include bullet files must use the following:
~~~
#include <LinearMath/btVector3.h>
~~~

It would be nice to allow the following:
~~~
#include <bullet/LinearMath/btVector3.h>
~~~

This helps indicate where header files are coming from (since LinearMath is a pretty generic name) and helps avoid header file name conflicts (however unlikely they may be).

This could be fixed by adding an additional `-I` flag to the pkg-config CFLAGS (keeping the existing one for backwards compatibility).

scpeters:
Fixes #321


erwincoumans:
ok, accepted. thanks for the contribution!

PR #328, opened 2015-02-21
Fix maxLinMotorForce for btSliderConstraint

scpeters:
In btSliderConstraint::getInfo2NonVirtual,
the maxLinMotorForce should be multiplied by
the simulation time step to compute an impulse
that is used to set up the constraints for the
linear motor. Instead of multiplying by timestep,
it was multiplying by fps, so it was effectively
dividing by the timestep. This fixes the error
by dividing by fps.

I also increased the max motor force (incorrectly named `maxMotorImpulse`) in the ForkLiftDemo so that it will have enough force to lift the load.

I've been using the joint motors to [implement Coulomb joint friction in Gazebo](https://bitbucket.org/osrf/gazebo/issue/381), and I successfully [implemented it in ODE](https://bitbucket.org/osrf/gazebo/pull-request/1221/implement-coulomb-joint-friction-for-ode) and for [btHingeConstraint](https://bitbucket.org/osrf/gazebo/pull-request/1317/enable-joint-friction-for-bullethingejoint), but my [test was failing for btSliderConstraint](https://bitbucket.org/osrf/gazebo/issue/1348/implement-coulomb-joint-friction-in). That led me to this bug. I'll make a workaround for 2.82, but it would be nice to fix this for 2.83.

### DIFF
--- a/Demos/ForkLiftDemo/ForkLiftDemo.cpp
+++ b/Demos/ForkLiftDemo/ForkLiftDemo.cpp
@@ -27,7 +27,7 @@ subject to the following restrictions:
 #include "BulletDynamics/MLCPSolvers/btMLCPSolver.h"
 
 
-btScalar maxMotorImpulse = 1400.f;
+btScalar maxMotorImpulse = 4000.f;
 
 //the sequential impulse solver has difficulties dealing with large mass ratios (differences), between loadMass and the fork parts
 btScalar loadMass = 350.f;//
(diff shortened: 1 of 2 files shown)


erwincoumans:
thanks, I'll look into it before next 2.83 release

scpeters:
Thanks.

For another reference point, note that [btGeneric6DofSpringConstraint](https://github.com/bulletphysics/bullet3/blob/d347bca2bad80420869217282535fb4ff919e8b5/src/BulletDynamics/ConstraintSolver/btGeneric6DofSpringConstraint.cpp#L134) divides by `fps` as well.

erwincoumans:
As the demos moved to examples, I manually applied your patch, thanks for your contribution!
https://github.com/erwincoumans/bullet3/commit/a165ff54c9e6ec789945a3d7c7e37cda47573efb
so it will be applied with that pull request later today

PR #330, opened 2015-02-23
Missing negative index check

lordloki:
Cause of https://developer.blender.org/T43124 crash

### DIFF
--- a/src/BulletDynamics/Dynamics/btDiscreteDynamicsWorld.cpp
+++ b/src/BulletDynamics/Dynamics/btDiscreteDynamicsWorld.cpp
@@ -776,7 +776,9 @@ void	btDiscreteDynamicsWorld::calculateSimulationIslands()
 				if (((colObj0) && (!(colObj0)->isStaticOrKinematicObject())) &&
 					((colObj1) && (!(colObj1)->isStaticOrKinematicObject())))
 				{
-					getSimulationIslandManager()->getUnionFind().unite((colObj0)->getIslandTag(),(colObj1)->getIslandTag());
+					if ((colObj0)->getIslandTag() != -1 && (colObj1)->getIslandTag() != -1) {
+						getSimulationIslandManager()->getUnionFind().unite((colObj0)->getIslandTag(),(colObj1)->getIslandTag());
+					}
 				}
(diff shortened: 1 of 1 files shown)


erwincoumans:
Thanks for the report!
This is a very odd bug. Is that a constraint between two rigid bodies that are not inserted in the dynamics world? If not, can you try to explain what is happening?

lordloki:
The issue occurs when we link, from other file, a group with 2 rigid bodies that have a constraint in between. If we link the same 2 rigid bodies individually, instead of forming a group, then there is no crash.

erwincoumans:
I suspect that only the constraint is added, and not the rigid bodies, could that be the case? I would have to setup a way to compile recent Blender source code for more info what happens with 'linking' .blend files and adding grouped rigid bodies/constraints. It used to be easy to compile Blender on Windows, using a lib/Windows repository for external dependencies, does that still exist?

lordloki:
I will investigate deeper to see what you comment.

the main steps to build blender from http://wiki.blender.org/index.php/Dev:Doc/Building_Blender/Windows/msvc/CMake

- Get Blender code:
cd C:\BlenderDEV
git clone git://git.blender.org/blender.git
cd blender
git submodule update --init --recursive
git submodule foreach git checkout master
git submodule foreach git pull --rebase origin master

- Get Visual 2013 libraries:
cd C:\BlenderDEV
svn checkout https://svn.blender.org/svnroot/bf-blender/trunk/lib/win64_vc12  lib/win64_vc12

- Generate Visual 2013 project files using CMake in cd C:\BlenderDEV\build
- Load Blender.sln solution
- Build INSTALL target

lordloki:
Hi,

as you can see in the fix https://developer.blender.org/rB0b4a71b07245d5370a02fae4dbde9195c9c58881
the issue raised when the group definitions were converted more that one time. This issue happened when a group instance was made from a linked group instance and both were on the active layer.

Due to this, I think that the pull request is not necessary anymore. Thanks for reviewing, Erwin.

erwincoumans:
Great, thanks for reporting the fix on the Blender side of things. Let's close the issue then!

PR #346, opened 2015-03-25
Add global and local implicit Gyroscopic force integration, with Dzhanibekov effect demo

erwincoumans:
See App_AllBullet2Demos and commit messages
Also added the option to create videos in that demo, using the --mp4=video.mp4 command-line argument. Requires ffmpeg to be installed and in the path.


scpeters:
I have a test for angular momentum conservation in gazebo; it would be interesting to run the test with each of these flags.

erwincoumans:
@scpeters See this video, the OpenDe/Cooper doesn't capture the effect: https://www.youtube.com/watch?v=qw74KV_R5Y0
Yes, some more tests would be useful. The implicit integrator dampens the coriolis/gyroscopic effect.

scpeters:
We are using a fork of ODE, so I'm not sure if we have that OpenDe/Cooper code.

PR #353, opened 2015-04-06
Fix a potential division by zero for cone twist constraints.

cameronwhite:
This could occur if the two objects are exactly aligned and the angle limits are configured like a hinge. This can reproduced by changing line 211 of `RagdollDemo.cpp` to `coneC->setLimit(0, M_PI_4, 0);`


erwincoumans:
Thanks for the report and the fix!

PR #361, opened 2015-04-21
Prevent SSE Alignment crash in VS2010 and up

kbogert:
Note, I'm not the original author of this fix (author's name is Viatrak in the forums).

See http://bulletphysics.org/Bullet/phpBB3/viewtopic.php?t=8915

### DIFF
--- a/src/LinearMath/btQuadWord.h
+++ b/src/LinearMath/btQuadWord.h
@@ -73,7 +73,7 @@ public:
 
 	public:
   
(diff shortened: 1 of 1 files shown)


AndresTraks:
This problem also comes up when writing the result of btRigidBody.getOrientation to an unaligned location, so +1 to this fix.
https://code.google.com/p/bulletsharp/issues/detail?id=42

erwincoumans:
Thanks for the report and fix, good find!

PR #363, opened 2015-04-29
replace btgui/Demos/Demos3 with 'examples' and move all demos in 'ExampleBrowser'

erwincoumans:
work-in-progress for the Bullet 2.83 release. Still need a bit more work and re-insert documentation.


scpeters:
Looks like there are some linking errors in travis.

erwincoumans:
yes, cmake files are not up-to-date, only premake. i'll fix that today

PR #375, opened 2015-05-07
don't use glSwapAPPLE, it fails on some OSX versions

erwincoumans:
thanks Steven for the report

### DIFF
--- a/examples/OpenGLWindow/MacOpenGLWindow.mm
+++ b/examples/OpenGLWindow/MacOpenGLWindow.mm
@@ -5,7 +5,6 @@
 #include "OpenGLInclude.h"
 
 
-
 #include <stdlib.h>
 #include <stdio.h>
 #include <stddef.h>
@@ -1036,7 +1035,12 @@ void MacOpenGLWindow::startRendering()
 void MacOpenGLWindow::endRendering()
 {
     [m_internalData->m_myview MakeCurrent];
-    glSwapAPPLE();
+//#ifndef NO_OPENGL3
+//	glSwapAPPLE();
+//#else
+ [m_internalData->m_myview->m_context flushBuffer];
+//  #endif 
+
 }
 
 bool MacOpenGLWindow::requestedExit() const


scpeters:
Sorry, now I get the following on Yosemite:
~~~
/Users/scpeters/osrf/bullet3/examples/OpenGLWindow/MacOpenGLWindow.mm:1041:29: error: instance variable
      'm_context' is protected
 [m_internalData->m_myview->m_context flushBuffer];
                            ^
1 error generated.
~~~

scpeters:
We should set up CI for OS X as well.

scpeters:
It fails on both yosemite and mountain lion for me.

erwincoumans:
Yes, we should setup CI for various versions of OSX, because fixing it in one version can still be broken in another, as we noticed a few times. Do you have time to help set it up with appveyor or travis?

In the meanwhile, I have a fix ready, commit in a few minutes :-)

erwincoumans:
if you have time, please try this one: 
https://github.com/bulletphysics/bullet3/pull/376

scpeters:
testing...

scpeters:
looks good on ubuntu 14.04, yosemite and mountain lion

scpeters:
~~~
ndefined symbols for architecture x86_64:
  "GLPrimitiveRenderer::setScreenSize(int, int)", referenced from:
      SimpleResizeCallback(float, float) in SimpleOpenGL3App.o
  "GLPrimitiveRenderer::drawTexturedRect3D(PrimVertex const&, PrimVertex const&, PrimVertex const&, PrimVertex const&, float*, float*, bool)", referenced from:
      SimpleOpenGL3App::drawText3D(char const*, float, float, float, float) in SimpleOpenGL3App.o
  "GLPrimitiveRenderer::GLPrimitiveRenderer(int, int)", referenced from:
      SimpleOpenGL3App::SimpleOpenGL3App(char const*, int, int) in SimpleOpenGL3App.o
  "GLInstancingRenderer::InitShaders()", referenced from:
      SimpleOpenGL3App::SimpleOpenGL3App(char const*, int, int) in SimpleOpenGL3App.o
  "GLInstancingRenderer::GLInstancingRenderer(int, int)", referenced from:
      SimpleOpenGL3App::SimpleOpenGL3App(char const*, int, int) in SimpleOpenGL3App.o
  "OpenGL2RenderCallbacks::OpenGL2RenderCallbacks(GLPrimitiveRenderer*)", referenced from:
      SimpleOpenGL3App::SimpleOpenGL3App(char const*, int, int) in SimpleOpenGL3App.o
  "_stbi_write_png", referenced from:
      SimpleOpenGL3App::swapBuffer() in SimpleOpenGL3App.o
ld: symbol(s) not found for architecture x86_64
~~~

erwincoumans:
How do I replicate this? I never use BUILD_SHARED.
I tried running this, but couldn't replicate an error on Yosmite:
cmake .. -G "Unix Makefiles" -DBUILD_SHARED_LIBS=ON
make

scpeters:
It works on yosemite, just not on Mountain Lion.

You can decide how much you want to support.

scpeters:
I'll try Mavericks as well

scpeters:
works on Mavericks, the only failure I see is on Mountain Lion.

scpeters:
It could be a bug in the linker?

scpeters:
The `GLPrimitiveRenderer` is [disabled if OpenGL3 is not available](https://github.com/bulletphysics/bullet3/commit/0da584a42d1dc8792f2efd6e169cea6b5dc0caf8#diff-07c1d8af86c634224739c8b549cd999eR1). I don't know why the shared library linker wants to see them.

erwincoumans:
So the static version works fine, just the shared lib version on Mountain Lion is broken?
Then I'm happy to go ahead with marking a new release.

scpeters:
It's complaining about symbols from `SimpleOpenGL3App.o`; I thought those were disabled. Can we double check that logic really quick?

scpeters:
I think [SimpleOpenGL3App](https://github.com/bulletphysics/bullet3/blob/master/examples/OpenGLWindow/SimpleOpenGL3App.cpp) is being compiled on Mountain Lion. We should only compile that if OpenGL3 is available right? I think that is causing the linking errors.

PR #379, opened 2015-05-07
Don't compile SimpleOpenGL3App.* if NOT OPENGL3

scpeters:
This should fix the Mountain Lion shared library issues


erwincoumans:
I already added the check inside the cpp file, so we don't have to change each build system (premake, cmake and customers own build system). See https://github.com/erwincoumans/bullet3/commit/cedd0432bfe6789445e9031fcd8cd7b27e18ac53

PR #380, opened 2015-05-07
don't compile SimpleOpenGL3App if NO_OPENGL3 is defined

erwincoumans:


scpeters:
See also #379, either one should work.

scpeters:
This works on my virtual machine.

PR #382, opened 2015-05-08
Update .travis.yml

erwincoumans:
add '- sudo make install' and hope it works on Travis' OSX machines too.


scpeters:
Looks like it succeeded. Thanks for adding it!

PR #385, opened 2015-05-17
New btConeshape member functions setHeight and setRadius

lordloki:
These 2 new functions allow to modify the cone parameter after cone creation. It is useful when you use the cone as radar and you want to modify the raange or/and angle.


erwincoumans:
Seems fine. Just remember to remove the rigid body from the world before making those changes, and then re-insert it.

PR #386, opened 2015-05-18
Fix #359 - Improve performance for many islands of constraints.

### LINKED ISSUE #359: Solving constraints becomes very slow if the number of islands is large
For every island, `InplaceSolverIslandCallback::processIsland()` iterates through the entire list of constraints to find the constraints that belong to that island. Since the constraints are already sorted by island id, it would be much faster to iterate through `m_sortedConstraints` once in `InplaceSolverIslandCallback::setup()` and precompute the index into `m_sortedConstraints` where each island's constraints begin.

cameronwhite:
For a modified ragdoll demo (35 rows of ragdolls and more spacing between them), this improved performance by about 30% overall. Inside of `processIslands` in the profiler, the time spent in `solveGroup` went from around 35% to >95%.


erwincoumans:
Thanks for this effort, and excuses for the huge delay. I am cleaning up old PR's. Tried this on the Ragdolls Benchmark in the ExampleBrowser, disabling gravity, press 'g' (disable GUI) and hold 'p' for a while: recording detailed performance trace json file. Open the timings_0.json trace in Google Chrome using "about://tracing"). Comparison shows no benefit. Let't close it for now, unless you can show more convincing benefits. See old and new images, both around 3.7-3.9 ms.

old

new

cameronwhite:
Hi Erwin,

Did you also adjust the ragdoll demo to create more constraints and islands? In `BenchmarkDemo::createTest3()`, I increased `size` to 35, and `sizeX` and `sizeY` to 2.0, and saw about 20ms spent in `processIslands()` with the original code, compared to 7.5ms with these changes.

erwincoumans:
No I didn't. Let me try it again soon, and I'll report the timings. Thanks for getting back to me. Also, I will look at the other pull requests soon, there are too many outstanding.

cameronwhite:
I forgot this PR was still open - I've rebased it on the latest upstream master and applied clang format

With the changes to the ragdoll example described in https://github.com/bulletphysics/bullet3/pull/386#issuecomment-329633730, I see a drop of about 10ms per timestep overall

PR #387, opened 2015-05-18
Fix #69 - Fix collisions between concave shapes and transformed planes.

### LINKED ISSUE #69: Concave shapes collide incorrectly against transformed btStaticPlaneShapes (was googlecode issue  Issue 781)
copied from https://code.google.com/p/bullet/issues/detail?id=781&colspec=Modified%20ID%20Type%20Stars%20Status%20Owner%20Summary
patch is there

In btGImpactCollisionAlgorithm, the plane equation is not transformed correctly (the plane constant is incorrect). I've attached a patch that fixes this issue.

cameronwhite:
This corrects the computation of the plane constant for transformed plane shapes.


erwincoumans:
This patch seems to be already applied, thanks for your contribution.

Issue #298, opened 2014-12-16
static variable use in btMLCPSolver.cpp prevents independent solving in multithreaded environments


CodeMatias:
Several variables in btMLCPSolver::createMLCPFast and btMLCPSolver::createMLCP are defined as static, which prevents independent simulations within the same process from properly executing with MLCP solvers.

Tests done by moving the variables to non-static with class scope allows safe behavior, but is there a performance or other reason for those variables to be static?

erwincoumans:
I'm not planning to depend on C++11 just yet, so moving the memory as a class member should be fine. Do you mind creating a patch? Otherwise I'll do it later.

lunkhound:
Someone on the Bullet forums (Basroil), have said they are making a patch for this.

[http://bulletphysics.org/Bullet/phpBB3/viewtopic.php?f=9&t=10232&sid=c266427de1e84e2b153b5de6f98dfdad&start=45]

erwincoumans:
Would it be fixed by this large pull request?
https://github.com/bulletphysics/bullet3/pull/303
I still have to review if all those changes are acceptable (rather not break the API etc)

lunkhound:
No, I was waiting for Basroil to submit his patch, otherwise I would have included it in that pull request.

If it would help, I can try to condense the number of commits in #303, there were a few dead-end lines of development that I ended up backing off of.

PR #390, opened 2015-05-21
Threading for Bullet2 using OpenMP, TBB, or PPL

lunkhound:
Adds a multithreading demo to the example browser which uses either OpenMP, Intel TBB or Microsoft PPL to provide task scheduling. (issue #126) Demo allows task scheduler API to be changed on the fly. On my quad-core machine the bullet part of the demo (ignoring rendering, etc) runs about 3 times faster with threading enabled than in single-threaded mode.

Also makes Dvbt rayTest threadsafe (issue #254, #258)
Discussion (here)[http://www.bulletphysics.org/Bullet/phpBB3/viewtopic.php?f=9&t=10232]

Areas that run in parallel:

narrowphase (dispatchAllPairs)
discreteDynamicsWorld solve islands (each island is a separate task)
discreteDynamicsWorld predictUnconstraintMotion
discreteDynamicsWorld createPredictiveContacts
discreteDynamicsWorld integrateTransforms

There were some significant changes to how the btSimulationIslandManager works, so that islands could be dispatched in parallel to different solvers.
Based on an earlier (pull request)[https://github.com/bulletphysics/bullet3/pull/303].
This version has been collapsed into fewer commits and many dead ends were stripped out. All of the previous patch's changes in the sequential impulse constraint solver were dropped, it is simpler and more effective to run islands of constraints in parallel.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ SET(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG")
 OPTION(USE_DOUBLE_PRECISION "Use double precision"	OFF)
 OPTION(USE_GRAPHICAL_BENCHMARK "Use Graphical Benchmark" ON)
 OPTION(BUILD_SHARED_LIBS "Use shared libraries" OFF)
+OPTION(BULLET2_USE_THREAD_LOCKS "Build Bullet 2 libraries with mutex locking around certain operations" OFF)
 
 OPTION(USE_MSVC_INCREMENTAL_LINKING "Use MSVC Incremental Linking" OFF)
 OPTION(USE_CUSTOM_VECTOR_MATH "Use custom vectormath library" OFF)
@@ -144,6 +145,10 @@ IF(USE_GRAPHICAL_BENCHMARK)
 ADD_DEFINITIONS( -DUSE_GRAPHICAL_BENCHMARK)
 ENDIF (USE_GRAPHICAL_BENCHMARK)
 
+IF(BULLET2_USE_THREAD_LOCKS)
+ADD_DEFINITIONS( -DBT_THREADSAFE=1 )
+ENDIF (BULLET2_USE_THREAD_LOCKS)
+
 IF (WIN32)
 OPTION(USE_GLUT "Use Glut"	ON)
 ADD_DEFINITIONS( -D_CRT_SECURE_NO_WARNINGS )
@@ -184,8 +189,6 @@ ENDIF (OPENGL_FOUND)
 #FIND_PACKAGE(GLU)
 
 
-
-
 IF (APPLE)
   FIND_LIBRARY(COCOA_LIBRARY Cocoa)
 ENDIF()
@@ -224,6 +227,15 @@ IF(BUILD_BULLET2_DEMOS)
         IF(EXISTS ${BULLET_PHYSICS_SOURCE_DIR}/examples AND IS_DIRECTORY ${BULLET_PHYSICS_SOURCE_DIR}/examples)
                 SUBDIRS(examples)
         ENDIF()
+
+        IF (BULLET2_USE_THREAD_LOCKS)
+            OPTION(BULLET2_MULTITHREADED_OPEN_MP_DEMO "Build Bullet 2 MultithreadedDemo using OpenMP (requires a compiler with OpenMP support)" OFF)
+            OPTION(BULLET2_MULTITHREADED_TBB_DEMO "Build Bullet 2 MultithreadedDemo using Intel Threading Building Blocks (requires the TBB library to be already installed)" OFF)
+            IF (MSVC)
+                OPTION(BULLET2_MULTITHREADED_PPL_DEMO "Build Bullet 2 MultithreadedDemo using Microsoft Parallel Patterns Library (requires MSVC compiler)" OFF)
+            ENDIF (MSVC)
+        ENDIF (BULLET2_USE_THREAD_LOCKS)
+
 ENDIF(BUILD_BULLET2_DEMOS)
 
 
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,1 +1,2 @@
 SUBDIRS( ExampleBrowser HelloWorld BasicDemo ThirdPartyLibs/Gwen OpenGLWindow)
+
--- a/examples/ExampleBrowser/CMakeLists.txt
+++ b/examples/ExampleBrowser/CMakeLists.txt
@@ -19,6 +19,9 @@ SET(App_ExampleBrowser_SRCS
 	../BasicDemo/BasicExample.h
 	../ForkLift/ForkLiftDemo.cpp
 	../ForkLift/ForkLiftDemo.h
+	../MultiThreadedDemo/MultiThreadedDemo.cpp
+	../MultiThreadedDemo/MultiThreadedDemo.h
+    ../MultiThreadedDemo/ParallelFor.h
 	../GyroscopicDemo/GyroscopicSetup.cpp
 	../GyroscopicDemo/GyroscopicSetup.h
 	../Planar2D/Planar2D.cpp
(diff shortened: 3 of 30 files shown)


cameronwhite:
I think the multi-threaded narrowphase here will produce non-deterministic simulation results (the constraint solver is sensitive to the ordering of the contact manifolds, and allocating manifolds from multiple threads will result in a different order for `m_manifoldPtr` when re-running a simulation). I wrote a multi-threaded btCollisionDispatcher subclass for the Bullet solver in Houdini, so I'll see if I can make that available (it currently has some dependencies on other parts of our codebase).

My implementation doesn't involve any locking, though - the simulations we deal with are typically large enough that the memory pools don't help much, and the `m_manifoldPtr` list is built using the `getAllContactManifolds` method after dispatching the pairs and allocating the manifolds - this ensures that the order of the manifolds is deterministic (provided that the list of overlapping pairs has a deterministic order).

lunkhound:
Yes this narrowphase is about as simple as possible, and doesn't attempt to be deterministic. That's a good point, thanks for pointing that out.

I'd certainly be interested in seeing your narrowphase if you are able to share it. Did you need to make any changes to Bullet source for it?

cameronwhite:
I'll double check our Bullet patches tomorrow, but I think the only required changes were to add a lock in the concave collision detection algorithm and to allocate the `btVoronoiSimplexSolver` on the stack in the convex vs convex algorithm, and it looks like you've made similar changes.

lunkhound:
I didn't make any change the concave-convex algorithm, do you mean Gimpact? Thanks for looking into that.

cameronwhite:
Sorry, yeah, I meant the Gimpact shape. It looks like I did it by adding actual locking around the modifications inside of `TrimeshPrimitiveManager::lock()` and `TrimeshPrimitiveManager::unlock()`

lunkhound:
It's reassuring that you identified the same threadsafe issues that I did. Does your implementation solve constraints in parallel also?

cameronwhite:
We do have the option to use `btParalleConstraintSolver` (which was removed in 2.83), but it's on my todo list to look into multi-threading the regular constraint solver. Many of our simulations have large islands (e.g. a building collapsing), so I suspect that a graph coloring approach to splitting the constraints into batches might provide better performance in our case than solving islands in parallel.

lunkhound:
Yep if everything is in one big island, then there's no benefit to running the islands in parallel. It is very scene-dependent. It would definitely be nice to parallelize the constraint solver. I tried to do that with `btSequentialImpulseConstraintSolver` but it got messy and the results were mixed so I pulled it all out. The graph coloring idea sounds interesting. I don't see how it could be deterministic though, could it be?

One nice feature about running the islands in parallel is that it works with different constraint solvers, not just the sequential impulse one.

cameronwhite:
The idea is to split the constraints into batches that don't share any bodies, and then for each batch you can safely solve those constraints in parallel. For e.g. a chain of bodies A-B-C-D-E you can solve constraints AB and CD in parallel, followed by BC and DE in parallel

erwincoumans:
Yes, graph coloring (splitting constraints in batches that don't share bodies) is/was implemented both in the btParallelConstraintSolver (which is now removed) and in the OpenCL GPU rigid body pipeline.
I removed the BulletMultiThreaded because it was not fully supporting most features, and because I want to move to Bullet 3.x and re-implement multi-threading.

Problem is that there is no team working on Bullet, it is just myself and occasional contributors, so things move slowly. Now you are pulling the attention back to Bullet 2.x, re-enabling multithreading using Intel TBB etc. Of course I welcome contributions and give some preliminary feedback on this pull request. I'll try to give some feedback later.

1) By default, Bullet should compile out-of-the-box without any external dependencies: Bullet cannot have any dependency on C++11, BOOST, STL, Intel TBB, OpenMP.
We can have some optional CMake/premake flag that enables Intel TBB and pthreads for example.

2) If we re-enable multi-threading for Bullet 2.x, it should support all features of one or more of the existing constraint solvers and collision dispatcher.

3) I prefer parallelizing using graph coloring, instead of per-island. Perhaps we can have both.

There are some threading classes that I want to re-enable, if possible, similar to the BulletMultiThreading classes. I haven't looked your the code of this pull request yet, but perhaps we can enable Intel TBB, OpenMP and those b3ThreadSupportInterface and user-defined thread support? See also https://github.com/bulletphysics/bullet3/tree/old_demos/btgui/MultiThreading

Thanks!
Erwin

tamaynard:
For the MLCP solver suite (and possibly Featherstone as well), it would be useful to keep the per-island approach. I don't see how a graph-colouring approach would apply for those methods, and the nature of matrix factorization means that the overhead involved in parallelizing some of the dot product operations, etc., would outweigh any benefits for solving the smaller, stiff systems that one would typically use a direct solver for anyway. At least that's my thought on it, I would love for someone to prove me wrong however :-)

lunkhound:
Thanks for the feedback Erwin.

On (1). Yes I've tried to adhere to that, not introducing any extra dependencies out of the box. I've also tried to follow the principal of "you only pay for what you use" -- meaning that if someone doesn't compile with the threadsafe flag, there shouldn't be any performance or memory penalties compared to without this patch.
The Cmake flag is called `BULLET2_USE_THREAD_LOCKS`, which defaults to off. With that flag off, no dependencies are added. In addition, there should be no performance degradation because the locking calls (`btMutexLock`, `btMutexUnlock`) become inlined empty functions.
If `BULLET2_USE_THREAD_LOCKS` is **enabled**, then the patch requires having C++11 or a recent MSVC compiler. As the name suggests, this flag by itself doesn't enable multithreading, it merely makes certain operations threadsafe.
The only thing that runs multithreaded is the MultiThreadedDemo. In order to build that, one must set the `BULLET2_USE_THREAD_LOCKS` flag as well as `BUILD_BULLET2_DEMOS`. Then 3 new Cmake flags should appear (all defaulted to off):
  `BULLET2_MULTITHREADED_OPEN_MP_DEMO`
  `BULLET2_MULTITHREADED_TBB_DEMO`
  `BULLET2_MULTITHREADED_PPL_DEMO`
Each of these adds a dependency. Any combination of the flags is supported, though at least one of them must be enabled to see the MultiThreadedDemo. TBB is the only external dependency, if TBB is selected, Cmake will require directories be specified pointing to an installed TBB.

On (2). You'll notice that in this patch, no changes were made to any constraint solvers. I simply added the ability to simulate islands in parallel. For collision detection, minor changes were made to make some narrowphase checks threadsafe. There should be no loss of functionality. Although, as cameronwhite pointed out above, the collision dispatcher in MultiThreadedDemo is nondeterministic.

So a big difference between this patch and the old BulletMultiThreaded is that in this patch, the multithreaded version is running basically the same code as single-threaded bullet. The only difference is that a few `for` loops are swapped out with a `parallelFor`. The old BulletMultiThreaded stuff looked like quite a maintenance burden due to duplicated classes such as `btParallelConstraintSolver`.

On (3). No reason it has to be one or the other -- both would be best. It works fine with TBB to have tasks launching other tasks. But I think the user should be able to choose one or the other (or both simultaneously).

One reason I have all of the task scheduler specific stuff outside of the main bullet libraries is that I didn't want Bullet to be dictating which task scheduler to use. I'd like to be able to use my own custom task scheduler if I choose, so that it integrates better with the rest of my application.

Looking at the `b3ThreadSupportInterface` class, I can't work out how you start tasks with it. The interface I'm using is `parallelFor`, look here [https://github.com/lunkhound/bullet3/commit/c820ed13452eb977b23a2e0c335695fc4244a9a5]
and search for `parallelFor` to see how it is used.

My apologies for taking your attention away from Bullet3. But until Bullet3 is stable and ready for production, I'd really like to be able to have multithreading in Bullet2 working. And by reusing the single-threaded code, I'm hoping to minimize the burden.

lunkhound:
Added fix for [https://github.com/bulletphysics/bullet3/issues/298].

erwincoumans:
I just applied the patch for testing on Linux/Ubuntu. Without enabling any flags, cmake builds fine.

If we enable the definition BULLET2_USE_THREAD_LOCKS in cmake, the following error shows up:
/usr/local/google/home/erwincoumans/develop/bullet3/src/LinearMath/btThreads.cpp:151:2: error: #error "no threading primitives defined -- unknown platform"

Note I didn't enable any other options yet. Is this to be expected?

lunkhound:
Hmmm. I wouldn't expect that. I know at least one person had it working with GCC, though I don't know if it was Ubuntu or not.
The check for C++11 compliance must not be working for your configuration (btThreads.cpp, line 43). The check is `#if __cplusplus >= 201103L` which is a more conservative check than it needs to be since it only requires a subset of <atomic> and <thread> from C++11 in order to work.
Is there some compiler flag that suppresses C++11 compatibility being used?

erwincoumans:
No idea about the C++11 thing, but I likely want to implement a solution based on C++2003 with the b3ThreadSupportInterface that works out-of-the-box. I'll give some more feedback about the patch soon. Thanks!

lunkhound:
b3ThreadSupportInterface lacks some functionality. It lacks any way for code to ask if it is the main thread or a worker thread. This is needed to keep the profiling system in Bullet 2.x from running amok.
Also it lacks a way of starting tasks which would be needed to do things.
b3ThreadSupportInterface looks like it was designed for the Playstation 3 (Cell) hardware, rather than modern multicore CPUs. Do you intend to support PS3 type hardware in Bullet 2?

erwincoumans:
Don't worry, I will make it work.

lunkhound:
FYI, I think a bug was introduced in this commit: [https://github.com/lunkhound/bullet3/commit/5b8f306271933511384d59319e0660f3e3fc64e8]
I'm looking into it now.

lunkhound:
A race condition was discovered with solving islands in parallel if kinematic objects are present. If a kinematic object is part of 2 or more islands, and the islands are running in separate threads, the `companionId` field of the kinematic body will get stomped on in `btSequentialImpulseConstraintSolver::getOrInitSolverBody`.

Basically the solver is storing it's own private data in an object that is being shared between threads. That can't work. If each body had a unique-id assigned to it, then the solver could store the data it wants in an array indexed by the body's unique-id.

miskeletor:
Was the kinematic thing ever solved?
If this branch can multithread solving of MLCP islands, this would help me a lot.

lunkhound:
This branch [https://github.com/lunkhound/bullet3/tree/kinematic-fix] contains a fix for this kinematic issue. Basically I added a unique id to each collision body and each island solver can store thread-specific info in a separate array indexed by the id. I have kept that change in a separate branch because it likely causes trouble for the featherstone multi-body point-to-point constraint. The MultiBody dynamics world isn't threadable anyway, so it's no loss if you want to use threads.

miskeletor:
This is great!
The speedup of my project is really impressive using this branch.
The kinematic-fix doesn't work for me though. It crashes when i remove a static body and then change it and recreate another one (with modifications to a compundshape) in the same place.
Since i use the MLCP solver another thing that helped a lot with performance was removing the minimum size of a simulation island and reducing solver batch size.

codetiger:
I am wondering why MultiThreading was removed from BulletPhysics. I understand that you are moving towards more advanced OpenCL and GPU. 
Please remember we are using BulletPhysics for Mobile phone gaming also. Personally I've been a big fan of MultiThreading, VFP/NEON on iOS devices gives us upto 3x speed. Now we are badly missing it.

BluePrintRandom:
does this mean blender game bullet may  possible to thread soon?

benelot:
Basically, as soon as there are no problems with the multithreading pipeline, yes. Somebody would need to port Blender to the newest Bullet version though.

PR #392, opened 2015-05-22
Auto limitation of spring stiffness and damping in btGeneric6DofSprin…

GaborPuhr:
…g2Constraint is now optional.

Fix: spring stiffness limitation used the mass incorrectly in btGeneric6DofSpring2Constraint.


erwincoumans:
Thanks for the fix/improvement!

See also discussion here:
http://www.bulletphysics.org/Bullet/phpBB3/viewtopic.php?f=9&t=10541

PR #399, opened 2015-06-02
Avoid calling calcAngleInfo2() twice when setting up cone twist constraints.

cameronwhite:
Currently it is called from both `getInfo1()` and `getInfo2()`, but `getInfo1()` is always called before `getInfo2()` and the constraint is never modified in between. This change is similar to the optimization that is performed in btHingeConstraint.


erwincoumans:
Closing very old PR's. It doesn't seem to be a significant performance improvement, so let's drop it.

PR #404, opened 2015-06-08
 Improve Compound Shape Construction.

donggas90:
If pass children's initial capacity of CompoundShape, we can avoid memory reallocation and copy occured from Array when adding a lot of children. 
Fixed some odd code and removed needless code in Array.(Not All problems, particularly sticked outs.)


erwincoumans:
I'll look at the compound shape construction. We cannot use the 'explicit' keyword because we don't require c++11 for Bullet. Thanks for the contribution!

donggas90:
Hello, Mr. Coumans.
Explicit Constructor is not a C++11's feature. :)
http://en.cppreference.com/w/cpp/language/explicit
https://msdn.microsoft.com/en-us/library/h1y7x448(v=vs.80).aspx

erwincoumans:
ok, it looks ok to me. Thanks for the contribution!

PR #444, opened 2015-08-03
Make btCompoundShape data members protected instead of private

AndrewMeadows:
Without risk the data members of btCompoundShape can be made protected, thereby allowing derived classes to change them directly.

While it is true some data members are already exposed for modification via getDataMember() methods (such as btCompoundShape::getChildShape() and getDynamicAabbTree()) not all of them are exposed (such as btCompound::m_updateRevision).


erwincoumans:
sure, that is a harmless change. Thanks for the patch.

Issue #64, opened 2014-03-30
btQuaternion - Slerp sometimes return error values (was googlecode Issue 786)


erwincoumans:
https://code.google.com/p/bullet/issues/detail?id=786&colspec=Modified%20ID%20Type%20Stars%20Status%20Owner%20Summary

Reported by creagamestudio, Jan 11, 2014
What steps will reproduce the problem?
Answer:
at btQuaternion.h (line 508): const btScalar d = btScalar(1.0) / btSin(theta);
"btSin(theta)" sometimes may be equal 0...

My fix is:
const btScalar sin_theta = btSin(theta);
if (sin_theta == 0)
  return q;
const btScalar d = btScalar(1.0) / sin_theta;

T4Larson:
See #446 : Mathematically, theta and d should be larger than _epsilon for 1.0-SIMD_EPSILON (also increased precision for results near 0.0), so if the dotproduct is below the presicion threshold, theta and d would be >0, so no extra check is probably needed after that point.

PR #446, opened 2015-08-05
fix: btQuaternion.slerp() may result in NaN-quaternions

T4Larson:
The btQuaternion.slerp() may produce result quaternions full of NaNs when the interpolated quats are very similar or identical, when compiled with optimizations.
This happened in optimized builds of another project (https://github.com/opentomb/OpenTomb) using gcc 4.9.2 with -O3, where the absproduct<1.0 condition is seemingly optimized to compare the dot product with higher (native) float  precision, but is truncated to 0 in the following calculation, leading to 0 theta and d.
Upd: This is very likely related to #64


erwincoumans:
thanks for the contribution, it took too long to just merge this one.

PR #454, opened 2015-08-07
fix and test on MinGW-w64

xantares:


d3x0r:
+           RUNTIME DESTINATION bin
+           LIBRARY DESTINATION lib${LIB_SUFFIX}
+           ARCHIVE DESTINATION lib${LIB_SUFFIX})

I guess Bullet builds static lib... but LIBRARY DESTINATION is also where .DLL files go, which should be bin not lib for windows.

include( GNUInstallDirs ) is the recommended place to get destinations by cmake; though it really doesn't work all that well... since it doesn't produce bin/lib for windows but tries to drop all into the install destination.

and do you really HAVE to change all the install targets to a macro?

xantares:
@d3x0r:
 - no, you're wrong, .DLL files are treated as RUNTIME
- do you have a case against macros that avoid duplicate code, or it's just personal ?

d3x0r:
(dlls are runtime) yeah; forgot again what my issue with that was... library destination SHOULD apply to DLLs and doesn't... or RUNTIME should apply to .so's and doesn't.  

My reaction was more that it was a 'FIX for MinGW-w64' and it was a do-nothing change that just complicates finding how things installed.

Looking further; shouldn't the macro also handle the APPLE target switch; and the header file install?

PR #457, opened 2015-08-12
Make installation of generated CMake files optional

aaronmjacobs:
When bullet is used as a subproject, installation of the generated CMake files is not always desired. This change makes it optional (enabled by default).


erwincoumans:
Trying to process some old PRs, thanks for the help (applied it manually, due to conflict)
https://github.com/bulletphysics/bullet3/pull/1335

Issue #389, opened 2015-05-19
MinGW segmentation fault when building


demystificator:
Hi,

I encounter a segmentation fault everytime I try to build Bullet ; at the same line 

C:\Bullet\git\src\BulletCollision\CollisionDispatch\btInternalEdgeUtility.cpp:310:2: internal compiler error: Segmentation fault

I'm building on :
Win7 64bits
GCC : 4.8.1 (MinGW32)
CMake : 3.1.0
Trying to build only the core lib.

Re-dl / reinstalling my compiler doesn't change anything. The same goes with bullet version (tried from zip and directly with git hub).

I found this : https://github.com/bulletphysics/bullet3/issues/39

It is the closest topic to my problem I could find (and the most recent).

Hope it'll get fixed (or I'd be told why did I miss).

erwincoumans:
You are using a buggy compiler. If you have a workaround for your buggy compiler that makes Bullet compile, can you please share it? Otherwise, just delete the btInternalEdgeUtility.cpp and you should be good to go.

demystificator:
Well, sorry couldn't find any workaround so I just deleted the file :/ 
Hope it'll get fixed asap.

PR #466, opened 2015-08-26
Update btInternalEdgeUtility.cpp, closes issue #389

rioharu:
Fixes issue #389 which causes a segfault in MinGW's GCC by removing code which did not affect subsequent operations.


erwincoumans:
sure, looks harmless and good if that worksaround some GCC issues.

PR #473, opened 2015-09-09
Detach posix shared memory correctly

takuyanakaoka:
* shmget() fails after shmctl() is called with IPC_RMID. This patch
removes shmget() in releaseSharedMemory so that a client can detach
a shared memory correctly after a server removed it.
* A server should also detach a shared memory.


erwincoumans:
thanks for the fixes!

Issue #37, opened 2014-02-09
Add Approximate Convex Decomposition (HACD, v-HACD etc)


erwincoumans:
Bullet 2.x has convex decomposition, 
https://code.google.com/p/bullet/source/browse/#svn%2Ftrunk%2FExtras%2FHACD
It would be good to add it again.
There are some (possible) improved implementations, see
https://code.google.com/p/v-hacd/
https://code.google.com/p/testhacd/

PR #477, opened 2015-09-16
Replace HACD with V-HACD and add demo

AndresTraks:
V-HACD issue: https://github.com/bulletphysics/bullet3/issues/37

V-HACD forum thread:
http://www.bulletphysics.org/Bullet/phpBB3/viewtopic.php?f=4&t=6619&start=15


erwincoumans:
Great, thanks for the contribution! It is one of the first user-contributed examples using the new Example Browser :-) I'll have a look at it soon. It looks like the DLL version has a compile issue, perhaps you need to add a dependency to LinearMath library?

AndresTraks:
It does seem like it. It works on Windows, but I can check Linux tomorrow.

AndresTraks:
Ah, I think at some point LinearMath was no longer needed in HACD and the link was commented out from the build script. Now that V-HACD uses LinearMath again, the source code includes btConvexHullComputer.h and btAlignedObjectArray.h. But I left them out and included the ones in bullet instead without adding back the linker dependency to LinearMath.
Should work now.

AndresTraks:
VHACD is included now, so I'll close this.

PR #483, opened 2015-10-08
Minor fix to test of compiler symbol

d3x0r:
#ifdef _MSVC_VER 
should be
#ifdef _MSC_VER

I know this looks messy, but it all boils down to a single line change in one file.


d3x0r:
Ended up with other disrelated 'fixes' which should have been rolled back before making this request.

PR #484, opened 2015-10-08
add_missing_stdio_h

juj:
Add missing <stdio.h> include in src/LinearMath/btMatrixX.h for printf() access. (missing in Emscripten/musl libc builds otherwise)

### DIFF
--- a/src/LinearMath/btMatrixX.h
+++ b/src/LinearMath/btMatrixX.h
@@ -19,6 +19,7 @@ subject to the following restrictions:
 
 #include "LinearMath/btQuickprof.h"
 #include "LinearMath/btAlignedObjectArray.h"
+#include <stdio.h>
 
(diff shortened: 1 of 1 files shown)


scpeters:
+1

AndresTraks:
Only your branch uses printf, though.

scpeters:
@AndresTraks what about the following:

* https://github.com/bulletphysics/bullet3/blob/master/src/LinearMath/btMatrixX.h#L294-L307

AndresTraks:
Sorry, you're right! I was looking for "printf", but Visual Studio didn't look into btMatrixX.h since the file isn't included in the project file created by CMake.

scpeters:
No problem :)

erwincoumans:
sure, that's a bug, thanks for the report and fix!

PR #491, opened 2015-10-15
Fixed missing include directories

nicokruithof:
When I added the SimpleOpenGL3, two include directories were missing


erwincoumans:
Sure, that looks good. We could as well enable the compilation of the SimpleOpenGL3 example (it is mainly a simple start project to create an OpenGL context, without external dependencies. We can also remove 'gwen' library from SimpleOpenGL3 because it is not used.

PR #496, opened 2015-10-21
Fix slow simulation in PhysicsServerExample

takuyanakaoka:
This change removes maxSteps to regain simulation speed in PhysicsServerExample.

I thought it is unintentionally committed debugging code in #488 . If you have any reason to limit the number of client commands per one draw, I'd like to know it.


erwincoumans:
thanks, it was an accident indeed!

erwincoumans:
Actually, I put a breakpoint while running PhysicsClient/PhysicsServer and after a long while continued running. Then the timestep became huge, and the PhysicsServer 'stepSimulation' took forever. That's why I clamped to 10, which got accidently committed. Might be worth thinking about a way to clamp the stepSimulation to some reasonable amount for such debug sessions (1 or 2 seconds).

Issue #347, opened 2015-03-28
Assert/crash in btDbvt::collideOCL


antarespilot:
We've stumbled upon a bug in btDbvt::collideOCL. We're using the current master branch of bullet3 (currently at 29949b63ccd1ba7720fc9f4d09c5184d5e656ff1)

If running Bullet in debug mode (or at least when having DEBUG defined), we hit an assertion in btAlignedObjectArray:

```
/LinearMath/btAlignedObjectArray.h:178: T& btAlignedObjectArray<T>::operator[](int) [with T = int]: Assertion `n<size()' failed.
```

The assert occurs at line 1197 in btDbvt.h, at the memmove operation. The size of "stack" is 2, and the value of "j" is 1, meaning that when calling 

```c++
memmove(&stack[j+1],&stack[j],sizeof(int)*(stack.size()-j-1));
```

the first argument will go beyond the current size of "stack", thus triggering the assert. When running a release build, there is no assert, and everything seems to be working fine. Still there is something there that could write beyond the size of the "stack" array, which could cause other weird issues. 

Below is a complete "main.cpp" that reproduces the issue. This uses Bullet in double-precision mode. It does view frustum culling on a real-earth-sized world and is extracted from a larger application. 

I have not tried to reproduce this in single-precision-mode, but it seems like this issue has been around for quite a while, see http://bulletphysics.org/Bullet/phpBB3/viewtopic.php?t=9112
This guy does not mention anything about double-precision. 

I'm not sure what has to be done to fix this, as I am not familiar with exactly how the collideOCL function works. Maybe it's as easy as an additional check on the size of "stack"? 

```c++
#include <iostream>
#include <BulletCollision/BroadphaseCollision/btDbvt.h>

using namespace std;

#ifndef M_PI
#define M_PI		3.14159265358979323846
#endif

const double deg_to_rad = M_PI/180.0;
const double earthRadius    = 6378000.0; ///< Radius for spherical earth calculations.

/// Convert from spherical latitude/longitude to geocentric XYZ coordinates.
inline btVector3 geoToXYZ(double lat, double lon, double alt)
{
   btVector3 ret;

   double phi = lat * deg_to_rad;
   double lam = lon * deg_to_rad;
   double r   = earthRadius + alt;

   double sin_phi = sin(phi);
   double cos_phi = cos(phi);
   double sin_lam = sin(lam);
   double cos_lam = cos(lam);

   ret.setX(r * cos_phi * cos_lam);
[...]

PR #499, opened 2015-10-21
Dbvt memmove fix

antarespilot:
Hi Erwin,

I finally found time to look into the memmove issue described at https://github.com/bulletphysics/bullet3/issues/347

Here is my proposed fix. I've measured the performance using the google benchmark utility, and the memmove version is always significantly faster than the for-loop (magnitude of app. 4x or more). Of course, this is just an isolated test of the for-loop versus the if+memmove operation. 

I've also tested this memmove version against my applications that use the frustum tests, as well as small unit test applications that show that the two versions produce the same results. 

Basically the fix compared with the original memmove version is just an additional test to see if there are any bytes to move at all. 

Cheers,
Ola


antarespilot:
Oops far too many commits in this pull request, was only supposed to be a single one (381eee6)

erwincoumans:
sure, thanks for looking into this!

erwincoumans:
By the way, can you explain how you used the  google benchmark utility exactly?

antarespilot:
I'll write an explanation in the issue #347

PR #501, opened 2015-10-22
fix for broken memmove used in btDbvt.h, issue 347.

antarespilot:
Here is the isolated fix for issue #347. Basically all that was needed was to avoid the memmove if there were no bytes to move. This prevents accessing elements outside the size of "stack" which triggered the assert in btAlignedObjectArray. 

Cheers,
Ola


erwincoumans:
That looks good to me. At some stage it would be useful to create unit tests to catch those cases and avoid regression.

erwincoumans:
Thanks for the contribution by the way!

PR #502, opened 2015-10-22
add .gitignore file to filter build artifacts when using cmake on Linux

antarespilot:
I suggest adding a .gitignore-file. This is helpful especially when using git on the command line. This version will filter out the build-files that appear when running "cmake ." and "make" in the root of the repository.


scpeters:
I thought it was better practice to use a separate build folder:

~~~
mkdir build
cd build
cmake ..
make
~~~

benelot:
I would say so too. Generally you should make a separate build folder to build a cmake project. By that, you can easily delete the folder again or have multiple build folders with multiple cmake configurations such as Release and Debug compilations.

vacary:
In siconos, we use OutOfSourcesBuild.cmake to forbid building in the source directory. You have to switch the IN_SOURCE_BUILD variable if you want to change this default behavior.

antarespilot:
Sure I know about shadow builds :-) The .gitignore file won't prevent anyone from using them either. I just find this helpful in some cases. Just reject the pull request if it's not useful.

erwincoumans:
Thanks for the report, but indeed it is better to build out-of-source if that is the issue. Let's close the pull request then.

PR #503, opened 2015-10-23
Remove usage of deprecated `register` keyword

ajshort:
C++11 deprecated the register keyword, so don't use it if we're using C++11 or newer.


erwincoumans:
That sounds good to me. It look like it is barely used in the source tree anyway. Thanks for the contribution!

PR #518, opened 2015-11-08
MinGW fixes

d3x0r:
changes _WIN32 to _MSC_VER in a couple files
changes _MSVC_VER to _MSC_VER
## found some other things; guess I should just make another branch to include those.

Yes I know you don't have a 'mingw maintainer' and no I'm not volenteering; but when I come across them I can fix such simple things.


d3x0r:
Replaced with #1410

PR #519, opened 2015-11-08
Minor code fixes

d3x0r:
Ran across 3 duplicated lines.
Can add break in voronoi simplex : inSimplex test.


erwincoumans:
Both look valid fixes, thanks for the contribution!

Issue #520, opened 2015-11-08
Crash in  gjkepa2_impl::GJK::EncloseOrigin()


kingchurch:
Using HEAD of master as of Nov 7. Make a OSX 64-bit release build.  Run the "Rolling Friction"  demo and the Example Broswer crashed when the blocks hit the ground. See the backgract in the attached screenshot.

![screen shot 2015-11-08 at 11 31 24 am](https://cloud.githubusercontent.com/assets/5223822/11022144/5e66583c-860c-11e5-9bfa-3ac7e6994263.png)

erwincoumans:
Thanks for the report. I can't reproduce such crash, on Mac OSX, 64bit, release build with Rolling Friction demo. What build system did you use, premake or cmake, all details/flags etc, please.
What version of the compiler (Xcode and clang version) and operating system version?

erwincoumans:
I could reproduce it, it might be some clang issue related to namespaces.
Can you see if this pull request fixes the issue for you too?
https://github.com/bulletphysics/bullet3/pull/521

erwincoumans:
Clang got confused by duplicate class names in Bullet3OpenGL and BulletCollision. Fixed in latest revision, see https://github.com/bulletphysics/bullet3/pull/521

PR #521, opened 2015-11-08
clang gets confused/crashes when using the same namespace,

erwincoumans:
need to look if it is a real error or clang issue
should fix Issue 520
https://github.com/bulletphysics/bullet3/issues/520


kingchurch:
Thank you for the quick fix. I confirmed that the problem is fixed by the  latest commits

PR #522, opened 2015-11-09
Fix memory leak in PhysicsClientSharedMemory

takuyanakaoka:
I found a memory leak in PhysicsClientSharedMemory::processServerStatus() and fixed it using strdup().
I made a new function strDup() to wrap the difference between windows and others.


erwincoumans:
Thanks for the pull request. Let me have a look at it, I may move the strDup into an existing class.

PR #523, opened 2015-11-11
add deprecated methods to avoid breaking API

erwincoumans:


benelot:
Hi Erwin,

Since you start to improve the btMultiBody model, do you have any idea if you will improve the spherical joint usage and its joint limit capabilities soon? Is that on the roadmap of the near future?

Best,

P.S. I would love to help you on that, but I do not know enough about the codebase to help you improve it.

PR #526, opened 2015-11-16
Remove unused members (collision world)

d3x0r:
Remove unused member in hittest
Remove redundant if test (seq impulse solver) (redundant if isEnabled() )

Remove unused matrix copies. 
 Remove extra matrix copy and just use the original reference in computation.
orgInterpolationTrans[0/1] was entirely unused.
orgTrans[0/1] was just a copy of col[0/1]ObjWrap->getWorldTransform() and otherwise unused.


d3x0r:
Remove redundant if test (seq impulse solver) (redundant if isEnabled() )  
I actually deleted the first one; but the diff ended up deleting inbetween... so it's not as easy to see the delete of the empty if test...

d3x0r:
Almost 2 years and this is still outstanding?  
I know; it's probably not a huge issue what with compilers eliminating unused variables automatically and all... but still.

d3x0r:
Replaced with composite changes in #1410

PR #533, opened 2015-11-24
fix example issues

erwincoumans:


erwincoumans:
accidently merged a pull request, hitting the wrong button, github GUI :-1:

PR #544, opened 2015-12-04
cleanup main premake script

AndrewMeadows:
Cleaned up some bad indentation in the main **premake4.lua** script.

Also added some new premake options:

  --no-demos
  --no-test
  --no-bullet3
  --no-extras

The motivation for this work was to speed up complete rebuilds of the core Bullet2 libraries, by not bothering to build most of the other unnecessary stuff.

BTW, I noticed that the script indentation has a mixture of tabs and spaces.  I'm wondering what is the correct indentation coding style for Bullet.

### DIFF
--- a/build3/premake4.lua
+++ b/build3/premake4.lua
@@ -1,5 +1,5 @@
 
-    solution "0_Bullet3Solution"
+	solution "0_Bullet3Solution"
 
 	local osversion = os.getversion()
 	print(string.format(" %d.%d.%d (%s)",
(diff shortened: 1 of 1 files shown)


build3/premake4.lua
```diff
@@ -49,43 +49,55 @@
...
-		trigger = "without-gtest",
+		trigger = "no-test",
+		description = "Disable all tests"
+	}
+
+	newoption
+	{
+		trigger = "no-gtest",
 		description = "Disable unit tests using gtest"
 	}
```
AndrewMeadows: Note: the **--without-gtest** option was replaced with **--no-gtest**.  If this causes a problem we can roll this change back.

build3/premake4.lua
```diff
@@ -49,43 +49,55 @@
 
 --	--_OPTIONS["midi"] = "1";
 
-	newoption
-	{
-		trigger = "bullet2demos",
-		description = "Compile the Bullet 2 demos (Demo/Extra folder)"
-	}
```
AndrewMeadows: The **--bullet2demos** option was cruft: none of the other **premake4.lua** scripts were using it.

erwincoumans:
Thanks for the contribution!
Bullet uses tabs for indentation, but it is not always consistent for various (sloppy) reasons.
Here is the old code style document: https://docs.google.com/document/d/1u9vyzPtrVoVhYqQOGNWUgjRbfwfCdIts_NzmvgiJ144/edit?usp=sharing
In the future for Bullet 3.x we might switch to a different coding style, but Bullet 2.x will stick to this.

AndrewMeadows:
Here's a tip for those who don't know: add ?w=1 to the end of the github diff URL to make force it to ignore whitespace differences.

erwincoumans:
That ?w=1 is a handy option. It is still preferable to keep the whitespace fixes in a separate commit from the actual changes in the code.

PR #545, opened 2015-12-04
also allow premake to build Bullet-2.x shared libs (cmake already allows shared libs)

AndrewMeadows:
Adding premake projects to build shared versions of Bullet-2.x libs.

If this PR is worth merging, and you also want shared versions of Bullet-3 libs then let me know and I'll modify those premake scripts as well and add them to this PR.

### DIFF
--- a/build3/premake4.lua
+++ b/build3/premake4.lua
@@ -18,7 +18,8 @@
 		act = _ACTION
 	end
 
-	newoption {
+	newoption 
+    {
 		trigger     = "ios",
 		description = "Enable iOS target (requires xcode4)"
 	}
@@ -94,15 +95,24 @@
 	newoption
 	{
 		trigger = "no-bullet3",
-		description = "Do not build bullet3 libs"
+		description = "Don't build bullet3 libs"
 	}
 
-	configurations {"Release", "Debug"}
-	configuration "Release"
+	newoption 
+    {
+		trigger = "target-suffix",
+		value = "string",
+		description = "Use custom target suffix string"
+	}
+
+	configurations {"Release", "Debug", "ReleaseDLL", "DebugDLL" }
+
+	configuration { "Release", "ReleaseDLL" }
 		flags { "Optimize", "EnableSSE2","StaticRuntime", "NoMinimalRebuild", "FloatFast"}
-	configuration "Debug"
-		defines {"_DEBUG=1"}
+	configuration { "Debug", "DebugDLL" }
 		flags { "Symbols", "StaticRuntime" , "NoMinimalRebuild", "NoEditAndContinue" ,"FloatFast"}
+		defines {"_DEBUG=1"}
+	configuration{}
 
 	if os.is("Linux") or os.is("macosx") then
 		if os.is64bit() then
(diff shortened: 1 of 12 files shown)


AndrewMeadows:
I just realized that I'm not happy with the name of the output shared libraries.  I'm going to fight with premake to see if I can get it to do what I want.  Marking this PR as **DO NOT MERGE** until that fight is over.

AndrewMeadows:
Ok the names of shared libs are fixed.

erwincoumans:
I need to think about it, there are some issues. First of all, I prefer not to build DLL by default, and only make it an option that has to be turned on. 2nd, this is going to add maintenance, due to the duplication of all the projects. Did you consider using CMake instead? It has more features than premake and it already supports dynamic libraries.

Have you considered using CMake? Bullet's CMakeLists do support dynamic libraries already.

AndrewMeadows:
Yeah, the proliferation of default projects and bulk-copying of code sucks.  Let me see if I can clean it up and come up with something cleaner and easier to maintain.  I'm enjoying playing with **premake** and am motivated to learn more about it.  If I can't come up with something acceptable then I'll settle for using **cmake**.

AndrewMeadows:
Ok the premake scripts have been improved.  The shared libs are not build automatically, one must explicitly specify "config=releasedll64" at the **make** step.  I assume on windows one would select the correct config via the UI.

I also added a **--target-suffix** option so that it is possible to produce custom library names that don't have 'gmake' in them.

erwincoumans:
I just tried compiling the dll version of BulletCollision on Mac OSX, at it produces a linker error:
make config=releasedll64 BulletCollision

Linking BulletCollision
ld: warning: directory not found for option '-L/usr/lib64'
Undefined symbols for architecture x86_64:
  "_maxdot_large(float const*, float const*, unsigned long, float*)", referenced from:
      FindMaxSeparation(int*, btBox2dShape const*, btTransform const&, btBox2dShape const*, btTransform const&) in btBox2dBox2dCollisionAlgorithm.o
      btConvexHullShape::localGetSupportingVertexWithoutMargin(btVector3 const&) const in btConvexHullShape.o
      btConvexHullShape::batchedUnitVectorGetSupportingVertexWithoutMargin(btVector3 const*, btVector3*, int) const in btConvexHullShape.o
  
I guess we need to explicitly express dependencies (similar to  TARGET_LINK_LIBRARIES(BulletCollision LinearMath) in CMAKE) ?

erwincoumans:
Creating XCode project fails when applying this patch. I'll try Windows and report later.
It might be better to create either static or dynamic libraries, but not both at the same time.

./premake_osx xcode4
Project 'Bullet3Common' uses more than one target kind; not supported by Xcode

AndrewMeadows:
Ok, I'll to dust off the OSX partition of this laptop and try to build on that platform.

AndrewMeadows:
After banging on this a bit more I've concluded that it is impossible to do what we want... using premake4.  We would have to upgrade to premake5 or maybe something closer to the bleeding edge (to get access to **removeconfigurations**).  Let's not close this PR quite yet: I'll try to make it work and if I come up with a solution then we can decide whether to keep or discard.

AndrewMeadows:
I'm going to close this PR, since I haven't been able to get back to it in a reasonable amount of time.  cmake is working fine.

PR #547, opened 2015-12-07
fix cmake to succeed with -DBUILD_BULLET3=0

AndrewMeadows:
Was playing with cmake and discovered that it would not build on linux with the -DBUILD_BULLET3=0 option setting, so I decided to fix it.


CMakeLists.txt
```diff
@@ -279,7 +279,9 @@ ENDIF()
 
 set (BULLET_CONFIG_CMAKE_PATH lib${LIB_SUFFIX}/cmake/bullet )
 list (APPEND BULLET_LIBRARIES LinearMath)
-list (APPEND BULLET_LIBRARIES BulletInverseDynamics)
+IF(BUILD_BULLET3)
+    list (APPEND BULLET_LIBRARIES BulletInverseDynamics)
+ENDIF(BUILD_BULLET3)
```
AndrewMeadows: **BulletInverseDynamics** depends on **Bullet3Common**.

erwincoumans:
Good point, we should compile one of the CI projects using this flag (travis and/or appveyor) to catch such issues automatically in the future.

erwincoumans:
Thanks for the contribution!

PR #548, opened 2015-12-07
fix whitespace formatting in cmake files

AndrewMeadows:
Since I spotted some improper formatting in the **CMakeList.txt** files I decided to clean it up and submit a formatting-only PR.


erwincoumans:
the ?w=1 helped reviewing this, thanks again!

PR #573, opened 2016-03-01
Remove deprecated reference to _STATIC_CPPLIB VS2015+

d3x0r:
which causes duplicate…… symbol conflicts when linking now, because those get exposed and cause linker failures now.
Perhaps there's another way; but this works.
App Example Browser builds without issue...


d3x0r:
replace with composite changes #1410

PR #576, opened 2016-03-03
The 'static' storage-class specifier is not supported in OpenCL.

hujiajie:
This is documented in Section 6.8 (g) in OpenCL 1.1 specification.


hujiajie:
@erwincoumans Please take a look

erwincoumans:
It looks good, but could you please rename the 'STATIC' to 'B3_STATIC'?

hujiajie:
Done

erwincoumans:
thanks for the report and contribution!

PR #577, opened 2016-03-04
fix for issue in getAngleShortestPath()

MichaelKorn:
Testing squared norm for <0 does not make any sense. In Addition giving the range of the return values is a great help to understand the descriptions.


erwincoumans:
Good point, that doesn't make any sense to check a squared norm < 0. Thanks for the report and fix!

PR #580, opened 2016-03-11
Fix ternary selection operator on float scalars in OpenCL kernels.

hujiajie:
According to Section 6.3(i) of the OpenCL 1.1 specification, the first
expression of the operator cannot be a float.


erwincoumans:
looks good to me, thanks for the help/patch!

PR #581, opened 2016-03-11
Use higher precision GJK/EPA collision tolerances in double precision mode. Fix Lua example

erwincoumans:
See also https://github.com/billhoffman/bullet-pod/blob/ac6aae3e3ee8137de484471094a65bfe166898b4/bullet_gjk_accuracy_patch.diff

### DIFF
--- a/src/BulletCollision/NarrowPhaseCollision/btGjkEpa2.cpp
+++ b/src/BulletCollision/NarrowPhaseCollision/btGjkEpa2.cpp
@@ -41,21 +41,38 @@ namespace gjkepa2_impl
 
 	/* GJK	*/ 
 #define GJK_MAX_ITERATIONS	128
-#define GJK_ACCURARY		((btScalar)0.0001)
-#define GJK_MIN_DISTANCE	((btScalar)0.0001)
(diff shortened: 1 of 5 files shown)


src/BulletCollision/NarrowPhaseCollision/btGjkEpa2.cpp
```diff
@@ -41,21 +41,38 @@ namespace gjkepa2_impl
 
 	/* GJK	*/ 
 #define GJK_MAX_ITERATIONS	128
-#define GJK_ACCURARY		((btScalar)0.0001)
-#define GJK_MIN_DISTANCE	((btScalar)0.0001)
-#define GJK_DUPLICATED_EPS	((btScalar)0.0001)
+
+#ifdef BT_USE_DOUBLE_PRECISION
+	#define GJK_ACCURARY		((btScalar)1e-12)
```
sherm1: Would this be a good time to fix the spelling? `GJK_ACCURARY` -> `GJK_ACCURACY` ?

erwincoumans: Good point Sherm. There are other typos in the code base, but it is good to address them.

PR #592, opened 2016-04-09
Update README.md

achlubek:


erwincoumans:
Thanks fix was manually applied. Thanks!

PR #609, opened 2016-04-27
optimize CCD against btCompoundShapes with dynamicAabbTree

AndrewMeadows:
These are some changes we've been using for several months and are very happy with.  I'm finally getting around to submitting as a proper PR.

The motivation was to solve some performance problems we noticed when trying to do CCD against large btCompoundShapes.  We saw the **TODO** comment in the code about using the shape's "dynamic aabb tree" to speed up collision queries and decided to implement it.

### DIFF
--- a/src/BulletCollision/CollisionDispatch/btCollisionWorld.cpp
+++ b/src/BulletCollision/CollisionDispatch/btCollisionWorld.cpp
@@ -790,124 +790,122 @@ void	btCollisionWorld::objectQuerySingleInternal(const btConvexShape* castShape,
 				}
 			}
 		} else {
-			///@todo : use AABB tree or other BVH acceleration structure!
 			if (collisionShape->isCompound())
 			{
-                struct	btCompoundLeafCallback : btDbvt::ICollide
-                {
-                    btCompoundLeafCallback(
-                                           const btCollisionObjectWrapper* colObjWrap,
-                                           const btConvexShape* castShape,
-                                           const btTransform& convexFromTrans,
-                                           const btTransform& convexToTrans,
-                                           btScalar allowedPenetration,
-                                           const btCompoundShape* compoundShape,
-                                           const btTransform& colObjWorldTransform,
-                                           ConvexResultCallback& resultCallback)
-                    : 
-                      m_colObjWrap(colObjWrap),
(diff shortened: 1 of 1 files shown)


src/BulletCollision/CollisionDispatch/btCollisionWorld.cpp
```diff
@@ -790,23 +790,50 @@ void	btCollisionWorld::objectQuerySingleInternal(const btConvexShape* castShape,
 				}
 			}
 		} else {
-			///@todo : use AABB tree or other BVH acceleration structure!
```
AndrewMeadows: Implementing this **TODO**.

erwincoumans:
Great, let me have a look and merge it soon.
Thanks for contributing!

erwincoumans:
Is there still some option to use the old (inefficient) path, for testing/comparing the new implementation?

AndrewMeadows:
The new optimization is bypassed for `btCompoundShapes` that don't have an `m_dynamicAabbTree`, however the dynamic tree feature is enabled by default, which means any A/B example test under the current state of this PR would require some code changes and a recompile.

The other spots in the code that pivot their logic on the existence of `btCompoundShape::m_dynamicAabbTree` don't have global test configurations.  For example, the code in `btCompoundCollisionAlgorithm::processCollision()` which was, I think, the pattern followed when making this new optimization.

If we were to add a `btCollisionWorld::m_useCompoundDynamicAabbTreeForObjectQuery` data member it would be useful for evaluating this PR, but probably wouldn't be something wanted for the long term.  Who would use it once the optimized method is verified?  Nevertheless I'm willing to add it if it helps move things forward.  We could remove that test hack later once we're confident this is working correctly.

erwincoumans:
We can have the optimization enabled by default, but add some option somewhere to disable it, just in case it causes issues. That way, we can ask people who have issues (unlikely) to try that option and see if it fixes their issue. Ideally we create a unit test comparing the two (I started with a few ones, and the build system supports automatically running them in CI)... Anyway, if you are too busy (which I understand) I may find some time next week to do this.

erwincoumans:
I'll pull the request and write a test soon. Thanks for the contribution!

PR #612, opened 2016-05-01
Improve HelloWorld

cirosantilli:
- explain in comments the shape of the ground and better group that code
- give enough time for the sphere to hit the ground
- don't ask for confirmation to exit, it's annoying


erwincoumans:
Looks good to me, thanks for the pull request!

PR #621, opened 2016-05-08
Add missing m_callback call in SliderMove()

### LINKED ISSUE #613: Slider changed callback in ExampleBrowser does not work
I tried to use it with a matching function, but it never got called. I looked into the example browser and only found an example that uses the button callback. Older examples used the slider change not event based, but updated within stepSimulation(). If it is working, can someone provide a simple example how it works? If not, can we fix this?

benelot:
The MySliderEventHandler is missing the m_callback whereas the SliderParams contains the m_callback field. Therefore, the m_callback is never called if the slider position is changed. This commit adds the missing m_callback field and call in SliderMove().
Fixes #613.


erwincoumans:
thanks for the contribution!

Issue #626, opened 2016-05-13
Cflags in bullet.pc are incorrect when built with CMAKE_INSTALL_PREFIX


simotek:
`bullet.pc.in` contains 

```
 Cflags: @BULLET_DOUBLE_DEF@ -I@CMAKE_INSTALL_PREFIX@/@INCLUDE_INSTALL_DIR@ -I@CMAKE_INSTALL_PREFIX@/include
```

while `BulletConfig.cmake.in` contains

```
set ( BULLET_INCLUDE_DIR  "@INCLUDE_INSTALL_DIR@" )
set ( BULLET_INCLUDE_DIRS "@INCLUDE_INSTALL_DIR@" )
```

when building with the following flags taken from the openSUSE build

```
[...]

PR #627, opened 2016-05-13
 Fix bullet.pc when built with CMAKE_INSTALL_PREFIX #626

simotek:
This makes the macro usage consitent between bullet.pc.in and
BulletConfig.cmake.in, previously if the CMAKE_INSTALL_PREFIX
was set you would end up with -I/usr/usr/include/bullet rather then
-I/usr/include/bullet


erwincoumans:
Thanks for the help! I trust you made the right change, as I don't use or test system-wide install.

erwincoumans:
Please provide a patch and if it doesn't break the build I'll apply it.

PR #630, opened 2016-05-16
add sudo make install

zdx3578:


zdx3578:
add sudo make install

erwincoumans:
In general I would rather not recommend to install Bullet system-wide, and rather copy the Bullet source code into your own project. It is a longer discussion why and how and it depends on the project as well. We can add some comment that it is possible to make install but not recommended.

Issue #130, opened 2014-03-30
Complete the Bullet C-API (was googlecode  Issue 43)


erwincoumans:
See https://code.google.com/p/bullet/issues/detail?id=43&colspec=Modified%20ID%20Type%20Stars%20Status%20Owner%20Summary

chino:
I really hope bullet 3 api will have a supported c api.

Lack of a stable supported fully featured wrapper has been biggest slow down with using Bullet in my non c++ projects.

Is the bullet 3 api going to be able to downgrade back to the cpu if the video card card isn't capable enough?

Thanks for all the great work.

AndresTraks:
I've also been working on a C API as part of a P/Invoke .NET wrapper. It's pretty much complete for bullet 2 and I've planned to start with bullet 3 this summer. https://github.com/AndresTraks/BulletSharpPInvoke/tree/master/libbulletc

It's a bit different though as it's first and foremost designed to be an intermediate layer between bullet and other higher level wrappers instead of being a natural C API. The code is completely independent from bullet, so nothing needs to be modified in bullet itself.

Here's the idea of how it currently works:
```cpp
//btCollisionShape_wrap.h
#ifndef BT_COLLISION_SHAPE_H
#define btCollisionShape void
#endif
EXPORT void btCollisionShape_delete(btCollisionShape* obj);
```
```cpp
//btSphereShape_wrap.h
#ifndef BT_SPHERE_MINKOWSKI_H
#define btSphereShape void
#endif
EXPORT btSphereShape* btSphereShape_new(btScalar radius);
EXPORT btScalar btSphereShape_getRadius(btSphereShape* obj);
```
```cpp
//btCollisionShape_wrap.cpp
#include <BulletCollision/CollisionShapes/btCollisionShape.h>
#include "btCollisionShape_wrap.h"

void btCollisionShape_delete(btCollisionShape* obj)
{
	delete obj;
}
```
```cpp
//btSphereShape_wrap.cpp
#include <BulletCollision/CollisionShapes/btSphereShape.h>
#include "btSphereShape_wrap.h"

btSphereShape* btSphereShape_new(btScalar radius)
{
	return new btSphereShape(radius);
}

btScalar btSphereShape_getRadius(btSphereShape* obj)
{
	return obj->getRadius();
}
```
```cpp
//bulletc.h
#include "btCollisionShape_wrap.h"
#include "btSphereShape_wrap.h"
```
```cpp
//test.cpp
[...]

PR #636, opened 2016-05-21
Basic infrastructure C API

Tal500:
A begining for solution to Bullet C API . See: https://github.com/bulletphysics/bullet3/issues/130
I wish that Bullet will have an embedded C API. (i.e same header files, same names)

What was mainley created is an C API for LinearMath, which I believe that is the basic infrastructure for C API, since LinearMath is the basic infrastructure for Bullet API.

See previous link for more details and discussions.

Notice that I focused only on Bullet 2 API.


benelot:
Why is one of the major contributions of this pull request that you rename the btVector3 to btVector? That sounds somewhat arbitrary if not less descriptive.

erwincoumans:
I would like to move away from the Bullet 2.x API, and create a new C-API that is more physics-engine agnostic. At the moment, Bullet 2.x doesn't make a difference between public API and internal API, and for a new API I rather start high-level. See the new b3RobotSimAPI.h
https://github.com/erwincoumans/bullet3/blob/master/examples/RoboticsLearning/b3RobotSimAPI.h
This wraps the https://github.com/erwincoumans/bullet3/blob/master/examples/SharedMemory/PhysicsClientC_API.h

Tal500:
Sorry for the late response.

benelot:
The reason I chose to change btVector3 to btVector, is because the naming seams to be wrong in Bullet itself.
Bullet use btVector3 both for 3D and 4D vector. btVector4 derive btVector3, and it's bad, IMHO, for 2 reasons:
1. btVector4 is not btVector3 (against the IsA rule of inheritance)
2. Inheritance of such elementary type is one of the obstacles for C API

Therefore, the first thing I should have done is to change btVector3 to btVector, and then tell other classes(btVector3&btVector4) to derive from it.

If you see #130, it will be much more clear why I chose to do so. (also, see the next lines)

erwincoumans:
Personally, I don't know whether public and internal API distinction is good, but it shouldn't be the anyway related to the infrastructure for C API.

As you saw what I posted in #130, I claim that the strengths of Open Dynamic Engine(ODE) is full support of C API. I think that Bullet shouldn't create C API like they did(C only support), but more friendly C&C++ integration.

I think that the most important part for good C API is 100% same structs for C&C++ for **LinearMath**. In C++ however, those structs will have additional possibilities for development experience. (constructors, functions and operators)

For non LinearMath, the rules I proposed is much less stricter. The reason for that is that there classes usually represented as a pointer, and not by a value(unlike LinearMath).

Do you think Bullet can go that way, or the current aim is _low level C++& high level C_?

rweichler:
I'm not too experienced with C++, but wouldn't it make more sense for the math stuff to just be structs, instead of classes, if the goal is to have transparent types in C?

Tal500:
rweicher, I totally agree with you, and this is exactly the point of my pull request.

rweichler:
Hm, I think it might be a lost cause if backwards-compatibility is a priority. I'm sure there are some people out there who are subclassing Vector3 and friends for whatever reason.

At this point it's probably better to just treat the existing C++ API as private and wrap that, it'd be easier to write and @erwincoumans would be more likely to accept it since he wants a distinction between a private and public API anyway. The only sucky part is the small performance loss. But I think not much else can be done at this point without a major rewrite

erwincoumans:
There is no plan to change the Bullet 2.x C++ API, I prefer to avoid changes that break backward compatibility for the old 2.x C++ API.

>> or the current aim is low level C++& high level C?

Yes, the current aim is the [abstract C-API ](https://github.com/erwincoumans/bullet3/blob/master/examples/SharedMemory/PhysicsClientC_API.h)and low level C++ backend that can be replaced. We are doing a lot of our recent experiments in pybullet, the Python API on top of this C API.

PR #649, opened 2016-06-09
render image with ability to set pixel resolution and initial camera

hellojas:
- fix pybullet_renderImage method to set array with RGBA values (4 channels instead of 3)

- add pybullet_internalSetMatrix which takes a pyobj and float[16] to set both view and projection matrices
- added ability to call renderImage() to set pixel resolution and optionally set initial camera position; can call renderImage(w, h) or renderImage(w, h, float[16], float[16]); both have been tested with success

- TODO: rendered image is still upside down


erwincoumans:
Thanks for the contribution. There is some conflict in one of the files (checks failed) and it is difficult to review because of all the merge and white-space changes etc. There is a way to tell git to ignore whitespace changes.
Could you start from a clean slate using latest revision and re-apply your changes?

hellojas:
Okay, will take a look! Sorry my editor uses uncrustify to automatically beautify C, I'll turn it off for this time.

hellojas:
Oh! Actually I didn't finish merging another file, I'm pretty sure this is causing the error:
/Users/hellojas/Development/bullet3/examples/SharedMemory/PhysicsClientExample.cpp:632:1: error:
      version control conflict marker in file

hellojas:
github diff without whitespace:

https://github.com/bulletphysics/bullet3/pull/649/files?w=1

erwincoumans:
Using the link with ?w=1 hides the whitespace from view, but all the whitespace changes are still in the pull request. This is not going to work. Can you please re-apply the changes in pybullet.c from a prestine latest version, without the whitespace changes, and then create a new pull request without the entire history of merges etc?

PR #650, opened 2016-06-09
Fixed uninitialised info2->findex for btGeneric6DofSpring2Constraint.

jbrd:
Hi,

We recently encountered a bug with `btGeneric6DofSpring2Constraint` in which accumulated impulse was not being calculated properly. Upon further investigation we realised that `findex` in the info2 structure was not being initialised for this constraint type. Here is a patch that addresses the problem, by re-purposing the 'rotational' argument on `get_limit_motor_info2` to pass an findex value instead. Let me know what you think,

Thanks,

James Bird
Double Negative.

### DIFF
--- a/src/BulletDynamics/ConstraintSolver/btGeneric6DofSpring2Constraint.cpp
+++ b/src/BulletDynamics/ConstraintSolver/btGeneric6DofSpring2Constraint.cpp
@@ -537,7 +537,7 @@ int btGeneric6DofSpring2Constraint::setLinearLimits(btConstraintInfo2* info, int
 			{
 				rotAllowed = 0;
 			}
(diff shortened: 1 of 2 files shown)


erwincoumans:
findex is not used for btTypedConstraint, and the findex pointer is initialized to -1.
This is clearly not good, so better to just remove findex instead.

jbrd:
Ah yes, sorry about that. We're still using a version of the solver that uses findex to compute the accumulated impulse, but I see that has since been refactored. I should have checked that! I will close this pull request, and submit another one soon that just removes findex. Thanks.

erwincoumans:
Sure. It should be fixed in this pull request: https://github.com/bulletphysics/bullet3/pull/654

PR #652, opened 2016-06-10
render image with ability to set pixel resolution and initial camera position

hellojas:
- call renderImage without args gives default resolution and camera position
- optionally call renderImage with just resolution or resolution and view/projection matrices (args size 2 or 4)
- successfully tested renderImage(), renderImage(w,h), renderImage(w,h,view[16],projection[16])
- added function comments in pybullet
- refactored renderImage with internalSetMatrices function to set view and projection matrix

- TODO: image is right side up but head is cut off
- TODO: depth values are now returned but need to be validated (check if they make sense)


erwincoumans:
thanks for re-submitting the patch with cleanup!

PR #660, opened 2016-06-16
Set view and projection matrices from camera and frustum parameters.

YunfeiBai:


erwincoumans:
That looks good to me, thanks for the contribution!

PR #668, opened 2016-06-23
Apply gravity to each simulation sub step

ktfh:
btDiscreteDynamicsWorld::applyGravity() is only being executed once but needs to be executed for each simulation sub step. Either inside the for loop or inside btDiscreteDynamicsWorld::internalSingleStepSimulation()

### DIFF
--- a/src/BulletDynamics/Dynamics/btDiscreteDynamicsWorld.cpp
+++ b/src/BulletDynamics/Dynamics/btDiscreteDynamicsWorld.cpp
@@ -447,12 +447,9 @@ int	btDiscreteDynamicsWorld::stepSimulation( btScalar timeStep,int maxSubSteps, 
 
 		saveKinematicState(fixedTimeStep*clampedSimulationSteps);
 
-		applyGravity();
-
-
-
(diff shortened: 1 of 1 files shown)


ktfh:
After more thought, it might be even better to remove the applyGravity call entirely and have users call applyGravity inside m_internalPreTickCallback if they want gravity? Having a variable timestep causes simulations to diverge at the moment so some change needs to be made.

erwincoumans:
The gravity force is added before all the substeps, and the force remains active over all substeps. After the substeps, all forces are cleared. This should work fine. If you prefer, you can also set the gravity to zero, and manually apply external forces in the internal pretick callback. No bug/change is necessary in the API, so let's close the issue.

ktfh:
I see my problem is that custom behaviors that apply force in the pre tick callback are executed an arbitrary number of times with a variable time step, applyGravity and clearForces are only executed once. I can work around this by making my pre tick callback clearForces then applyGravity then do its thing like btRigidBody->applyForce(customForce, position). I thought it would be more elegant if stepSimulation executed applyGravity and clearForces for each simulation sub step though. It would eliminate one way of screwing up when implementing a variable time step.

PR #678, opened 2016-06-30
Made Collision Shape names consistent.

### LINKED ISSUE #677: Inconsistent Collision Shape Naming
I'm using `btCollisionShape::getName()` in my GUI to determine the type of collision shape.

[`btSphereShape`](https://github.com/bulletphysics/bullet3/blob/master/src/BulletCollision/CollisionShapes/btSphereShape.h#L56) returns **"SPHERE"** (all caps), while [`btBoxShape`](https://github.com/bulletphysics/bullet3/blob/master/src/BulletCollision/CollisionShapes/btBoxShape.h#L274) returns **"Box"** (title case). Not a huge issue, but any reason for the discrepancy?

winduptoy:
Fixes #677.

### DIFF
--- a/src/BulletCollision/CollisionShapes/btBvhTriangleMeshShape.h
+++ b/src/BulletCollision/CollisionShapes/btBvhTriangleMeshShape.h
@@ -75,7 +75,7 @@ public:
 	void	partialRefitTree(const btVector3& aabbMin,const btVector3& aabbMax);
 
 	//debugging
-	virtual const char*	getName()const {return "BVHTRIANGLEMESH";}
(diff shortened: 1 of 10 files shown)


erwincoumans:
"UniformScaling" doesn't make much sense, without the 'Shape' part.
Once we start changing it like this, I'd rather have the original C++ class name.
btCapsuleShape, btBoxShape, btSphereShape, btUniformScalingShape.
How about that?

benelot:
I thought so too when I looked at the changes. The names of the classes
sound better.

PR #679, opened 2016-06-30
pkg-config: Fixing regression introduced by e46b7b2

jbohren:
PR #626 introduces a regression for projects which rely on `$PREFIX/include/bullet` being in the list of include paths. The `@INCLUDE_INSTALL_DIR@` variable is always set to just `include/bullet` (a relative path) but pkg-config paths should be absolute. This commit effectively reverts @scpeters patch: d7131e9bb0df5163050f046bc432148fbd23befd

If someone relies on this pkg-config file from a CMake project, they will end up with include paths which are relative to the calling CMakeLists.txt file, which is definitely the incorrect behavior.

This PR reverts the bug introduced by e46b7b2 / #626.

@simotek @erwincoumans


erwincoumans:
I'm not using or testing pkg-config and accept your contribution. Let's hope it converges to what most people want.

PR #685, opened 2016-07-08
There was a typo I came across when cross compiling

ComradeKeys:
TLDR; I fixed a typo for the windows build.
 - bkeys


erwincoumans:
We could fix this, but it makes me curious why you bother?
Is there a case sensitive windows compiler?

scpeters:
seems to be an issue with cross-compiling with mingw:

* http://stackoverflow.com/a/15466951

erwincoumans:
ah, didn't think of people cross-compiling windows executables. ok, then, let's apply it.
Thanks for the fix!

ComradeKeys:
Yee, thanks for taking it.

PR #687, opened 2016-07-09
There were some casts in the modified files that were very sketchy, i…

ComradeKeys:
…nsomuch that they were causing compiler errors for me on windows. (I am done cross compiling for now)

Basically the standard libraries have a function called atol that will convert the char to a long in a much cleaner way than an old style cast will. I went through and replaced them so I would stop getting the compiler errors; this is also most likely a better practice than old style casting, since it is in the standard libraries

### DIFF
--- a/Extras/Serialize/BulletFileLoader/bDNA.cpp
+++ b/Extras/Serialize/BulletFileLoader/bDNA.cpp
@@ -391,7 +391,7 @@ void bDNA::init(char *data, int len, bool swap)
 
 	
 	{
-		nr= (long)cp;
+    	        nr= atol(cp);
 	//long mask=3;
 		nr= ((nr+3)&~3)-nr;
 		while (nr--)
@@ -427,7 +427,7 @@ void bDNA::init(char *data, int len, bool swap)
 	}
 
 {
-		nr= (long)cp;
+                nr= atol(cp);
 	//	long mask=3;
 		nr= ((nr+3)&~3)-nr;
 		while (nr--)
--- a/Extras/Serialize/BulletFileLoader/bFile.cpp
+++ b/Extras/Serialize/BulletFileLoader/bFile.cpp
@@ -461,7 +461,7 @@ void bFile::swapDNA(char* ptr)
 
 	
 	{
-		nr= (long)cp;
+	  nr= atol(cp);
 	//long mask=3;
 		nr= ((nr+3)&~3)-nr;
 		while (nr--)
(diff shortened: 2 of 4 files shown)


erwincoumans:
a cast is totally different from a string to long conversion, this will break the code.

What compiler+version generates errors? Or did you force the compiler to treat warnings to errors?

ComradeKeys:
I made no changes to the build system, so the compiler is as strict as it is by default.
I am compiling on Windows 10
I am using the Mingw-w64 compiler which can be found here
http://mingw-w64.org/doku.php
The version is 4.0, but it uses GCC 5.3 I believe

erwincoumans:
What is the warning/error?

erwincoumans:
Unsurprisingly, the tests fail, a long cast is not the same as atol.
char* cp = "1";
long nr = (long)cp;//nr != 1
long nr2 = atol(cp);//nr = 1

On Windows, MinGW is not supported/tested, only Visual Studio is, but if you have a better fix, I consider applying it.

ComradeKeys:
Alright I did not figure this would break the code so good thing you caught that. I was hoping that would work since that is the only visible barrier to supporting GCC on windows, let me see what I can come up with

ComradeKeys:
selection_042
Here is the error I am receiving with an untouched version of Bullet

erwincoumans:
That is the error of your modified version. Can you revert back to the original code (atol) and report that error/warning, if any?

ComradeKeys:
No, this version is fresh off of the current master (notice the (long) instead of the atol), my modified version actually reports no errors at all whenever it is finished compiling however I noticed the Example Browser did not run correctly after compilation

erwincoumans:
I already realized this and removed this comment. Please hold on, I'll have a look at it.

ComradeKeys:
Alright, I am also going to look around for a solution to this.

ComradeKeys:
http://stackoverflow.com/questions/21574449/why-is-cast-from-x-to-y-loses-precision-a-hard-error-and-what-is-suitable
This article may be helpful, the issue is I am pretty sure you would lose more precision if you were to convert it into an intptr_t but it may lead us in the right direction

erwincoumans:
This is about aligning a pointer to 4-byte boundary, after serialization. Bullet already has a helper for this, but this very old code hasn't been updated. I did this in my current work-in-progress, so you can close this pull request. FYI:

remove these blocks
{
		nr= (long)cp;
	//	long mask=3;
		nr= ((nr+3)&~3)-nr;
		while (nr--)
		{
			cp++;
		}
	}

by this line
cp = btAlignPointer(cp,4);
(or cp = b3AlignPointer(cp,4); for newer Bullet 3 code (which is not used by most yet)

ComradeKeys:
Alright thanks for taking the time to look at this.

PR #688, opened 2016-07-10
Suppress compiler warnings

hujiajie:
Suppress several warnings about unused variables.


erwincoumans:
Thanks for the help. I'd like to remove all warning before the next release, so all bits help.

PR #693, opened 2016-07-12
[WIP] Suppress compiler warnings.

hujiajie:
Disable a few Visual Studio warnings:

C4244: conversion from 'type1' to 'type2', possible loss of data
C4267: conversion from 'size_t' to 'type', possible loss of data

Considering the GCC/Clang counterpart '-Wconversion'
(https://gcc.gnu.org/wiki/NewWconversion) is not covered by '-Wall', it
seems to make more sense to disable these two warnings in Visual Studio,
rather than fixing these warnings with hundreds of static_cast<> in the
source code.


erwincoumans:
looks good to me, thanks for the contribution!

PR #694, opened 2016-07-12
fix for vertically flipped depth buffer in TinyRenderer::render

matpalm:
thanks for the snippet erwin, all works fine.


erwincoumans:
thanks for the pull request and trying out the 'flip' :-)

matpalm:
and congratulations erwin, you passed the interview!

PR #700, opened 2016-07-15
[WIP] Suppress compiler warnings.

hujiajie:
Suppress Visual Studio warning C4018: '<': signed/unsigned mismatch.


erwincoumans:
Thanks again! The unsigned/signed int is a can of worms, that was introduced by size_t. I prefer to keep signed int wherever possible, and not convert things to unsigned int. I'll have a look at it later.

PR #707, opened 2016-07-20
fix error msg when loading sdfs

matpalm:


erwincoumans:
thanks for fixing the copy/paste bug

PR #708, opened 2016-07-20
GetJointInfo used linkIndex param, switched to jointIndex for naming consistency

matpalm:
( i should probably batch these little ones up sorry... )


erwincoumans:
good point, thanks for the fix.

PR #715, opened 2016-07-25
Add inverse kinematics example with implementations by Sam Buss.

erwincoumans:
Uses Kuka IIWA model description and 4 methods:
Selectively Damped Least Squares,Damped Least Squares,
Jacobi Transpose, Jacobi Pseudo Inverse
Tweak some PD values in Inverse Dynamics example and Robot example.


BluePrintRandom:
are you making a walking radoll rig inside bullet :D

PR #717, opened 2016-07-26
update pybullet_setJointMotorControl with seperate args for targetPosition, targetVelocity, kp and kd

matpalm:
( based on yesterday's b3RobotSimAPI.cpp change )

### DIFF
--- a/examples/pybullet/pybullet.c
+++ b/examples/pybullet/pybullet.c
@@ -294,27 +294,29 @@ pybullet_resetSimulation(PyObject* self, PyObject* args)
 
 static PyObject* pybullet_setJointMotorControl(PyObject* self, PyObject* args)
 {
-    //todo(erwincoumans): set max forces, kp, kd
+    //TODO(matkelcey): should this just be three methods?
 
     int size;
     int bodyIndex, jointIndex, controlMode;
     double targetValue=0;
+    double targetPosition=0;
+    double targetVelocity=0;
     double maxForce=100000;
-    double gains=0.1;
+    double kp=0.1;
+    double kd=0.1;
     int valid = 0;
-     
-     
+
     if (0==sm)
     {
         PyErr_SetString(SpamError, "Not connected to physics server.");
         return NULL;
     }
-    
+
     size= PySequence_Size(args);
-    
     if (size==4)
     {
-        
+        // for CONTROL_MODE_VELOCITY targetValue -> velocity
+        // for CONTROL_MODE_TORQUE targetValue -> force torque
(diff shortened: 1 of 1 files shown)


examples/pybullet/pybullet.c
```diff
@@ -357,27 +371,26 @@ static PyObject* pybullet_setJointMotorControl(PyObject* self, PyObject* args)
...
+
         switch (controlMode)
         {
             case CONTROL_MODE_VELOCITY:
             {
-				double kd = gains;
                 b3JointControlSetDesiredVelocity(commandHandle, info.m_uIndex, targetValue);
-                b3JointControlSetKd(commandHandle,info.m_uIndex,kd);
-                b3JointControlSetMaximumForce(commandHandle,info.m_uIndex,maxForce);
+                b3JointControlSetKd(commandHandle, info.m_uIndex, kd);
```
matpalm: oh wait, with CONTROL_MODE_VELOCITY kd isn't configurable. more strongly thinking that this being three methods would be much cleaner; there's a mix of args that are/aren't use by the different control modes and most of this code is checking the combos...

erwincoumans:
Thanks for the patch, I'll look at it.
 //TODO(matkelcey): should this just be three methods?
One idea is to use optional named arguments rather than additional methods.

matpalm:
i think the trouble is it's more than named args with sensible defaults; some args just aren't applicable to some control modes.... e.g. CONTROL_MODE_TORQUE only uses the target & doesn't care about max force, kd or kp.

erwincoumans:
This is mainly a documentation issue: pybullet follows the underlying C-API. Arguments that are not applicable for a certain mode are ignored. It is possible to protect the user, report errors/warnings and add convenience functions in object-oriented wrappers around pybullet,

matpalm:
ah yes, true. following the c-api is an important point that had slipped my mind...

erwincoumans:
Once we introduce named arguments, we can get rid of a lot of clutter that check for the number of args etc. It will break the API though, but at this still it is still very experimental anyway.

erwincoumans:
The default kd when using CONTROL_MODE_VELOCITY should be 1.0, not 0.1, when not specified. 

It is not due to this pull request, but there is an issue with defaults values not overridden by the user. When the user does NOT explicitly override the kd/kp values, pybullet should not come up with some default, but it should NOT call the function to override it. That way, the defaults are maintained in one place, in the C-API. So we need to fix that soon.

erwincoumans:
Thanks for the help, and hopefully we sort out this named arguments stuff soon!

PR #720, opened 2016-07-28
Fix kinematic character controller

MiCroN3000:
Ok, this contains the kinematic character controller with various fixes and a few features like, being able to set any vector for gravity/up (not just for the x, y or z-axis), possibility to use collision masks, jumping into a certain direction, angular & linear velocity, angular & linear damping.

I think the controller is quite hard to read and I worry that it is build on hacks because it is hard to read. It feels like I beat it into submission, but it does seem to work. My intent was to make the controller a bit more generic, so it behaves a bit more as a rigid body, hence some of the features I added. I think that some behaviors are just too limited. For example, how would one implement a double jump if the jump method always checks if it is on the ground first. Or why can one set a gravity vector, but only the x, y or z-axis is is used. Things like that are what I tried to improve here, they fit my needs better, but I hope they are also an improvement for Bullet.


erwincoumans:
Thanks for the contribution. As there hasn't been any development in this area, I'll accept it. It would be great if we can port the old CharacterDemo into the new Example Browser structure. If you have any time for that, it would be appreciated!

PR #721, opened 2016-07-28
fix warning: extra tokens at end of #endif directive

matpalm:


erwincoumans:
Yes, that should be fixed indeed, thanks!

PR #723, opened 2016-07-31
Simple Neural Network 3D Walkers example

benelot:
#  tl;dr Check summary comments below
Hello!

I created a basic example of walkers that get a random Neural network initialization that enables them to walk (or not). The walker's body parts all have sensors, which get linked to the joint motors of each walker by randomized connection weights, which leads to non-random movement behavior. My next ideas would be to implement an optimization strategy like evolution (hence the folder name evolution) to find a fit walker. Alternatively, I could look into OpenGym-like Deep Reinforcement learning to make it work with Tensor flow. Any suggestions? I just wanted to publish early to have a basis for discussion before I go too far into the details of either one.


erwincoumans:
That's great, thanks for the contribution(s). I'll have a look at it as soon as possible.

BluePrintRandom:
Very intresting!

Have you included an  'ear'/ horizon sensor in your bot? for 'feeling' it's
orientation?

Could an ai choose an ik animation to try to match with physics,  and thenb
choose how to deviate that ik pattern best for the environment?

erwincoumans:
Can you set the gContactProcessedCallback back to 0, in exitPhysics?
Otherwise examples afterwards will crash (run BasicExample after NN3DWalkers to reproduce issue).

void NN3DWalkers::initPhysics()
{
	gContactProcessedCallback = legContactProcessedCallback;

benelot:
I cleared the gContactProcessedCallback on exiting physics. I tried something today, but failed a bit to detect the following: How to detect perpetually interpenetrating btRigidBodies? Meaning two rigidbodies that are constrained to be unable to ever resolve the interpenetration. Can I get some information from the contact points? Or would you propose another strategy?

erwincoumans:
The gui sliders make it too easy to create spiders with overlapping geometry. Usually you disable collisions between bodies connected by constraints. You cannot create arbitrary large collision geometries that overlap and expect things to work. Yes, you can query contact points etc to detect overlap. If you want to evolve the geometry without overlap this would require more effort indeed.

benelot:
Do you have a suggestion on how to detect geometry without overlap easily? Can I somehow detect if an overlap is resolvable or if it is resolved after some time? Or would you just check in the beginning if there are spider body intercollisions and just drop those? I need to improve this for my own simulator as well, that is why I want invest more time into this.

erwincoumans:
I would check in the beginning if the spider body parts are overlapping, and remove/shrink the collision shapes until there is no overlap. It would be best to disable collisions between parts that are directly connected by a joint.

Some form of evolution would make it more interesting indeed. 
I also plan to create examples for various learning strategies (including Reinforcement Learning) using TensorFlow. Such examples can co-exist.

benelot:
Ok, I will implement that strategy and see how this  goes:) For the evolution, it would be helpful to be able to remove btRigidbodies again from the simulation (evaluating spiders one by one needs removal of them after evaluation) and I know from a previous discussion with you that this is not possible in the example browser. Can this be integrated easily or is that very hard? I could try to do this if you point me into the right way.

benelot:
@BluePrintRandom: What do you mean by an ik animation? It probably could, though. The question is how to give it the animation pattern, but as soon as you have a network that performs it, you could definitely make it adapt to the environment. You could probably train a network to do so, but that it way beyond this example. Stay tuned, there are more such examples to come.

benelot:
I implemented the collision avoidance strategy by dropping bodies that cause overlap. It is no longer possible to create broken walkers. Also it runs now a basic evolutionary algorithm training the walkers for distance within a certain evaluation time. Since it is not possible to drop graphical elements, I just kept them in the scene showing where the walkers ended up. Pretty cool for comparison but gets a bit crowded with more than 10 walkers. Is there an option to delete all current graphical objects?

I plan on creating a TimeWarpBase, which makes it easy to speed up the example (based on the TimeWarp example I wrote, see other pull request) and furthermore it could be a good idea to make is possible to print out the current population to store it and continue in some way. I will figure out something. Can I somehow read from console or similar to input a string that represents a population? 

Last but not least, I want to create a mode that just lets all walkers walk at the same time with collision turned off between them. And maybe a graph showing how they improve over generations.

benelot:
Graph, parallel evaluation and population print out are ready.  Next up, TimeWarpBase. Furthermore I will see if I can quickly build some more body structures (snakes, dogs, centipedes). Maybe I could also copy the example and create one wih softbodies and basic flying, swimming creatures. I am still searching for a way to input a data string to the example.

benelot:
TimeWarpBase is finished. I can speed up any of the example browser examples with minimal changes (a bit more than changing the inheritance). Just testing it with my Walker example, it runs amazingly quick now.

benelot:
# Summary

Implemented...

- [X] 
3D Walkers using a simple neural network to connect ground touch sensors with hinge joint motors. The Neural Network weights for successful walking are found using an Evolutionary Algorithm, training the Walkers for distance.
- [X]
Dimensions of the walker can be changed by using sliders. The walker's dimension will be changed when reloading the example. If the dimensions do not allow for a properly working walker (inner collisions), the colliding body parts added last are dropped. Do not be suprised if your walker lacks some parts of the legs.
- [X]
Parallel evaluation of the walkers
- [X]
Speedup of the evaluation and headless evaluation (minimal graphics, the one I can not turn off)

Up for discussion...

- [ ] Other morphologies like Snakes, Quadrupeds, Centipedes.
- [ ] Other environments using Softbody physics to approximate swimming or flying creatures (possibly in another example)
- [ ] Store evolution progress in file, export via configuration string (how should I import again?)

erwincoumans:
Aside from const reference in b3ReferenceFrameHelper.hpp, the demo crashes on Windows in various places: First in void NN3DWalkersExample::drawMarkings(), char performance[10]; -> sprintf out of bounds? Then in \src\BulletDynamics\ConstraintSolver\btSequentialImpulseConstraintSolver.cpp around line 1466, btAssert(fsum > SIMD_EPSILON);. Usually this means something went wrong in setup of the constraints. I would highly recommend dropping this demo and start looking to use pybullet instead.

benelot:
Hi Erwin, 

Thank you for the review! 

1) On linux, everything works well, but there seem to be issues on Windows. I will setup Visual Studio on my Windows machine to fix the issues. It seems that my code is not completely portable.

2) How about I finish the demo as such, for developers not yet adapting to the pybullet and b3RobotSimAPI and then recreate it using pybullet to have a version ported to pybullet?

What should I handle/remove the std::map? Interested in a btHashMap or similar?

d3x0r:
Just food for thought.
http://www.biologic.com.au/bugbrain/

Old project that lets you wire a simple neural network.  My thought was to make a hybrid of these and classic wide neural nets so that it's half constructed and half evolutionary.
A slight modification to the 'analog' type neuron so the output is the difference above the threshold instead of full out, and it's easy to make tracking logic... had a project with a bug wandering around with 3 raycast radar and it was able to avoid walls based on the relative distances of the radar... parameters that could be evolved - the max distance of detection, the scanning rates (the center radar was semi fixed, and the outside had a small wandering feel... like a blind person with a cane) .... and the input from the rays was 0-1 for how close it was to the radar on the body

meshula:
Cool work! 

@d3x0r I wrote a paper about a simulator I wrote back in 1990, along similar lines - 

http://nickporcino.com/posts/hexapod90.html

erwincoumans:
Yes, please fix the Windows build and I'll accept the pull request. Moving forward, it would be great to move this towards using pybullet and/or newer API. This new API abstracts so that we can move simulation to run on GPU (using OpenCL Bullet 3.x) and using machine learning (Bullet 4.x).

Yes, please use src/LinearMath/btHashMap.h instead of std::map.

erwincoumans:
Cool @meshula . Our new development with pybullet (Python) and VR will make it easier and fun to play with creature development. There will be TensorFlow in the mix as well, for machine learning...

benelot:
Hi Erwin, I am just about to finish it so that it can be merged. I am going to see how I can get up to speed with pybullet. 

Oh, so you already have a btHashMap? I thought I would have to contribute that as well, but well it is already there.

benelot:
All done and fixed. An uninitialized transform messed it all up. Happy to say it works now on Windows and Linux here.

erwincoumans:
Great, thanks a lot for the work on this. I'll check it out and try to merge the pull request asap.

erwincoumans:
There are a few minor issues that I fix after merging the pull request.
Now it would be great to reproduce this in pybullet (Python) and TensorFlow ... 👍

erwincoumans:
I moved the CommonInterfaces/CommonTimeWarpBase into Evolution/3DWalkersTimeWarpBase, since I didn't intent to add it as a common interface. Once you use pybullet you don't need any of this stuff anyway.

Also, why are you reimplementing the subdivision of the timestep?
https://github.com/bulletphysics/bullet3/blob/master/examples/Evolution/NN3DWalkersTimeWarpBase.h#L778
Bullet already has implemented the "fix your timestep" stuff in the stepSimulation:
https://github.com/bulletphysics/bullet3/blob/master/src/BulletDynamics/Dynamics/btDiscreteDynamicsWorld.cpp#L401

benelot:
That is interesting, my timestep code looks basically just like yours :) The determinism page on the bullet physics wiki gave me the impression that I had to implement this myself, that is why I did it. That is also why I added a TimeWarpBase to give people a basis to understand how to speed up simulations deterministically. I will try to cancel the TimeWarpBase and have the same effect. Thanks for giving me a heads-up! 

Also thanks for the link to your friend's repository. That and other setups are basically what I wanted to create for the example browser. I am fairly new to the DDPG approach, but I want to read the paper on it and study the code. Is this about what you would like to get as examples? Next will definitely be a ported pybullet NN walkers example.

erwincoumans:
there are a bunch of nice ideas in the timewarp base class, especially selecting solvers and gui and other stuff. let me look at how to merge it in the current classes.

benelot:
Nice that you found some ideas you could integrate into your own classes. Tell me if I can help with the integration into your classes. It would probably go into the RigidBodyBase, right? Maybe I can do it all in while cancelling the TimeWarpBase.

benelot:
I found a minor bug that plots the fitnesses of the walkers wrongly and I found out that the runs seem not to be deterministic (a performance is not at all the same when run the several times). I will have to fix this.

PR #724, opened 2016-07-31
Breakout example

benelot:
# tl;dr Check summary comments below

I created a simple Breakout example for people to enjoy. I am not completely sure if the hit reflections are correct, but generally it looks okay (bad reflections might be more problematic for a pool simulation, which is an issue I am helping a user with on the forum). Any suggestions to the breakout game?


benelot:
Border.h:24:7: error: non-aggregate type 'vec3' cannot be initialized with an initializer list. Ok, I will fix this.

benelot:
Nice. It compiles.

BluePrintRandom:
in blender for reflections of a ray I use
Vector = own.worldOrientation.col[2]
heading = Vector.reflect(surface)

however this only works with 1d objects

I think you need to calculate the normal of the average of the surface
defined by the collision contacts and then factor that in*

(in reality a sphere may  bounce  different then a cube depending on the
contact surface*)

benelot:
Thanks for the insight, I could definitely fix it by doing a ray reflecting using the surface normal of the objects. But do you have any clue why it does not reflect correctly by default? I would like to get the correct surface reflections when having a rotating ball etc...

BluePrintRandom:
I am afraid I don't know enough - but something to think about,

I think each collision, you either have friction, or reflection or both?

what is your surface friction like on your ball?

erwincoumans:
Thanks for the contribution. It should be possible to get the right reflection in both breakout and pool games using Bullet. There are various subtle issues that can prevent it from working, I haven't looked at your demo, not at the forum discussion. It would be good to create some pool example as well, especially when using the VR / HTC Vive/Oculus Rift physics server!

benelot:
Thanks Erwin for the comment, I will see if I can find anything wrong in it. As soon as I get it right, I will add something to the discussion on the forum. I will also contribute a pool example at some point, but about the VR physics server I have no idea how to use this, so I would need some instructions on how to use it. Is there already a good example in the browser and can I test it without VR (say lookaround with the mouse and move the hands if available with something else?).

erwincoumans:
Thanks for the breakout game. There are many files added that are almost adding no functionality. 
1) Can you merge everything into a single cpp file (and the usual header) like other examples?
2) Moving the bat requires hiting the left/right keys many times, instead of maintaining the velocity until it hits the wall.
Aside from this, it would be a good example to merge.
Thanks!

benelot:
Hi Erwin,

I merged everything into one file and added some more game features to it (Bricks are put aside when hit and are put back when all of them are hit once.).

Moving the bat continuously can be achieved by pressing the n/m button for a longer time. That is not how I programmed it, but that is how I get the events from:

    keyboardCallback(int key, int state) 

You can verify this by enabling the b3Printf() within the keyboardCallback. Then you see the prints as the events come in. It must be somewhere in the Example browser where this delay is added, probably because the buttons are mainly used for toggling functions in other examples.

erwincoumans:
Thanks a lot, I'll check it out.

>>Moving the bat continuously can be achieved by pressing the n/m button for a longer time.

No, you should not rely on the keyboard repeat feature, even if some platforms support keyboard repeat, it is not supported/enabled on all platforms. One really easy solution is to cache the keyboard state yourself, and update it when the state _changes_. 

Is that OK?

benelot:
True. I forgot about the keyboard repeat feature of some OSes. I cache the keyboard state and update the paddle movement according to it.

benelot:
# Summary

Implemented...
- [X] A basic Breakout game, move the paddle with n and m, shoot the ball with space.
- [X] Basic progress representation reusing the hit bricks (for the lack of a remove graphical entity method). Resets bricks if all are hit.

Up for discussion...
- [ ] Deep Reinforcement Learning similar to https://www.cs.toronto.edu/~vmnih/docs/dqn.pdf
- [ ] You tell me!

examples/ExtendedTutorials/Breakout.h
```diff
@@ -0,0 +1,135 @@
...
+ * All game objects inherit from this class.
+ */
+class GameObject {
+public:
+	GameObject():m_position(0,0,0),m_orientation(0,0,0,1),m_body(0),m_shape(0),m_tag(PADDLE) {}
+	virtual ~GameObject() {}
+	void createShapeWithVertices(const Vertex* vertices, unsigned int vertexCount,bool convex);
+	void createBodyWithMass(float mass);
+
+	void setPosition(btVector3 position);
```
erwincoumans: We need to pass btVector3 and btQuaternion using const reference, otherwise SSE/SIMD compilers break the build.
Here is the error message: >..\..\examples\ExtendedTutorials\Breakout.cpp(366): error C2719: 'position': formal parameter with requested alignment of 16 won't be aligned

 Can you please change it into: void setPosition(const btVector3& position); and void setOrientation(const btQuaternion& orientation);

erwincoumans:
I finally got to try the breakout game, but the ball doesn't return (tested on Windows), see video:
https://www.youtube.com/watch?v=mKu00bpyWb8&feature=youtu.be
There are 2 other small issues that the CI didn't catch, see inline code comment.

benelot:
Hello Erwin,

Thanks for the review. In Linux, I do not get the SSE/SIMD error, how can I cause them to appear? It would be very helpful for me to detect such issues before you have to point them out to me. I am going to fix the other issue immediately.

erwincoumans:
Let's move on to more interesting stuff like the Gym/pybullet work for now.

PR #734, opened 2016-08-10
allow command line config of example browser initial camera distance, pitch & yaw

matpalm:
the first thing i do opening an example browser is zoom in a bit and move around so making these command line flags seem useful.


erwincoumans:
Each example initializes its preferred camera perspective, so it should not go through the command-line.
A recent change in separating the physics and rendering in separate threads introduced this issue.
Here is the fix: https://github.com/bulletphysics/bullet3/pull/735

matpalm:
Fair enough. I don't understand fully how #735 helps me though sorry; if I've just instantiated bullet with `b3CreateInProcessPhysicsServerAndConnect` can I immediately set a new distance, pitch, yaw? Not sure what API to call....

PR #739, opened 2016-08-15
Fixed constraint for btMultiBody and btRigidBody.

YunfeiBai:


benelot:
Hi!Sounds like you know what you are doing! Would you be interested to add limits to the multibody joints as well? Also a quaternion PID is needed for the spherical joint. Just in case you would be interested to contribute :)

erwincoumans:
Thanks for the contribution, I'll look at it asap.

@benelot The btMultiBody joints already have limits, see the btMultiBodyJointLimit abd examples loading URDF and SDF in example browser.

PID for spherical joint would be nice indeed, we could sork on that later.

erwincoumans:
btMultiBodyJointLimitConstraint is the name. It works for revolute and prismatic joints of btMultiBody.

spherical joints are not used in many of the robots, so we haven't prioritized them.

benelot:
@Erwin: I should have known that this might be one of your guys
contributing such a pull request :)

erwincoumans:
Thanks for the help Yunfei!

PR #742, opened 2016-08-17
Fixed btKinematicCharacterController runtime crashes

sinkingsugar:
The previous btKinematicCharacterController pull request added some pretty obvious runtime crashes.
This is a quick attempt to fix it.
Works for us (Xenko engine) but a better review would be nice.


erwincoumans:
It it works for you, it works for me, the btKinematicCharacterController isn't actively being developed, and relies on contributions. Thanks!

PR #750, opened 2016-08-19
Fixed btKinematicCharacterController from clipping through walls in corners.

TIHan:
Noticed that btKinematicCharacterController was changed, decided to give it a whirl and noticed it had a problem as it was going through walls in corners. This should fix it.


erwincoumans:
Thanks. As there is no active development form our side, your contribution is welcome.
BTW, it would be good to create some kinematic character controller example, in the Bullet ExampleBrowser, or port the old demo.

PR #751, opened 2016-08-20
Added .gitignore - osx for right now

TIHan:
I noticed bullet doesn't have a `.gitignore` file. I think it would be helpful to have one. I don't see any downsides with having one here.

I did look to see if anyone else made a PR to add `.gitignore`, and it was rejected as it looks like you are supposed to build out of source. If that's the way it is, that's fine, but if the documentation to build the source involves a bunch of files to appear as untracked, then we should add documentation on building out of source or even doing it automatically for you on build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,27 @@
+
+# Files
+
+*.cmake
+*.dylib
+*.a
+CMakeCache.txt
+CMakeFiles
+Makefile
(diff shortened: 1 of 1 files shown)


erwincoumans:
There is some .gitignore now, due to some other PR. Thanks for the contribution anyway!

PR #752, opened 2016-08-21
Add swig for Python

portalfire:
Added Swig interface for Python.
Added demo based on “examples/HelloWorld/HelloWorld.cpp”
Added build flag “—enable_swig”
Added BT_ALLOW_SSE4 guards where they were missing.

### DIFF
--- a/build3/premake4.lua
+++ b/build3/premake4.lua
@@ -89,12 +89,18 @@
                 description = "Enable high-level Python scripting of Bullet with URDF/SDF import and synthetic camera."
         }
 
-if os.is("Linux") then
+	newoption
+  {
+    trigger = "enable_swig",
+    description = "Enable swigging of Bullet for use in Python."
+  }
+
+if os.is("Linux") or os.is("MacOSX") then
  		default_python_include_dir = "/usr/include/python2.7"
  		default_python_lib_dir = "/usr/local/lib/"
 end
 
-		
+
 if os.is("Windows") then
  		default_python_include_dir = "C:\Python-3.5.2/include"
  		default_python_lib_dir = "C:/Python-3.5.2/libs"
@@ -106,7 +112,7 @@ end
 			value       = default_python_include_dir,
 			description = "Python (2.x or 3.x) include directory"
     }
-    
+
     newoption
     {
 			trigger     = "python_lib_dir",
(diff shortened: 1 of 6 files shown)


erwincoumans:
That is an interesting contribution, I didn't expect it :) I manually created pybullet as a higher level physics-engine agnostic Python API mainly for robotics (loading URDF/SDF, forward/inverse dynamics, render image etc). This Bullet 2.x specific swig wrapper may also be useful to some. I'll try it out soon and give some feedback. Thanks!

portalfire:
This is my first pull request; I'm not sure what the correct etiquette is (should I discuss changes on the forum first?). My goal with this change is to make it easier to create a new physics based openAI gym environment. Right now I don't have a license to mujoco, and would like to use Bullet instead. I figured if bullet was exposed at a low level, it would be easy for one to create environments and 'robots' completely in python. Next up would be wrapping the URDF loading and creating the openAI gym part.

erwincoumans:
As you did is perfectly OK, submitting a pull request and discussing it here. And yes, it would be great to have an OpenAI gym connection. Have you considered trying out pybullet? 
My preference would be to extend pybullet with whatever is needed in a slightly more abstract way, so we can easily move to a new Bullet API without changing the Python API. But if you think this is not what you want/need, we can discuss it more here.

portalfire:
pybullet looks fantastic! I also agree that decoupling the API has advantages.

The functionality I thought was missing was:
* Programmatically creating the environment (think maze or non-level ground for walker). I might be able to build a sdf on the fly and then use the loadSDF.
* Ability to drive motors and get feed back. (I think was looking at version of pyBullet from 3 months ago, it has it now).
* Getting the pos/rot of all of the links in a multibody.  I was hoping to keep the physics simulation pure so it would be easy to distribute on a cluster for tensorflow training. I was going to have a "peek" rpc to send this information to a remote renderer (webgl).

I think for now I can close this pull request and move forward with pyBullet.

erwincoumans:
That's great. Please give me feedback on anything that is missing from pybullet. Ideally you generate urdf or sdf files. Position/orientation for links is available in pybullet. Documentation is sorely missing but the source is pretty small, see pybullet.c. Actually, the rendering runs on CPU using tinyrenderer, so it runs on clusters with tensorflow etc.

matpalm:
luckily exposing as a gym environment is trivial [here's an example in python of a gym env using rendering to feed raw pixels to a convnet](https://github.com/matpalm/cartpoleplusplus/blob/master/bullet_cartpole.py)

+1 for programmatic creation of URDFs!! i've been writing them out to tmp files on the fly ;)

PR #754, opened 2016-08-24
Set physics clock subdivider and fixed constraint in RobotSimAPI.

YunfeiBai:


erwincoumans:
Great, looks good to me. Thanks for the contributions. I'll have a look at exposing this to pybullet.

PR #756, opened 2016-08-26
Warnings and bugs

MichaelMcDonnell:
Hi Erwin,

I was fixing some warnings and discovered two bugs. The last two patches in this pull request fix what turned out to be bugs. Please let me know if you what something changed in any of these patches.

Thank you,
Michael Mc Donnell


MichaelMcDonnell:
I just realized that the two first patches have some portability issues. I will open a new pull request with just the last two fixes.

PR #758, opened 2016-08-26
Expose slider constraint in RobotSimAPI.

YunfeiBai:


erwincoumans:
LGTM. Thanks for the contribution. It would be also nice to expose it to pybullet, I'll look into it if I find time for it.

PR #759, opened 2016-08-27
Two bugs

MichaelMcDonnell:
Hi Erwin,

I found two bugs while trying to fix some warnings. Please let me know if you what something changed in any of these patches (I sent you an earlier version with four patches, but dropped two of them because of portability issues).

Thank you,
Michael Mc Donnell


erwincoumans:
looks good to me, thanks!

MichaelMcDonnell:
Sounds good! Thanks!

PR #763, opened 2016-08-29
Add flags to load inertia values from URDF.

ludi1001:


erwincoumans:

PR #765, opened 2016-08-31
Torsional and rolling friction for btMultiBody

YunfeiBai:


erwincoumans:
looks good to me, thanks for the work and pull request!

PR #767, opened 2016-09-02
Fix joint orientations when loading URDF files.

ludi1001:


erwincoumans:
Yes, this is broken in the multibody version and not in the rigid body version. Thanks a lot for tracking this down and fixing it!

PR #769, opened 2016-09-02
Add Phong reflection to TinyRenderer

iwilkes1:
Updated the shader to use specular reflection.  Also updated the coefficients used to calculate intensity at each pixel.

Before:
beore

After:
after


erwincoumans:
thanks for the contribution!

PR #774, opened 2016-09-09
Modify pybullet renderImage to return numpy array

moof2k:
Improves testrender.py performance on large images an order of magnitude

@erwincoumans

### DIFF
--- a/examples/pybullet/pybullet.c
+++ b/examples/pybullet/pybullet.c
@@ -8,6 +8,8 @@
 #include <Python.h>
 #endif
 
+#include <numpy/arrayobject.h>
+
 #if PY_MAJOR_VERSION >= 3
 #define PyInt_FromLong PyLong_FromLong
 #define PyString_FromString PyBytes_FromString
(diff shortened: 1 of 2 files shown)


erwincoumans:
Thanks for the contribution! This adds a dependency to numpy, which some
people like, others don't (on windows, many don't have numpy installed)


Could you put it behind an #ifdef, disabled by default? Then we can add the
option in the various build systems(I can take care of cmake, premake and
google3 BUILD ;)

Thanks!
Erwin

BluePrintRandom:
How does one distribute a application, and have it automatically install
the dependencies?

Can pip do that?

(like if you turn on his option and numpy is not installed, open a prompt
that asks the user if they want python to retrieve it?)

erwincoumans:
I re-created this pull request with your changes, with optional numpy support, works through cmake if your environment is setup ok: https://github.com/bulletphysics/bullet3/pull/780
Thanks!

erwincoumans:
This was merged instead: https://github.com/bulletphysics/bullet3/pull/780
We could also move the contact point data (and upcoming raycast data) in numpy arrays.
Thanks for the contribution!

PR #776, opened 2016-09-11
[URDF] Edit loader to handle full inertia tensor.

bingjeff:
Previous URDF loader did not handle off-diagonal elements
in the URDF. This commit adds functionality to allow
the loading of an inertia matrix that has not already
been reduced to its principal components. It diagonalizes
the inertia matrix, tests that the values are real and
updates the inertial frame.


erwincoumans:
That's great, thanks a lot Jeff.

By the way, as a future todo/reminder: it would be good to create a small test/comparison test/compare the btMatrix3x3::diagonalize function with some Eigen/other implementation.

bingjeff:
This pull request is superseded by #783 which does not have white space edits.

PR #783, opened 2016-09-13
[URDF] Edit loader to handle full inertia tensor.

bingjeff:
This is a corrected patch that doesn't have additional whitespace corrections.

Previous URDF loader did not handle off-diagonal elements
in the URDF. This commit adds functionality to allow
the loading of an inertia matrix that has not already
been reduced to its principal components. It diagonalizes
the inertia matrix, tests that the values are real and
updates the inertial frame.

### DIFF
--- a/examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
+++ b/examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
@@ -295,26 +295,66 @@ void  BulletURDFImporter::getMassAndInertia(int linkIndex, btScalar& mass,btVect
 	if (linkPtr)
 	{
 		UrdfLink* link = *linkPtr;
+		btMatrix3x3 linkInertiaBasis;
+		btScalar linkMass, principalInertiaX, principalInertiaY, principalInertiaZ;
 		if (link->m_parentJoint==0 && m_data->m_urdfParser.getModel().m_overrideFixedBase)
 		{
-			mass = 0.f;
-			localInertiaDiagonal.setValue(0,0,0);
+			linkMass = 0.f;
+			principalInertiaX = 0.f;
+			principalInertiaY = 0.f;
+			principalInertiaZ = 0.f;
+			linkInertiaBasis.setIdentity();
 		}
 		else
 		{
-			mass = link->m_inertia.m_mass;
-			localInertiaDiagonal.setValue(link->m_inertia.m_ixx,link->m_inertia.m_iyy,
-										  link->m_inertia.m_izz);
+			linkMass = link->m_inertia.m_mass;
+			if (link->m_inertia.m_ixy == 0.0 &&
+			    link->m_inertia.m_ixz == 0.0 &&
+			    link->m_inertia.m_iyz == 0.0)
+			{
+				principalInertiaX = link->m_inertia.m_ixx;
+				principalInertiaY = link->m_inertia.m_iyy;
+				principalInertiaZ = link->m_inertia.m_izz;
+				linkInertiaBasis.setIdentity();
+			}
+			else
+			{
+				principalInertiaX = link->m_inertia.m_ixx;
+				btMatrix3x3 inertiaTensor(link->m_inertia.m_ixx, link->m_inertia.m_ixy, link->m_inertia.m_ixz,
+							  link->m_inertia.m_ixy, link->m_inertia.m_iyy, link->m_inertia.m_iyz,
+							  link->m_inertia.m_ixz, link->m_inertia.m_iyz, link->m_inertia.m_izz);
+				btScalar threshold = 1.0e-6;
+				int numIterations = 30;
+				inertiaTensor.diagonalize(linkInertiaBasis, threshold, numIterations);
+				principalInertiaX = inertiaTensor[0][0];
+				principalInertiaY = inertiaTensor[1][1];
+				principalInertiaZ = inertiaTensor[2][2];
+			}
 		}
-		inertialFrame = link->m_inertia.m_linkLocalFrame;
-		
+		mass = linkMass;
+		if (principalInertiaX < 0 ||
+		    principalInertiaX > (principalInertiaY + principalInertiaZ) ||
+		    principalInertiaY < 0 ||
+		    principalInertiaY > (principalInertiaX + principalInertiaZ) ||
+		    principalInertiaZ < 0 ||
+		    principalInertiaZ > (principalInertiaX + principalInertiaY))
+		{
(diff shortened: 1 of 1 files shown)


erwincoumans:
thanks a lot! I'm on travel but will look/merge it asap.

scpeters:
Cool, I hadn't noticed that `btMatrix3x3` already had a diagonalize method. Nice work.

erwincoumans:
I'd like to do a few small tests to see if the btMatrix3x3::diagonalize does what it is supposed to do. Comparing with Eigen for example. Once that is ok, I'll merge, early next week.

scpeters:
I have a few test cases of inertia matrices with analytical values of the diagonal and non-diagonal forms in an ignition-math test if you want to take a look:

* starting at [this line in MassMatrix3_TEST.cc](https://bitbucket.org/ignitionrobotics/ign-math/src/93cce26475e69d/src/MassMatrix3_TEST.cc#MassMatrix3_TEST.cc-238)

examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
```diff
@@ -295,26 +295,66 @@ void  BulletURDFImporter::getMassAndInertia(int linkIndex, btScalar& mass,btVect
...
+		{
+			b3Warning("Bad inertia tensor properties, setting inertia to zero for link: %s\n", link->m_name.c_str());
+			principalInertiaX = 0.f;
+			principalInertiaY = 0.f;
+			principalInertiaZ = 0.f;
+			linkInertiaBasis.setIdentity();
+		}
+		localInertiaDiagonal.setValue(principalInertiaX, principalInertiaY, principalInertiaZ);
+		inertialFrame.setOrigin(link->m_inertia.m_linkLocalFrame.getOrigin());
+		inertialFrame.setBasis(linkInertiaBasis * link->m_inertia.m_linkLocalFrame.getBasis());
```
ludi1001: I think the multiplication is backwards here. m_linkLocalFrame gives the transform from URDF inertial coordinates to world coordinates. linkInertiaBasis gives the transform from diagonalized inertia coordinates to URDF inertial coordinates. So inertialFrame, which should give transform from diagonalized inertia coordinates to world coordinates, should be m_linkLocalFrame.getBasis() * linkInertiaBasis?

examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
```diff
@@ -295,26 +295,66 @@ void  BulletURDFImporter::getMassAndInertia(int linkIndex, btScalar& mass,btVect
...
+			    link->m_inertia.m_iyz == 0.0)
+			{
+				principalInertiaX = link->m_inertia.m_ixx;
+				principalInertiaY = link->m_inertia.m_iyy;
+				principalInertiaZ = link->m_inertia.m_izz;
+				linkInertiaBasis.setIdentity();
+			}
+			else
+			{
+				principalInertiaX = link->m_inertia.m_ixx;
```
ludi1001: This line looks unnecessary.

bingjeff:
I actually have a unit test written already FWIW, since I didn't discover
the 3x3 diagonalization until after I had ported over some code to do SVD.
If you would like any of that code I can put it in.

examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
erwincoumans: Indeed ludu1001, the order should be m_linkLocalFrame.getBasis() * linkInertiaBasis. I'll fix that in a separate commit after the pull request is merged.

examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
bingjeff: Definitely a typo

examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
bingjeff: If you don't mind triple checking this. I presume that the inertia will be transformed as part of a similarity transform? I have checked this with a simple example and it appeared to be correct the way I have written, but I could definitely be wrong. I had assumed that the m_linkLocalFrame (short-hand R) was applied to the inertia matrix (shorthand M) as: R^T * M * R. Then with diagonalization (M=U^T * D * U) this would result in R^T * U^T * D * U * R. Finally this would lead to the new m_linkLocalFrame being R' = U * R.

erwincoumans: In our conventions it would be  R * M * R^T leading to R' = R * U.
See for example:
https://github.com/bulletphysics/bullet3/blob/master/src/BulletDynamics/Dynamics/btRigidBody.cpp#L256

bingjeff: Thanks Erwin! My test case appears to return the same result either way...so bad testing on my part.

scpeters: I made an SDF world file for gazebo that has several boxes with identical inertia in the world frame, but with different local frame inertia matrices. Not sure if that would be useful for your testing or not.

* [inertia_rotations.world](https://bitbucket.org/osrf/gazebo/src/38cd2490494bd0092ee7f5df522ba6b6f4ecea5f/test/worlds/inertia_rotations.world?at=default&fileviewer=file-view-default)
* it's an expansion from the following template using embedded ruby: [inertia_rotations.world.erb](https://bitbucket.org/osrf/gazebo/src/38cd2490494bd0092ee7f5df522ba6b6f4ecea5f/test/worlds/inertia_rotations.world.erb?at=default&fileviewer=file-view-default)

erwincoumans: Thanks Steve, will check it out (we have basic SDF loading)

erwincoumans: Just for the record, the swap of inertia/link local is in this merged pull request: https://github.com/bulletphysics/bullet3/pull/796/commits/0a628f06cc51a9856717e8e8a963fe98df5928f8#diff-3c1884a4d04d70625160133d7528653fL350

scpeters: That world file loads several boxes of identical inertia on a plane. I typically change the gravity vector to [2 0 -9.8] to tip them over and watch that the fall at the same rate.

PR #795, opened 2016-09-20
Inverse kinematics with orientation constraint.

YunfeiBai:


erwincoumans:
Brilliant (as they say in the UK), thanks for the work and pull request!

PR #798, opened 2016-09-21
Create a demo for one motor gripper grasp.

YunfeiBai:


erwincoumans:
great, thanks for the new gripper demo contribution!

PR #802, opened 2016-09-23
Do not add limits for continuous joints.

ludi1001:
It's more readable to add the check directly in joint loading rather than intentionally setting jointUpperLimit < jointLowerLimit in the parser in order to avoid having continuous joints be affected.


erwincoumans:
In the Bullet UrdfParser, the limits are default set to lower=0, upper=-1, and continuous joints are not overriding this (see https://github.com/bulletphysics/bullet3/blob/master/examples/Importers/ImportURDFDemo/UrdfParser.cpp#L1060). So this case/pull request is not needed. Perhaps a custom UrdfParser may not include this logic?

ludi1001:
It's not necessary, but I think it's more readable that way. I'm coming from strictly code readability standpoint where it's not very clear from the code that to ensure that continuous joints behave as continuous joints that the upper limit should be intentionally set to be below the lower limit.

erwincoumans:
agreed, so the pull request is merged. Thanks!

PR #803, opened 2016-09-23
expose b3GetLinkState in pybullet

matpalm:
also fix error msg & some more arg checks to getJointState


erwincoumans:
thanks for the contribution, looks good to me!

PR #808, opened 2016-09-27
KUKA and gripper control in VR.

YunfeiBai:


erwincoumans:
Thanks for the contributions!

PR #812, opened 2016-09-29
Add a new one motor gripper.

YunfeiBai:


erwincoumans:
thanks again for the contribution!

PR #816, opened 2016-09-30
Add null space task as an option for IK in VR. Fix angular velocity target in IK to prevent flip motion.

YunfeiBai:


erwincoumans:
LGTM, thanks!

PR #820, opened 2016-10-07
Fixed linker error in Visual Studio 2015

adam4813:
When using set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON) to generate DLLs 2 linker errors crop up from linking to LinearMath.lib. These errors are from btPolarDecomposition::DEFAULT_TOLERANCE and btPolarDecomposition::DEFAULT_MAX_ITERATIONS being declared in the header, but only defined in the btPolarDecomposition.cpp file. Therefore the definitions aren't found in btSoftBody.lib. The simple solution is to just include btPolarDecomposition.cpp into the btSoftBody compilation.

### DIFF
--- a/src/BulletSoftBody/CMakeLists.txt
+++ b/src/BulletSoftBody/CMakeLists.txt
@@ -15,6 +15,7 @@ SET(BulletSoftBody_SRCS
 	btSoftRigidDynamicsWorld.cpp
 	btSoftSoftCollisionAlgorithm.cpp
 	btDefaultSoftBodySolver.cpp
+	btPolarDecomposition.cpp
 
 )
 
--- a/src/BulletSoftBody/CMakeLists.txt
+++ b/src/BulletSoftBody/CMakeLists.txt
@@ -15,7 +15,7 @@ SET(BulletSoftBody_SRCS
 	btSoftRigidDynamicsWorld.cpp
 	btSoftSoftCollisionAlgorithm.cpp
 	btDefaultSoftBodySolver.cpp
-	btPolarDecomposition.cpp
+	../LinearMath/btPolarDecomposition.cpp
 
 )
 


erwincoumans:
That doesn't seem right. Can you modify the Bullet/src/BulletSoftBody/CMakeLists.txt line 44 into
TARGET_LINK_LIBRARIES(BulletSoftBody BulletDynamics BulletCollision LinearMath)
and see if that fixes things?

adam4813:
I just tried you suggesting, but nothing same linker errors. I did double check before I submitted this to make sure it was linking LinearMath.lib in the Visual Studio project properties and it was there.

erwincoumans:
It should work, please delete all build files, re-create cmake projects and build again, using that new line. I'll try it out as well in the meanwhile.

erwincoumans:
Also, please tell exactly what cmake options you use, and how you invoke cmake. Do you use cmake from command-line or cmake-gui?

erwincoumans:
If I use this line in command-line:
cmake .. -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON -G "Visual Studio 14 2015"
cmake reports an error:
  Manually-specified variables were not used by the project:

    CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS

adam4813:
I am trying to include it is vcpkg using this file
https://github.com/adam4813/vcpkg/blob/12958410a09b0b45afab2c9a3a3a116fa08ecbb1/ports/bullet3/portfile.cmake .

adam4813:
-DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON
    -DUSE_MSVC_RUNTIME_LIBRARY_DLL=ON
    -DBUILD_DEMOS=OFF
    -DBUILD_CPU_DEMOS=OFF
    -DBUILD_BULLET2_DEMOS=OFF
    -DBUILD_BULLET3=OFF
    -DBUILD_EXTRAS=OFF
    -DBUILD_UNIT_TESTS=OFF
    -DBUILD_SHARED_LIBS=ON
    -DINSTALL_LIBS=ON

erwincoumans:
I can't support custom build rules etc. Please try building Bullet (and optionally its demos) using its original unmodified cmake or premake build files.

adam4813:
Will do when I arrive back home later.

erwincoumans:
Thanks for the report, I could reproduce it. Let's remove those useless global constants, I created this pull request instead:
https://github.com/bulletphysics/bullet3/pull/821

adam4813:
I will check that pr to see if it solves my issue as well (can't imagine it won't).

adam4813:
Fixed by #821

PR #821, opened 2016-10-07
remove some useless global constants, move them right into the constructor

erwincoumans:
fixes https://github.com/bulletphysics/bullet3/pull/820


adam4813:
Confirmed fixes #820

PR #837, opened 2016-10-17
Add btSoftMultiBodyDynamicsWorld and set up the soft body grasp demo.

YunfeiBai:


erwincoumans:
Thanks for the pull request. It would be nice to add the option to disable the feature while it is under construction. For example, add a #ifdef USE_SOFT_BODY_MULTI_BODY_DYNAMICS_WORLD around the new blocks inside PhysicsServerCommandProcessor.cpp, and define this in the project build system, so we have an option to not link/use btSoftBody*.

YunfeiBai:
Thank you Erwin! I have added an option for using btSoftMultiBodyDynamicsWorld.

erwincoumans:
Great, thanks for the changes, LGTM.

PR #841, opened 2016-10-21
Programmatic render API for changing texture in TinyRenderer.

YunfeiBai:


erwincoumans:
looks good to me, thanks for the contribution!

PR #850, opened 2016-11-02
Contact handling between btMultiBody and btSoftBody.

YunfeiBai:
Add contact handling between btMultiBody and btSoftBody. Add demo for two way coupling between btMultiBody and btSoftBody. Add demo for grasping bunny with multibody gripper.


erwincoumans:
Thanks for the contribution! The 'kuka_iiwa/model_free_base.urdf seems missing, could you add it please?

YunfeiBai:
Thanks Erwin! I have added the 'kuka_iiwa/model_free_base.urdf'.

To load the bunny and use the softbody dynamics in shared memory API, the preprocessor USE_SOFT_BODY_MULTI_BODY_DYNAMICS_WORLD
needs to be defined. Cause the softbody contact is still work in progress.

erwincoumans:
Thanks! When I enabled #define USE_SOFT_BODY_MULTI_BODY_DYNAMICS_WORLD on latest trunk, the bunny 'explodes' when touching ground plane, presumably for static objects there should only be one-way response :)

YunfeiBai:
Hmm.. I don't see the explosion with the merged latest trunk and my master branch on my side... Is there a way that I can have the exact same code base that you are testing with.

PR #860, opened 2016-11-12
collisionObject: renamed uniqueId to worldArrayIndex

lunkhound:
This fixes a bug in the multithreaded dynamics world when objects are removed.
The uniqueId field in collision object has been replaced by a more aptly named worldArrayIndex.
The worldArrayIndex is -1 if the object is not in the world array.

I took the opportunity to use the worldArrayIndex to avoid a linear search through the collisionObjects array in CollisionWorld::removeCollisionObject().

### DIFF
--- a/src/BulletCollision/CollisionDispatch/btCollisionObject.cpp
+++ b/src/BulletCollision/CollisionDispatch/btCollisionObject.cpp
@@ -28,6 +28,7 @@ btCollisionObject::btCollisionObject()
 		m_collisionFlags(btCollisionObject::CF_STATIC_OBJECT),
 		m_islandTag1(-1),
 		m_companionId(-1),
+        m_worldArrayIndex(-1),
 		m_activationState1(1),
 		m_deactivationTime(btScalar(0.)),
 		m_friction(btScalar(0.5)),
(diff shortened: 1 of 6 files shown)


erwincoumans:
I'm not fond of this uniqueId/worldArrayIndex. Can you explain why it is needed? I'm reading this comment:

  // Kinematic bodies can be in multiple islands at once, so it is a
 // race condition to write to them, so we use an alternate method

https://github.com/lunkhound/bullet3/blob/2f4ec4f8c9faaf5ca0a8c4ae113d414f15cc334c/src/BulletDynamics/ConstraintSolver/btSequentialImpulseConstraintSolver.cpp#L751

Kinematic objects should have mass zero, so the constraint solver should never have to 'write' to them, right?

lunkhound:
It is the companionId of the collision body that is written to. If a kinematic body is in say 2 different simulation islands, the 2 solvers running concurrently will both write different values to the companionId (index of the solver body in the solver body-array).

lunkhound:
So while the worldArrayIndex may not be the most elegant approach, I would argue that the companionId is worse. The companionId is solver-specific data being stored in the global collision objects and being mutated by solvers. At least worldArrayIndex is not being mutated.

Another solution that would not require use of the worldArrayIndex would be for each solver to keep a hashMap for mapping from body-pointers to solver body index.

One benefit of having worldArrayIndex is that it speeds up removing objects from the world from being O(N) operation to O(1). So it has utility beyond just fixing this one issue with concurrent solvers sharing a kinematic body.

Let me know how you'd like this solved and I'll be glad to implement it.

erwincoumans:
ok, I agree. Let's make this change. Thanks for the contribution!

PR #861, opened 2016-11-14
Add shared memory API for setting light direction.

YunfeiBai:


erwincoumans:
Thanks for the contribution! 
I replaced renderImage by getCameraImage using keyword args (needs some testing!) 
this way, we can add more optional arguments in the future.
See this one instead: https://github.com/bulletphysics/bullet3/pull/864

PR #865, opened 2016-11-20
Set light color.

YunfeiBai:


erwincoumans:
Thanks for the contribution!
Do you have any example in the Example Browser (either 'Rendering' section or 'Experiments') or pybullet file to see the lightcolor in action?

PR #867, opened 2016-11-21
example browser: slider widget improvements

lunkhound:
- Added a void* parameter to the Slider's callback which is used to pass the user pointer
- Added a 'clamp-to-integers' option to the slider (and cleaned up various hacky workarounds)


lunkhound:
By the way, the clamp to integer sliders have a much improved UI feel versus what they were doing before. They felt kind of "laggy" before. But this fixes it.

erwincoumans:
looks good to me (LGTM), thanks for the contribution!

PR #870, opened 2016-11-22
Added Bazel build file.

Lunarsong:


erwincoumans:
Thanks for the contribution of a Bazel build file. Aside form getting the right projects and files in there, it would be also good to add some support for Travis continuous integration to build/test the Bazel build. I did a quick search, and it seems some projects already do it:
https://github.com/search?q=.travis.yml+bazel&type=Code&utf8=%E2%9C%93

In a nutshell, src/Bullet3Dynamics and src/Bullet3Collision is mainly for GPGPU Physics using OpenCL and not in production yet. So we are still using Bullet 2.x ,which is src/LinearMath, src/BulletCollision, src/BulletDynamics, src/BulletSoftBody. The ExampleBrowser takes a bit more work to build. You can see the non-public blaze BUILD file for details.

erwincoumans:
Realistically I don't have much time to maintain a public BUILD file: we already have CMakeLists.txt, premake and setup.py (Python setuptools). Let's close and abandon it for now, unless there is really more interest in this and someone who maintains it.

PR #884, opened 2016-12-07
Expose ambient, diffuse, and specular coefficients of the TinyRenderer light.

YunfeiBai:


erwincoumans:
Thanks for the contribution/pull request, I'll process it later today.

PR #885, opened 2016-12-10
Fix a bug when rendering texture.

YunfeiBai:


erwincoumans:
thanks for fixing it!

PR #894, opened 2016-12-20
FIX: kinematic objects treated as static when BT_THREADSAFE is defined

lunkhound:
The first commit changes the MultiThreadedDemo example to work as a test case to replicate the bug. In the example, use the "Ground horiz amp" slider to add some horizontal shaking to the ground. The stacks of boxes should react when the ground moves and fall down pretty quickly. Because of the bug however, the stacks of boxes don't see any motion from the ground. Vertical shaking still affects the stacks because of penetration.

The second commit fixes the bug. Now the boxes will tumble when the ground moves horizontally.

Tested on Windows 10, Visual Studio 2013, win32, built with Cmake with BULLET2_USE_THREAD_LOCKS enabled, as well as BULLET2_MULTITHREADED_*_DEMO enabled.

### DIFF
--- a/examples/MultiThreadedDemo/CommonRigidBodyMTBase.h
+++ b/examples/MultiThreadedDemo/CommonRigidBodyMTBase.h
@@ -406,6 +406,17 @@ struct CommonRigidBodyMTBase : public CommonExampleInterface
 		return body;
 	}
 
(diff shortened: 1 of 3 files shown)


erwincoumans:
thanks for your fast help!

PR #895, opened 2016-12-20
example browser: restore on-screen text

lunkhound:
The examples in the example browser which are based on CommonRigidBodyMTBase are supposed to be displaying on-screen text to display various information such as profiling stats (but only when the "display profile timings" button is on). Currently this is only drawn when in 'wireframe' display mode which is odd.

This PR fixes that so the text appears consistently.


erwincoumans:
thanks for the contribution, and happy 2017!

PR #905, opened 2017-01-03
example browser: add controls for changing the solver type and solver flags

lunkhound:
For examples based on CommonRigidBodyMt, add GUI buttons for switching between various solver types. Also add buttons for toggling various solver mode flags. Solver flags can be changed in realtime but changing the solver type requires a restart of the example.
Also updated GUI toggle buttons so that they can be started in a specific state (instead of always off).
And the toggle button callback now passes the current button state (instead of always true).
The GUI button changes were needed so that you can see the state of the solver flags by looking at the buttons.

### DIFF
--- a/examples/CommonInterfaces/CommonParameterInterface.h
+++ b/examples/CommonInterfaces/CommonParameterInterface.h
@@ -43,6 +43,7 @@ struct ButtonParams
 	int m_buttonId;
 	void* m_userPointer;
     bool m_isTrigger;
+    bool m_initialState;
 
 	ButtonParamChangedCallback m_callback;
 	ButtonParams(const char* name, int buttonId, bool isTrigger)
@@ -50,6 +51,7 @@ struct ButtonParams
 		m_buttonId(buttonId),
 		m_userPointer(0),
     m_isTrigger(isTrigger),
+    m_initialState(false),
 	m_callback(0)
 	{
 	}
--- a/examples/ExampleBrowser/GwenGUISupport/GwenParameterInterface.cpp
+++ b/examples/ExampleBrowser/GwenGUISupport/GwenParameterInterface.cpp
@@ -3,12 +3,14 @@
 
 struct MyButtonEventHandler : public Gwen::Event::Handler
 {
+    Gwen::Controls::Button* m_buttonControl;
 	ButtonParamChangedCallback m_callback;
 	void* m_userPointer;
 	int m_buttonId;
 
-	MyButtonEventHandler(ButtonParamChangedCallback callback, int buttonId, void* userPointer)
-		:m_callback(callback),
+	MyButtonEventHandler(Gwen::Controls::Button* buttonControl, ButtonParamChangedCallback callback, int buttonId, void* userPointer)
+		:m_buttonControl(buttonControl),
+        m_callback(callback),
 		m_userPointer(userPointer),
 		m_buttonId(buttonId)
 	{
@@ -18,7 +20,12 @@ struct MyButtonEventHandler : public Gwen::Event::Handler
 	{
 		if (m_callback)
 		{
-			(*m_callback)(m_buttonId, true, m_userPointer);
+            bool buttonState = true;
+            if (m_buttonControl->IsToggle())
+            {
+                buttonState = m_buttonControl->GetToggleState();
+            }
+            ( *m_callback )( m_buttonId, buttonState, m_userPointer );
 		}
 	}
 };
@@ -140,10 +147,11 @@ void GwenParameterInterface::registerButtonParameter(ButtonParams& params)
 {
 	
 	Gwen::Controls::Button* button = new Gwen::Controls::Button(m_gwenInternalData->m_demoPage->GetPage());
-	MyButtonEventHandler* handler = new MyButtonEventHandler(params.m_callback,params.m_buttonId,params.m_userPointer);
+	MyButtonEventHandler* handler = new MyButtonEventHandler(button, params.m_callback,params.m_buttonId,params.m_userPointer);
 	button->SetText(params.m_name);
 	button->onPress.Add( handler, &MyButtonEventHandler::onButtonPress );
     button->SetIsToggle(params.m_isTrigger);
+    button->SetToggleState(params.m_initialState);
     
 	m_paramInternalData->m_buttons.push_back(button);
 	m_paramInternalData->m_buttonEventHandlers.push_back(handler);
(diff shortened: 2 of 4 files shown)


erwincoumans:
We need to add multiple build systems in the continuous integration, not just cmake:
I'm getting this error, when I apply this patch locally (build using

cd build3
./premake4_linux64 gmake
cd gmake
make

../../examples/MultiThreadedDemo/CommonRigidBodyMTBase.cpp: In member function ‘virtual void CommonRigidBodyMTBase::createEmptyDynamicsWorld()’:
../../examples/MultiThreadedDemo/CommonRigidBodyMTBase.cpp:751:42: error: ‘BT_MAX_THREAD_COUNT’ was not declared in this scope
             btConstraintSolver* solvers[ BT_MAX_THREAD_COUNT ];
                                          ^
../../examples/MultiThreadedDemo/CommonRigidBodyMTBase.cpp:755:17: error: ‘solvers’ was not declared in this scope
                 solvers[ i ] = createSolverByType( m_solverType );
                 ^
../../examples/MultiThreadedDemo/CommonRigidBodyMTBase.cpp:757:52: error: ‘solvers’ was not declared in this scope
             m_solver = new MyConstraintSolverPool( solvers, maxThreadCount );
                                                    ^

lunkhound:
I'll take a look in a couple of days when I have some time.

erwincoumans:
No rush, I've been running the Bullet project for over 13 years now, and hope on another decenium at least :)

lunkhound:
It seems that that error you got was not specific to the premake build -- same error happens with the Cmake build as well. At the time of the PR, there was no compile error and so continuous integration checks passed. Sometime later, a commit was made (Jan 8) that broke the PR but the continuous integration failed to re-check. I had assumed that the CI would catch that sort of thing...
Anyway it should be fixed now.

erwincoumans:
Thanks!

>> I had assumed that the CI would catch that sort of thing...

Indeed, it would be nice to have some 'retrigger' button for CI or have it done automatically.

erwincoumans:
By the way, I haven't tried any of the multithreaded work, since it doesn't compile out-of-the-box.
Do you have a few lines to explain how to install the dependencies (if any, for example TBB+version, OpenMP etc) for Windows, Linux and Mac OSX, and how to enable it?

lunkhound:
I can't say much about Mac OSX since I don't have any OSX machines handy. Also I don't recall anyone trying to use it with OSX from the MT thread on the forums.

For OpenMP there are no external dependencies, so enabling it is just turning on a flag in Cmake. OpenMP works with Windows/Visual Studio and also Linux/GCC. When I tested it with Linux/Clang there was a problem with a missing "omp.h" system header, so it doesn't work with Clang at the moment.

Microsoft's Parallel Patterns Library (PPL) is also supported and has no dependencies, but it's Visual Studio only.

For Intel's TBB you'll need to install it separately, and then point Cmake to the include and lib directories.
Download a precompiled distribution of TBB from [https://www.threadingbuildingblocks.org/download](https://www.threadingbuildingblocks.org/download) (choose your OS), or you can grab it from [https://github.com/wjakob/tbb](https://github.com/wjakob/tbb) and build it from source. The latest version of TBB should work (just tested it on Windows), I've also tested it in the past with 4.3 and 4.4 versions with no issues. The version on Github is unofficial but adds a Cmake build system which could be nice for those who wish to build from source.

To configure Bullet for the multithreaded demo, use the Cmake build system to enable the flags

- BULLET2_USE_THREAD_LOCKS
- BUILD_BULLET2_DEMOS

and at least one of the following flags:

- BULLET2_MULTITHREADED_OPEN_MP_DEMO (Windows/Visual Studio and linux/GCC; not Clang)
- BULLET2_MULTITHREADED_PPL_DEMO (Windows/Visual Studio 2010 or later only)
- BULLET2_MULTITHREADED_TBB_DEMO (requires external dependency -- see below)

Once TBB is installed somewhere, go back to the Bullet Cmake configuration and set paths for:

- BULLET2_TBB_INCLUDE_DIR (point this to the top "include" directory in the TBB install)
- BULLET2_TBB_LIB_DIR (point this to the directory with the correct "tbb.lib" file for your OS and compiler)

erwincoumans:
Great, thanks lurkhound! Did you know that you can press 'p' in the latest Bullet ExampleBrowser to start recording performance traces, and releasing 'p' to write those multithreaded performance tracing files to disk? Those can be imported directly into Chrome using about://tracing and load the .json file. Here are some screenshots, made by pressing (and releasing) 'p' during the raycast benchmark, serial and parallel:

sequential_raycast
parallel_raycast

lunkhound:
I did not know about that--that looks super-useful!

Incidentally, I've been working on a graph-coloring parallel solver. It runs but I'm not getting much of a speedup. This profiling tool looks like just what I need to figure out my problem.

Very nice!

erwincoumans:
That sounds good. Note that there are two graph coloring parallel solvers in Bullet, one in the OpenCL GPU pipeline. This OpenCL parallel solver is most fancy: it also has the parallel version of the graph coloring (batch creation) in two stages, all on GPU. You can see some description in Bullet/docs/GPU_rigidbody_using_OpenCL.pdf
Code of parallel solve iterations in OpenCL is here:
https://github.com/bulletphysics/bullet3/blob/master/src/Bullet3OpenCL/RigidBody/kernels/solveContact.cl#L399
Parallel batch creation (graph coloring) is here:
https://github.com/bulletphysics/bullet3/blob/master/src/Bullet3OpenCL/RigidBody/kernels/batchingKernelsNew.cl#L144
It still runs on my NVIDIA 1080GTX GPU, when running the ExampleBrowser using the --enable_experimental_opencl (except the mouse picking crashes). I still intent to re-enable this OpenCL pipeline behind the pybullet, Bullet C-API, which is our main focus nowadays.

The old btParallelConstraintSolver.cpp in BulletMultiThreaded was here:
https://github.com/erwincoumans/bullet2stripped/blob/6b3828d74b06b55b58224cdee401ce9de8f131ef/src/BulletMultiThreaded/btParallelConstraintSolver.cpp

It was converted from the original old PhysicsEffects parallel solver (with graph coloring/batching) was here: https://github.com/erwincoumans/PhysicsEffects/blob/master/src/low_level/solver/pfx_constraint_solver_parallel.cpp

lunkhound:
By the way, thanks for all that helpful info!

After looking through much of this, one thing I can't figure out is:
For the parallel batch creation in OpenCL, is it deterministic?
I'm having a hard time seeing how it could be.

Issue #909, opened 2017-01-05
Accessing the links of a URDF body


benelot:
I have an URDF file like here:

<?xml version="1.0"?>
	<robot name="cartpolev0">
	  <link name="cart">
	    <contact>
	      <lateral_friction value="1.0"/>
	      <rolling_friction value="0.0"/>
	      <inertia_scaling value="3.0"/>
	      <contact_cfm value="0.0"/>
	      <contact_erp value="1.0"/>
	    </contact>
	    <inertial>
	      <origin rpy="0 0 0" xyz="0 0 0"/>
	      <mass value="1"/>
	      <inertia ixx="1" ixy="0" ixz="0" iyy="1" iyz="0" izz="1"/>
	    </inertial>
	    <visual>
	      <origin rpy="0 0 0" xyz="0.0 0.0 0.0"/>
	      <geometry>
	        <box size="0.2 0.2 0.05"/>
	      </geometry>
	      <material name="blockmat">
	        <color rgba="0.9 0.2 0.1 1"/>
	      </material>
	    </visual>
	    <collision>
	      <origin rpy="0 0 0" xyz="0.0 0.0 0.0"/>
	      <geometry>
	        <box size="0.2 0.2 0.05"/>
	      </geometry>
	    </collision>
	  </link>

	  <joint name="cart_pole_joint" type="continuous">
		<axis xyz="0 0 1"/>
		<parent link="cart"/>
		<child link="pole"/>
		<origin rpy="1.57 0 0" xyz="0 0 0"/>
		<limit effort="100" velocity="100"/>
		<joint_properties damping="0.0" friction="0.0"/>
	  </joint>

	  <link name="pole">
	    <contact>
	      <lateral_friction value="1.0"/>
	      <rolling_friction value="0.0"/>
	      <inertia_scaling value="3.0"/>
	      <contact_cfm value="0.0"/>
	      <contact_erp value="1.0"/>
	    </contact>
	    <inertial>
	      <origin rpy="0 0 0" xyz="0 0 0"/>
	      <mass value="5"/>
	      <inertia ixx="1" ixy="0" ixz="0" iyy="1" iyz="0" izz="1"/>
	    </inertial>
	    <visual>
	      <origin rpy="0 0 0" xyz="0.0 0.25 0.0"/>
	      <geometry>
	        <box size="0.02 0.5 0.02"/>
	      </geometry>
	      <material name="blockmat">
	        <color rgba="0.2 0.7 0.1 1"/>
[...]

PR #910, opened 2017-01-05
[WIP] Basic DQN agent training on CartPolev0 env with pybullet

benelot:
Here is an example using a hand-implemented OpenAI environment. It is still work in progress since it does not use the URDF with the joint in it (I will fix it with the insights from #909).

I have some args to set to train it with and without gui. How should we make it runnable out of the box? Different scripts to train and one to load a trained agent to show that it works? 

I will additionally make an example loading the environment using the xml file. How do you like the organization of my files? Do you have something in mind on how to organize the python files in the pybullet folder in case they get more and more?

P.S. I will add the cartpole++ example of matpalm as a reference later, since I learned a lot from it to build this. 

Have a good day, it is time to sleep here!

### DIFF
--- a/data/cartpole-env/cart.urdf
+++ b/data/cartpole-env/cart.urdf
@@ -1,0 +1,32 @@
+<?xml version="1.0"?>
+<robot name="cart">
+  <link name="baseLink">
+    <contact>
+      <lateral_friction value="1.0"/>
+      <rolling_friction value="0.0"/>
+      <inertia_scaling value="3.0"/>
+      <contact_cfm value="0.0"/>
+      <contact_erp value="1.0"/>
+    </contact>
+    <inertial>
+      <origin rpy="0 0 0" xyz="0 0 0"/>
+      <mass value="1"/>
+      <inertia ixx="1" ixy="0" ixz="0" iyy="1" iyz="0" izz="1"/>
+    </inertial>
+    <visual>
+      <origin rpy="0 0 0" xyz="0.0 0.0 0.0"/>
+      <geometry>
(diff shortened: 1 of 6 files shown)


erwincoumans:
>>How should we make it runnable out of the box? Different scripts to train and one to load a trained agent to show that it works?

How does the original OpenAI GYM cartpole example with MuJoCO deal with this? Ideally we follow that approach as close as possible.

benelot:
I did some research on different examples and found out that there is no special structure at the OpenAI gym. They all include a description on how to reproduce the results. Even the example by John Schulman includes a very detailed description followed by scripts with tons of flags.

https://gym.openai.com/evaluations/eval_C7PJ590SxakxJnHcgGKIQ

So for now I will make a train script and a load weights and run script. The training takes pretty long, so most people should go for the load script to see the proof of concept. 

The basic principle for the gym is that they provide you with an environment which is implemented with their interface, which is what I did.

Then people provide an rl algorithm that optimizes its policy while interacting with the environment. But there is no particular structure on how to do this, because people have to read to run it in general.

erwincoumans:
There will likely be a bit more experimenting to see how to integrate pybullet with OpenAI Gym, and it seems better to create some separate repos for this. Could you create a separate repo similar to Matt Kelceys cartpoleplusplus first (instead of submitting it to Bullet main repo)?
Thanks!

erwincoumans:
(then after a while, we can decide how to merge this with Bullet)

benelot:
Ok, I can do this. I will let you know when I am done via mail.

benelot:
We can close this as this is now all in bullet-gym.

PR #921, opened 2017-01-19
pybullet: Error message wrong when linkIndex < 0 in getLinkState

benelot:
If the linkIndex is set to < 0, it says invalid jointIndex when it is actually about linkIndex.

This commit fixes it.


erwincoumans:
sure. note that jointIndex == linkIndex in Bullet/pybullet.

benelot:
True. I just found it strange that I get an error on incorrect jointIndex when I am querying the linkstate via the linkIndex.

PR #942, opened 2017-02-02
Fixed uninitialized default argument

willcassella:
The default value for the argument to btKinematicCharacterController::jump leaves it uninitialized and causes strange behavior.


erwincoumans:

PR #943, opened 2017-02-03
Rod holding demo (stability issues)

olegklimov:
My robot don't want to hold anything in its arm in a stable manner. I made am isolated demo of what's going on. 

No parameter tuning helps. In other words, instability is very stable.

It is not specific to multibody solver. btDantzigSolver behaves almost the same.

Nothing numerically extreme, all parameters of objects in question are in 1..10 range. Tested with double precision. Single precision behaves the same.

### DIFF
--- a/examples/ExampleBrowser/CMakeLists.txt
+++ b/examples/ExampleBrowser/CMakeLists.txt
@@ -336,6 +336,8 @@ SET(BulletExampleBrowser_SRCS
  ../MultiBody/MultiBodyConstraintFeedback.cpp
   ../MultiBody/MultiDofDemo.cpp
   ../MultiBody/MultiDofDemo.h
+  ../MultiBody/GripStability.cpp
+  ../MultiBody/GripStability.h
   ../RigidBody/RigidBodySoftContact.cpp
   ../Constraints/TestHingeTorque.cpp
   ../Constraints/TestHingeTorque.h
--- a/examples/ExampleBrowser/ExampleEntries.cpp
+++ b/examples/ExampleBrowser/ExampleEntries.cpp
@@ -28,6 +28,7 @@
 #include "../MultiBody/MultiBodySoftContact.h"
 #include "../MultiBody/MultiBodyConstraintFeedback.h"
 #include "../MultiBody/MultiDofDemo.h"
+#include "../MultiBody/GripStability.h"
 #include "../MultiBody/InvertedPendulumPDControl.h"
 #include "../RigidBody/RigidBodySoftContact.h"
 #include "../VoronoiFracture/VoronoiFractureDemo.h"
@@ -126,6 +127,7 @@ static ExampleEntry gDefaultExamples[]=
 
 	ExampleEntry(0,"MultiBody"),
 	ExampleEntry(1,"MultiDofCreateFunc","Create a basic btMultiBody with 3-DOF spherical joints (mobilizers). The demo uses a fixed base or a floating base at restart.", MultiDofCreateFunc),
+	ExampleEntry(1,"GripStability", "Hold a rod in the fingers.", GripStability),
 	ExampleEntry(1,"TestJointTorque","Apply a torque to a btMultiBody with 1-DOF joints (mobilizers). This setup is similar to API/TestHingeTorque.", TestJointTorqueCreateFunc),
 	ExampleEntry(1,"TestPendulum","Simulate a pendulum using btMultiBody with a constant joint torque applied. The same code is also used as a unit test comparing Bullet with the numerical solution of second-order non-linear differential equation stored in pendulum_gold.h", TestPendulumCreateFunc),
 
--- a/examples/ExampleBrowser/premake4.lua
+++ b/examples/ExampleBrowser/premake4.lua
@@ -120,6 +120,7 @@ project "App_BulletExampleBrowser"
 		"../Vehicles/*",
 		"../Raycast/*",
 		"../MultiBody/MultiDofDemo.cpp",
+		"../MultiBody/GripStability.cpp",
 		"../MultiBody/TestJointTorqueSetup.cpp",
 		"../MultiBody/Pendulum.cpp",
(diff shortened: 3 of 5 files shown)


olegklimov:
image

erwincoumans:
Hi Oleg,

Thanks for describing the issue, I'll check it out.

At a first glance, you don't seem to use the btMultiBody at the moment. It
is not enough to use the btMultiBodyDynamicsWorld, there is a totally
different way to create btMultiBody objects (unlike btRigidBody). See for
example under ExampleBrowser/Experimental/ GripperGrasp","Two Point Grasp"
and "One Motor Gripper Grasp"

I'll re-create your setup using URDF, loaded as a btMultiBody and start
from there. You can see some differences in ConvertURDF2BulletInternal in
Bullet/examples/Importers/ImportURDFDemo/URDF2Bullet.cpp

I suggest you give it a short break and give me some time to configure it.
Thanks!
Erwin

olegklimov:
You're right, I didn't try btMultiBody and probably I should. I was under the impression it makes sense to use it for bigger robots and stability within the robot itself, and here is very simple situation with three bodies.

So the key is to create this two fingers within one multibody? Or just create the same as multibody with numLinks==1 and it'll start working?

erwincoumans:
Grasping such round long thin object is hard problem in real-time robotics simulation for various reasons. You will need some spinning friction (there is very little the few contact points can do to prevent rotation), double precision, smallish time step, sufficient number of solver iterations (drift can be caused by the solver terminates early with large error), and keeping the gripped fingers parallel, making sure the closing forces of the gripper fingers are not in conflict with the contact forces and error reduction parameters of the contact (ERP) and many more. I'll look at it.

Again, in the meanwhile, I would suggest looking in

1) example under "ExampleBrowser/Experimental/GripperGrasp",
2) "ExampleBrowser/Experimental/Two Point Grasp"
3) "ExampleBrowser/Experimental/One Motor Gripper Grasp"
ConvertURDF2BulletInternal in Bullet/examples/Importers/ImportURDFDemo/URDF2Bullet.cpp

olegklimov:
I agree that it is a hard problem. I really appreciate that you look into it. To help, I have some more notes...

Probably, instability has nothing to do with long cylinder. Cylinder just makes it more profound. To illustrate, please remove cylinder from the code, set EFFORT=1, observe stable simulation. Now set EFFORT=10, now two fingers behave wobbly when they meet, moving not only along unconstrained direction, but also up and sideways, rotate. 

It hints for numerical problems somewhere, reproducible in simple setting, but I can only guess. It also doesn't go away with small step size or double precision. Only m_numIterations seem to help, instability becomes smaller in amplitude but still present.

Oleg.

erwincoumans:
In a nutshell, when using real-time robotics simulation you'll likely terminate the LCP constraint solver early with some residual error. This error will cause all kind of issues, and lots of effort goes into resolving this. This includes recovery from penetration, computing good reliably penetration depth/separating vectors. Recovering from penetration has its own issues (adding energy, which translates into jitter etc), so we need  another effort to cope with this. So we have 'split impulse', solving two separate LCPs (one on position level, other on velocity level) to not add too much kinetic energy due to penetration recovery. 

If you simulate joints as constraints, this LCP error translates into sloppy joints. This introduces even more issues, with feedback loops. Hence, for real-time Robotics simulation it is better to use reduced coordinate methods to simulate joints, and not using constraint to simulate the joints.  This means joints have no joint gap. This is what btMultiBody does, using Featherstone's Articulated Body Algorithm. 

I think we can get your use case working with existing Bullet btMultiBody and some careful tuning. 

Since this is my main effort (higher quality simulation for robotics and VR), there are a few near-term plans on my list that helps your challenging problem:

1) positional friction anchors. This will probably fix the issue entirely. In short, as a contact is created, it will add friction anchors, and make sure they are kept together using position correction. You can think of this as a error/drift correction, similar to penetration depth along the contact normal, but now applied laterally for friction (usually friction is only handled on the velocity level, hence the drift when your LCP solver has residual error).

2) Block LCP solver. This will solve small systems much better, without paying the cost of a global LCP solver (such as Dantzig or Lemke, which have their own numerical and other issues)

3) NNCG, this LCP solver is an improvement on the PGS (Projected Gauss Seidel aka btSequentialImpulseConstraintSolver ). NNCG is not yet coupled to the btMultiBody solution.

etc.

olegklimov:
That's a very clear explanation!

So in my own words, I've hit a simplest case when one needs reduced coordinate method already. The reason is, joints are simulated as constraints, on the same level as contacts. So to correct contact errors, joint constraint errors appear that also need correction. That leads to interplay and instability.

OK, that means I'm not posting another report of flyaway to infinity until I move the code to multibody :smile:

BluePrintRandom:
This was very intriguing to me,  I slowed my instability by sharing a % of
linv each frame to the root and back each frame for each joint of the sim,
 in blender game engine,  because we don't have a featherstone solver or
btmulibody exposed for some reason,

If a patch were pushed to upbge I am sure panzergame would accept it.
(the current place where real bge development is taking place)

On Feb 3, 2017 10:36 AM, "Oleg Klimov" <notifications@github.com> wrote:

That's a very clear explanation!

So in my own words, I've hit a simplest case when one needs reduced
coordinate method already. The reason is, joints are simulated as
constraints, on the same level as contacts. So to correct contact errors,
joint constraint errors appear that also need correction. That leads to
interplay and instability.

OK, that means I'm not posting another report of flyaway to infinity until
I move the code to multibody 😄

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/bulletphysics/bullet3/pull/943#issuecomment-277326008>,
or mute the thread

erwincoumans:
I added the tefal pan urdf and modified the 'ExampleBrowser/Experiments/OneMotorGripperGrasp' to grasp the pan. It is a very challenging problem and I had to increase the forces a lot, use sufficient spinning friction (in URDF), increase number of solver iterations to 150 etc.

PR #945, opened 2017-02-03
Set joint damping in IK.

YunfeiBai:


erwincoumans:
Thanks a lot for the contribution!  Since the Mac continuous integration takes too long, I expect it to be all fine, so merge it.

PR #949, opened 2017-02-07
EGL OpenGL window.

YunfeiBai:


examples/OpenGLWindow/EGLOpenGLWindow.cpp
```diff
@@ -30,178 +30,231 @@
 
 #ifdef BT_USE_EGL
 
-#include "EGLOpenGLWindow.h"
+#include <pthread.h>
```
erwincoumans: Are those headers really required (pthread.h)? Thanks!

erwincoumans:
Thanks for the contribution! I may cleanup some headers, if needed.

PR #956, opened 2017-02-10
Spin-mutex minor optimization

leonmaxx:
This is lightweight CPU instruction level sleep and event macros for spin-mutexes.
They does'nt affect performance but helps to reduce CPU stalls.

Sorry for my bad english, You can edit my comments.
BR, Leon.


lunkhound:
I tested this out on Windows 10/Win32/MSVC 2015. Looks good to me.

ahundt:
The ideas look good but these should be actual inlined function calls, not `#define`

leonmaxx:
I'll rewrite it to functions, should I make another pull request or append it here?

ahundt:
I'm sure appending is fine

erwincoumans:
Thanks for the contribution! I'll look into it soon and apply it.

dyanikoglu:
@erwincoumans This is a useful change for implementations highly utilizing threads. Any updates for the PR?

PR #965, opened 2017-02-18
Pull request

jietan:
refine the minitaur class to make it closer to the real robot.


data/plane.urdf
```diff
@@ -2,7 +2,7 @@
 <robot name="cube.urdf">
   <link name="planeLink">
   <contact>
-      <lateral_friction value="1.5"/>
+      <lateral_friction value="2"/>
```
erwincoumans: Thanks for the contribution! Since this plane.urdf is used by all demos, it is better to create a new plane2.urdf with higher friction. At some stage, we'll allow to programmatically change the friction etc.

PR #967, opened 2017-02-20
Generic robot state logging

YunfeiBai:


erwincoumans:
Thanks for the contribution! There are some redundant dependencies in the python scripts (I think the script doesn't use numpty, but it includes the module, most Windows python distributions don't have numpy) We can clear that up later.

PR #971, opened 2017-02-22
A bug fix for logging the orientation of the base of the object.

YunfeiBai:


erwincoumans:
getWorldToBaseRot is a trap indeed. It is better to use getWorldTransform() and get the rotation from there. I should have removed that API long ago.

PR #977, opened 2017-02-26
Update helloworld.tex with tip on proper library link order.

manmohanbishnoi:
Added a tip on how to properly link BulletPhysics libraries in order.

### DIFF
--- a/docs/latex/helloworld.tex
+++ b/docs/latex/helloworld.tex
@@ -6,6 +6,8 @@ It is a good idea to try to compile, link and run this HelloWorld.cpp program fi
 
 As you can see in \ref{helloworld_includefiles} you can include a convenience header file \path{btBulletDynamicsCommon.h}.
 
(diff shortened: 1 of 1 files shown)


erwincoumans:
Sure, thanks for the addition.
By the way, recently we started working on a more general robotics simulator, using [pybullet](https://docs.google.com/document/d/10sXEhzFRSnvFcl3XxNGhnD4N2SedqwdAvK3dsihxVUA/edit). There is also going to be a C++ version of this, which abstracts from Bullet 2.x, so we can easier move to Bullet 3.x and beyond.  See this header, in case it interests you:
https://github.com/bulletphysics/bullet3/blob/master/examples/RobotSimulator/b3RobotSimulatorClientAPI.h

manmohanbishnoi:
@erwincoumans first of all thanks for such a quick merge
I have a couple of updates for bullet documentation. Will make pull requests for them in a few days (gotta clean up those changes)
Also I work on OpenGL, OpenCL & Vulkan API and will surely join Bullet 3.x development

PR #987, opened 2017-03-04
remove `using namespace ConvexDecomposition` from library header file

pps83:
commit message I guess is self explanatory. There are a few very similar classes with identical names and if you include these files:
ConvexDecomposition/cd_hull.h
ConvexDecomposition/ConvexBuilder.h 
LinearMath/btConvexHull.h
you'll get compilation error.


erwincoumans:
Good point, thanks for the improvement!

PR #988, opened 2017-03-04
Log time stamp and set renderer in pybullet.

YunfeiBai:


erwincoumans:
Thanks for the contribution! 
I was also adding a small feature, allowing to set the maximum number of dofs to log (instead of hard-coded 12), since my minitaur has 16 dofs to log :-) I'll add this in the next commit.

PR #992, opened 2017-03-06
Fix warning when compiling with -Dundef on MinGW

taylor001:


erwincoumans:
thanks for the report/fix!

PR #996, opened 2017-03-07
Adjust thread-local local static variable syntax for MinGW/GCC.

iSLC:


erwincoumans:
thanks for the contribution, good to see some MinGW attention!

PR #1000, opened 2017-03-08
URDF loader improvement 1

olegklimov:
**What it does**

1. Tries to load resources from URDF file dir, then from parent dir, etc.

2. Removes "package://".

3. Keeps URDF source file reference in memory, some error messages now include source.urdf:line.

4. tiny_obj_loader.cpp LoadObj() material path can now be zero.

**Testing**

My URDFs work:

```
roboschool/models_robot/atlas_description/urdf/atlas_v4_with_multisense.urdf 
roboschool/models_robot/fetch_description/robots/fetch.urdf 
roboschool/models_robot/r2d2.urdf 
roboschool/models_kitchen/low60type2.urdf
```

Error message:

```
b3Printf: b3Warning[/home/kotik/bullet3/examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp,515]:

b3Printf: roboschool/models_robot/atlas_description/urdf/atlas_v4_with_multisense.urdf:799: cannot find 'atlas_description/meshes_unplugged/p__typo__elvis.dae' in any directory in urdf path
```

Search attempts:

```
tried 'roboschool/models_robot/atlas_description/urdf/atlas_description/meshes_unplugged/p__elvis.dae'
tried 'roboschool/models_robot/atlas_description/atlas_description/meshes_unplugged/p__elvis.dae'
tried 'roboschool/models_robot/atlas_description/meshes_unplugged/p__elvis.dae'
tried 'roboschool/atlas_description/meshes_unplugged/p__elvis.dae'
```

Automatic material path for .obj:

```
MTL PATH 'roboschool/models_kitchen' orig 'roboschool/models_kitchen/low60type0.obj'
```

@erwincoumans you have very different VR setting, some bug can show up!

### DIFF
--- a/examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
+++ b/examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
@@ -755,7 +755,8 @@ upAxisMat.setIdentity();
 	} // mesh case
 
         default:
-		b3Warning("Error: unknown visual geometry type\n");
+		b3Warning("Error: unknown collision geometry type %i\n", collision->m_geometry.m_type);
+		// for example, URDF_GEOM_PLANE
 	}
 	return shape;
 }
@@ -943,7 +944,7 @@ static void convertURDFToVisualShapeInternal(const UrdfVisual* visual, const cha
 		}
 
 		default:
-			b3Warning("Error: unknown visual geometry type\n");
+			b3Warning("Error: unknown visual geometry type %i\n", visual->m_geometry.m_type);
 	}
 
 	//if we have a convex, tesselate into localVertices/localIndices
--- a/examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
+++ b/examples/Importers/ImportURDFDemo/BulletUrdfImporter.cpp
@@ -457,9 +457,8 @@ static btCollisionShape* createConvexHullFromShapes(std::vector<tinyobj::shape_t
 	return compound;
 }
 
-static
 bool findExistingMeshFile(
-	const std::string& path_or_shorter, std::string fn,
+	const std::string& urdf_path, std::string fn,
 	const std::string& error_message_prefix,
 	std::string* out_found_filename, int* out_type)
 {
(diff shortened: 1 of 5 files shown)


olegklimov:
[TinyRendererVisualShapeConverter.cpp](https://github.com/bulletphysics/bullet3/blob/master/examples/SharedMemory/TinyRendererVisualShapeConverter.cpp) have the same file lookup code copy-pasted, want me to reuse `findExistingMeshFile()` there too?

erwincoumans:
Thanks a lot for the contribution. How does it remove the package://?
We do want to allow absolute paths ("file://c:/path/to/urdf.urdf"). Does it impact this?

olegklimov:
> How does it remove

It just removes characters in the beginning, if it's there.


> We do want to allow absolute paths ("file://c:/path/to/urdf.urdf").

It's not path to URDF, it's path to a resource referenced by URDF. My guess is we're unlikely to ever meet an absolute path inside URDF. Reason for this, you clone repositories in random directory on your computer, every absolute path will break.

erwincoumans:
Can you please explicitly remove just 'package://'?

I rather not remove more general characters from the resource paths, unless strictly necessary.

erwincoumans:
>> TinyRendererVisualShapeConverter.cpp have the same file lookup code 
>> copy-pasted, want me to reuse findExistingMeshFile() there too?

Probably a good idea, so if you have time, please do so. Thanks!

olegklimov:
> Can you please explicitly remove just 'package://'?

It is.


> TinyRendererVisualShapeConverter

Done.

Please test on VR, it is completely different setting from mine.

erwincoumans:
Thanks a lot for working on this. I'll test it in VR soon and merge it!

erwincoumans:
Loading the SDF file "two_cubes.sdf"  fails with this patch.
It first fails in 

if (!glmesh || !glmesh->m_vertices || glmesh->m_numvertices<=0)
			{
				b3Warning("%s: cannot extract anything useful from mesh '%s'\n", urdfPathPrefix, visual->m_geometry.m_meshFileName.c_str());
				break;
			}
Then it assets later

	// should never get here (findExistingMeshFile returns false if it doesn't recognize extension)
				btAssert(0);

It may be better if this patch was split into simpler to review patches, there is some refactoring going on.

olegklimov:
Oh, I finally found some .sdf-s in repository. Right, SDF branch wasn't working (untouched).

image

> It may be better if this patch was split into simpler to review patches

It is fairly simple... The reason diff in larger than it has to -- is the whitespace and indentation.

Like you suggested, I was going to fix error messages too, but decided to leave that for later.

erwincoumans:
Thanks a lot for the work, let's merge it!

erwincoumans:
>>It's not path to URDF, it's path to a resource referenced by URDF. 
>>My guess is we're unlikely to ever meet an absolute path inside URDF

Wrong guess. We have auto-generated urdf files that point to assets with absolute path. Let's not reduce previous functionality. Added check for original mesh filename first (without prefix etc):

https://github.com/bulletphysics/bullet3/pull/1017

PR #1003, opened 2017-03-09
pybullet VR controller controls the kuka arm

JulianYG:
I try to simulate the behavior from C++ files, but the IK here is very inaccurate and there's always a 0.03 offset between end effector and the VR controller.


erwincoumans:
Thanks for the pull request! We'll look into why the 0.03 offset.

JulianYG:
I got a few more things working today-- I think the shaking problem is caused by too much position gain. I'll update soon.

PR #1009, opened 2017-03-13
Fix some compiler warnings

cprogrammer1994:
- sinf to btSin
- acosf to btAcos
- float to btScalar
- 1e9 to ... (1e9 is treated as double)

### DIFF
--- a/src/BulletDynamics/Featherstone/btMultiBody.h
+++ b/src/BulletDynamics/Featherstone/btMultiBody.h
@@ -616,7 +616,7 @@ private:
     void operator=(const btMultiBody &);  // not implemented
 
 
-	void solveImatrix(const btVector3& rhs_top, const btVector3& rhs_bot, float result[6]) const;
+	void solveImatrix(const btVector3& rhs_top, const btVector3& rhs_bot, btScalar result[6]) const;
 	void solveImatrix(const btSpatialForceVector &rhs, btSpatialMotionVector &result) const;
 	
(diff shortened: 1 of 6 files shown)


src/LinearMath/btQuickprof.cpp
```diff
@@ -218,7 +218,7 @@ unsigned long long int btClock::getTimeNanoseconds()
 		QueryPerformanceCounter(&currentTime);
 		elapsedTime.QuadPart = currentTime.QuadPart - 
 			m_data->mStartTime.QuadPart;
-		elapsedTime.QuadPart *= 1e9;
```
cprogrammer1994: `1e9` is treated as `double`
`elapsedTime.QuadPart` is not a `double`

src/BulletDynamics/Featherstone/btMultiBody.h
```diff
@@ -616,7 +616,7 @@ void addJointTorque(int i, btScalar Q);
     void operator=(const btMultiBody &);  // not implemented
 
 
-	void solveImatrix(const btVector3& rhs_top, const btVector3& rhs_bot, float result[6]) const;
```
cprogrammer1994: this function is not used.

erwincoumans: it is a left-over from older code before we introduced spatial algebra. We can remove it.

erwincoumans:
thanks for the improvements!

PR #1010, opened 2017-03-13
Fixing a bug on Visualizer Camera and a few more

khansari:
- resetDebugVisualizerCamera now accepts negative values for pitch and
yaw angles

- Defining kitchen model to be concave, so as to be able to put floating
objects inside

- Fixed an indention error in testrender_np.py


erwincoumans:
Nice, LGTM, thanks for the contribution!

khansari:
Pleasure. Thanks for your help.

PR #1012, opened 2017-03-13
fix using SSE on non x86 targets

DaveeFTW:
Currently when building bullet for Windows on an ARM platform it will fail due to lack of SSE.


erwincoumans:
Thanks, looks good to me!

PR #1015, opened 2017-03-14
Fixed shaking issue, and hopefully the offset should be gone.

JulianYG:
I realized if I remove orientation of end effector in pybullet IK, it would follow the controller better. Also I reduced the position gain the make it smoother.


erwincoumans:
Thanks for the help. It would be nice not to depend on numpy, but just use python math (many people don't have numpy installed, especially on Windows). Could you fix that?
By the way, do you really need the arcin(sin)?

JulianYG:
Arcsin Sin is a trick to limit the rotation within joint range--it would also be easy to just add a joint range check on that, instead of using numpy. I'll fix that later.

JulianYG:
Fixed!

erwincoumans:
Yesterday I noticed the original C++ version of KUKA following-IK was much smoother than your python one, but didn't apply this pull request yet. Also, the gripper had some joints rotating and not smoothly closing the gripper. I haven't tried your new patch yet. Could you try to use an identical setup to the C++ version (in  PhysicsServerCommandProcessor::createDefaultRobotAssets() and https://github.com/bulletphysics/bullet3/blob/master/examples/SharedMemory/PhysicsServerCommandProcessor.cpp#L5503
We are busy with many other things, but will look into it if we don't get the Python version behave the same as the build-in C++ version. Thanks!

PR #1019, opened 2017-03-14
Kinematic controller initialization

erikogenvik:
m_gravity was used before it was initialized, which could lead to strange behaviour.
I added a test to flush it out.


erwincoumans:
Thanks for the contribution and test!

PR #1021, opened 2017-03-14
Heightfield demo

erikogenvik:
Added a demo of the height field shape in order to hunt down an issue with boxes slipping through. It's currently mainly of use for demonstrating the strange behavior.


erikogenvik:
Awesome with a coding style document!
I must admit that it's kinda hard to find. Would it perhaps be easier to find if it was placed alongside the code? Like "CODINGSTYLE.md" or similar?

erwincoumans:
Thanks for sharing. The issue you are pointing out is not surprising, but due to local penetration depth recovery: the height-field is a group of elements (extruded triangles, forming some kind of 3d prism). If a box smaller than the element width drop in-between multiple elements, those elements push the box sideways instead of upwards. 

We could use triangles instead of extruded triangles (prisms/boxes) elements to avoid this issue.

erwincoumans:
Closing some old issues/pull requests. There is a heightfield example now, see screenshot here:
Screenshot from 2021-03-04 13-41-45

PR #1024, opened 2017-03-16
expose video capture as logging command

erwincoumans:


benelot:
Wohoo! Just yesterday I thought how cool it would be if I had this
functionality in a programmatic fashion. Thanks Erwin, I am going to be
your tester for this feature.

erwincoumans:
You need to install ffmpeg. It works best on Linux and Mac OSX. 
On Windows, ffmpeg streaming rawvideo into a pipe works but has visual artifacts (wrong colors, tearing etc). In addition, it seems useful to have a more to disable real-time rendering updates, and explicitly ask to render frames in the debug visualizer.

benelot:
Ok, thanks for the input. I will check it out.

Issue #993, opened 2017-03-07
pybullet: getVisualShapeData() returns wrong link indexes


olegklimov:
Here is isolated example:

```python
import pybullet as p
p.connect(p.GUI)

huskypos = [0,0,0.1]
husky = p.loadURDF(sys.argv[1], huskypos[0],huskypos[1],huskypos[2])
cnt = p.getNumJoints(husky)

for joint in range(cnt) :
	print(p.getJointInfo(husky, joint))
	p.setJointMotorControl(husky, joint, p.VELOCITY_CONTROL, 0, 0)

viz = [""]*cnt
to_parse_array = p.getVisualShapeData(husky)
for uid, link_idx, geom, dims, filename, pos, orient in to_parse_array:
	if link_idx==-1: continue
	viz[link_idx] = str(filename)
	print("link %i have filename '%s'" % (link_idx, filename))

debug_text = []
for step in range (4000000):
	p.stepSimulation()

	for tid in debug_text:
		p.removeUserDebugItem(tid)
	debug_text = []
	for j in range(-1, cnt):
		if j==-1:
			p.getBasePositionAndOrientation(husky)
			continue
		pos, orient, inert_pos, inert_orient, frame_pos, frame_orient = p.getLinkState(husky, j)
		debug_text.append( p.addUserDebugText("O:" + viz[j][-12:], (pos[0]+1, pos[1], pos[2]) ) )
```

Output for this:

```
(0, b'back_bkz', 0, 7, 6, 1, 0.1, 0.0)
(1, b'back_bky', 0, 8, 7, 1, 0.1, 0.0)
(2, b'back_bkx', 0, 9, 8, 1, 0.1, 0.0)
(3, b'l_arm_shz', 0, 10, 9, 1, 0.1, 0.0)
(4, b'l_arm_shx', 0, 11, 10, 1, 0.1, 0.0)
...
link 23 have filename 'b'unplugged/pelvis.dae''
link 21 have filename 'b'unplugged/ltorso.dae''
link 22 have filename 'b'unplugged/mtorso.dae''
link 37 have filename 'b'unplugged/utorso.dae''
link 8 have filename 'b'unplugged/r_clav.dae''
link 16 have filename 'b'unplugged/r_scap.dae''
link 18 have filename 'b'unplugged/r_uarm.dae''
link 13 have filename 'b'meshes_v3/r_larm.dae''
```

[...]

PR #1025, opened 2017-03-16
URDF loader improvement 2

olegklimov:
Changes:

* MJCF now creates visual shapes, assigns body unique ids and link indexes to them.

* Capsule geometry accessible now through getVisualShapeData()

* Issue #993 (linkIndex and urdfIndex were mixed up)

* Fixed MJCF resources load, when run from directory outside Bullet.

* Killed `enum` I should have killed in previous PR.

* Renamed .STL to .stl for case-sensitive filesystems.

External rendering:

image

Demo repaired:

image


erwincoumans:
Wow, you already dug into the deep of the code, thanks for helping out!
Give me a few days to make sure it doesn't break some internal code here (some devs here have the habit of deriving classes of the Bullet examples, so that any change in API breaks...)
Minor note, I just renamed the https://github.com/bulletphysics/bullet3/blob/master/data/MPL/MPL.xml to use upper case .STL a few days ago, we can simply revert that :-)

erwincoumans:
Use B3_PI instead of M_PI (#include "Bullet3Common/b3Scalar.h" if needed)

olegklimov:
More colors and capsules work right.
image

erwincoumans:
Looks nice. How did you render the images?

olegklimov:
That's super-simple OpenGL code in roboschool. I plan to keep it, but also use proper external rendering engine (haven't chosen one yet) if it is installed.

Original colors for inverted pendulum in MuJoCo:

image

erwincoumans:
Nice. It would be nice to have some lightweight external renderer: Unreal 4 is a big beast, Unity 3D (and both proprietary, even though you can get the source code of UE4 under a non-FOSS license). I wonder if g4d is worth trying out, stripping what's not needed? What other options are you looking at?

olegklimov:
I asked on the internet, people suggest Ogre3d and bgfx... I will probably return to this soon. I want:

* Sky and day/night change,
* Indoor lighting, soft-shadows,
* Grass and vegetation in shaders,
* HDR, AA, postprocessing, god rays and things like this would be nice bonus.

I think it's not too much to ask, many game engines do this.

erwincoumans:
You may also check out g3d, http://g3d.cs.williams.edu/g3d/www/index.html. I'm also looking for better graphics features for the OpenGL Debug Visualizer (also used in VR), but prefer tiny bits of code rather than a huge framework like Ogre etc. I haven't tired g3d yet, it would be nice if you can take bits you need without using their framework. Another resource could be from NVIDIA, https://github.com/NVIDIAGameWorks/GraphicsSamples

benelot:
Hey Oleg, can you point me to where you found the urdf for the humanoid of
MuJoCo? I will need this to replicate the humanoid env.

erwincoumans:
@benelot  there is a humanoid.xml from MuJoCo in Bullet/data/mjcf

examples/SharedMemory/TinyRendererVisualShapeConverter.cpp
```diff
@@ -244,14 +292,6 @@ void convertURDFToVisualShape(const UrdfVisual* visual, const char* urdfPathPref
 			visualShapeOut.m_dimensions[1] = visual->m_geometry.m_meshScale[1];
 			visualShapeOut.m_dimensions[2] = visual->m_geometry.m_meshScale[2];
 
-			visualShapeOut.m_localVisualFrame[0] = visual->m_linkLocalFrame.getOrigin()[0];
```
erwincoumans: Why is the m_localVisualFrame initialization removed from URDF_GEOM_SPHERE?

olegklimov: You mean, from URDF_GEOM_MESH. Because it overrides defaults set by caller, defaults are the same.

(Of course please check again, I can very well make a mistake)

olegklimov:
About capsules in tiny render: they will appear with cones on both sides, if I'm understanding this right. Easy to improve (add a ring halfway).

erwincoumans:
ok, I'm going to merge it. If there is a regression somewhere, we'll fix that afterwards.
Thanks for the work/pull request!

PR #1036, opened 2017-03-23
Some improvements for better support swig generator.

nicolaichuk:
I update [bullet extension for libgdx](https://github.com/libgdx/libgdx/pull/4636), and need some change in bullet.

This change improvement support [swig generator](http://www.swig.org/) and fix some compile warnings.

@erwincoumans. please review and merge with upstream.


erwincoumans:
There is a bit too much change in this pull request, please reduce it a bit to strict necessary changes.
Why did refactor and split the header files and name them "Sructs"? Did you mean Structs?  Also, there are some new undocumented callbacks in the contact manifold etc.

nicolaichuk:
@erwincoumans, 

Yes, i mean Structs. I rename files.

I split header because in swig generator need parse files with class use in template parameter and only after that parse template that use this class.

for example 

```
class btContactArray:public btAlignedObjectArray<GIM_CONTACT>
```

used GIM_CONTACT, and both declaration in single file.

erwincoumans:
Why did you split the header files? Does SWIG not work with the original headers?

src/BulletCollision/Gimpact/btBoxCollision.h
```diff
@@ -234,7 +234,24 @@ ATTRIBUTE_ALIGNED16	(class) btAABB
 
 	btAABB()
 	{}
-
+
```
erwincoumans: Those operators don't make much sense, why do we need them and who is using them?

nicolaichuk:
>  Also, there are some new undocumented callbacks in the contact manifold etc.

this functionality used in [bullet libgdx extension](https://github.com/libgdx/libgdx/blob/master/extensions/gdx-bullet/jni/src/updating.txt). This callback for started and ended contact. Older callback also are undocumented. I do not know where to write the documentation for these methods. Please tell me where i can write information about this callback.

src/BulletCollision/Gimpact/btBoxCollision.h
nicolaichuk: i think that this operator don't have much sense, but 

GIM_BVH_DATA and GIM_BVH_TREE_NODE have btAABB. and they be used in 

btAlignedObjectArray<GIM_BVH_TREE_NODE> that have

```
		class less
		{
			public:

				bool operator() ( const T& a, const T& b ) const
				{
					return ( a < b );
				}
		};
```

So it code work i add operator compare in GIM_PAIR, GIM_BVH_DATA, GIM_BVH_TREE_NODE, and btAAB.

erwincoumans: What operator implementations was it using before for those? I'm not too familiar with the GIMPACT internals (never use it).

nicolaichuk: @erwincoumans:
I found how disable generate swig-wraper for this method. And revert compare operators in GIM_PAIR, GIM_BVH_DATA, GIM_BVH_TREE_NODE, and btAAB.

please review.

nicolaichuk:
@erwincoumans:
I found how disable generate swig-wraper for this method.

And revert compare operators in GIM_PAIR, GIM_BVH_DATA, GIM_BVH_TREE_NODE, and btAAB.

please review.

erwincoumans:
This is a very hard to review patch due to splitting of the files: I can't easily tell if you made any changes in the newly created files. Is there no other way to avoid splitting the files?

nicolaichuk:
@erwincoumans:

I do not made any changes in splitted files. 

if i will create separate PRs for each splitted files. Will it be easier for you to make a decision in this case?

nicolaichuk:
@erwincoumans:

You yourself write what it is desirable to place one class per one header file.

> 4. One class per header file, smaller structures can be combined in a single file.

https://docs.google.com/document/d/1u9vyzPtrVoVhYqQOGNWUgjRbfwfCdIts_NzmvgiJ144/edit

erwincoumans:
Thanks for your contribution, it looks good to me!

nicolaichuk:
@erwincoumans, thanks.

PR #1041, opened 2017-03-24
Random improvements (MJCF, error messages, gravity)

olegklimov:
I will be adding here any additional improvements, and I'll also always rebase on top. @erwincoumans please merge whenever you feel the right moment.

### DIFF
--- a/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
+++ b/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
@@ -3808,6 +3808,12 @@ bool PhysicsServerCommandProcessor::processCommand(const struct SharedMemoryComm
 									{
 										linkIndexB = mblB->m_link;
 										objectIndexB = mblB->m_multiBody->getUserIndex2();
+										if (
+											(clientCmd.m_updateFlags & CMD_REQUEST_CONTACT_POINT_HAS_LINK_INDEX_B_FILTER) &&
+											clientCmd.m_requestContactPointArguments.m_linkIndexBIndexFilter != linkIndexB)
+										{
+											continue;
+										}
 									}
 
 									int objectIndexA = -1;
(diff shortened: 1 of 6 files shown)


examples/SharedMemory/PhysicsServerCommandProcessor.cpp
```diff
@@ -868,7 +868,8 @@ struct PhysicsServerCommandProcessorInternalData
 
 
 	btScalar m_physicsDeltaTime;
-    btScalar m_numSimulationSubSteps;
+	btScalar m_numSimulationSubSteps;
+	btVector3 m_gravity;
```
erwincoumans: This is not desired: resetSimulation should reset the state as startup, undo all things the user did (loading URDF files, setting custom settings such as gravity). It should be trivial for a user to set the gravity after doing a reset.

erwincoumans:
The improvements are OK, except for remembering the gravity vector: resetSimulation should reset the known startup state. Can you please revert that change, and I'll merge the PR. Thanks!

olegklimov:
> remembering the gravity vector

Yeah that was silly of me. I'm seeing things that are not there.

Let me rebase this...

olegklimov:
image
image

Capsule/cylinder length is used in many places, can be a bug somewhere... Works for me, but I haven't check tiny renderer for example.

examples/SharedMemory/PhysicsServerCommandProcessor.cpp
```diff
@@ -3808,6 +3808,12 @@ bool PhysicsServerCommandProcessor::processCommand(const struct SharedMemoryComm
 									{
 										linkIndexB = mblB->m_link;
 										objectIndexB = mblB->m_multiBody->getUserIndex2();
+										if (
```
erwincoumans: Why is this filterneeded? Say the filter is for object 2 and object 3, then we should also allow object 3 and object 2 (the order doesn't matter). This check is already applied in the existing code.

olegklimov: link filter, not object filter

erwincoumans: Thanks, that's OK then. I'll review and test the other changes too.

erwincoumans:
LGTM, thanks for the contribution again!

PR #1054, opened 2017-04-03
pybullet setup.py / pip support, preliminary (only tested on Mac, will fix Linux/Windows soon)

erwincoumans:
See also https://pypi.python.org/pypi/pybullet


cprogrammer1994:
compiles fine on Windows 10

I get a lot of `'argument': truncation from 'double' to 'const btScalar'` but it is fine.

benelot:
These are going to disappear as people clean up the warnings, but it is not
an issue anyway as btScalar is either a double or a float depending on
compilation flags. So in the worst case, it is a double to float truncation.

erwincoumans:
Thanks for testing!
I'll enable BT_USE_DOUBLE_PRECISION in setup.py, also will remove warnings soon. Also, I'll enable TCP and UDP connection mode.
(all constants need to be cast to btScalar, so that both single float and double precision don't spawn warnings)

PR #1055, opened 2017-04-03
C API client installation (not a real PR)

olegklimov:
I don't want this to be merged as is, it is an experiment to estimate what it takes to create and install an actual library with Bullet C API.

Problems:

1. Gwen. Gwen is outdated, unergonomic widget library "without extravagant nonsense", bundled with Bullet. It has nothing to do with physics library. Code structure currently doesn't allow for it to be removed. I had to add installation procedure for Gwen.

2. Glew. Glew needed for Gwen. Modern operating systems have glew available, no reason to bundle it with Bullet. Bundled glew version is so outdated (or modified?), so it cannot be linked to system glew. For some reason, a number of symbols in the library don't link, even after I added bundled `glew.c` into library. But I was able to fix this by defining those in my program (that never calls any OpenGL in Bullet, I understand program can crash if it does), like this:

```C++
extern "C" {
void* __glewTexImage2D;
void* __glewPopMatrix;
...
```

(the trick will not work on windows.)

3. I had to add some spare parts from ExampleBrowser:

```
../../examples/ExampleBrowser/OpenGLExampleBrowser.cpp
../../examples/ExampleBrowser/OpenGLGuiHelper.cpp
../../examples/ExampleBrowser/GL_ShapeDrawer.cpp
../../examples/ExampleBrowser/CollisionShape2TriangleMesh.cpp
```

4. `BussIK`, `InverseDynamics `, `Serialize` are libraries that are not installed, only used in-place by examples. I've added some or all sources of this libraries into sources list.

The result is `libPhysicsClientC_API.so.2.87` that in turn depends on:

```
libBulletDynamics.so.2.87
libBulletCollision.so.2.87
libBulletInverseDynamics.so.2.87
libLinearMath.so.2.87
libBullet3Common.so.2.87
libgwen.so
```

It works, i.e. it is a library that implements C API.

This PR breaks pybullet.so (various OpenGL symbols missing).


erwincoumans:
Thanks for also looking into this. I would like to create a library like this, like Bullet/src/BulletRoboticsSimulator. I just created a first draft to install pybullet using pip. 
(see https://pypi.python.org/pypi/pybullet/0.1.3)
Once we have the BulletRoboticsSimulator library, it will be easier/faster to use pip install pybullet, linking against those libs instead of building everything in setup.py.

Note that This 'BulletRoboticsSimulator' goes beyond just a physics engine: it loads URDF, SDF, MJCF files, can render camera images, has an OpenGL debug visualizer and client-server connection. Even without GWEN, we use OpenGL for the debug visualization and optional rendering (aside from TinyRenderer). So all together, BulletRoboticsSimulator can replace Gazebo or MuJoCo. Now, some people may want to bring their own renderer. In addition, it could be useful to have some plugin registration mechanism to inject client-code such as IMU devices or control algorithms.

1) Gwen: In the future, we could replace Gwen with another simple widgets library such as IMGUI, for the debug visualizer. We cannot just get rid of it, since some GUI is important for the built-in debug visualization.

2) We need something like glew to load OpenGL. I had to modify it, so that we can dynamically load GL1.x functions (glew doesn't do it). In addition, we can dynamically load X11 and load GLX, instead of linking against it. I want to keep this feature of dynamic loading all. So if this modified glew becomes a bother, I rather rewrite from scratch some auto-generated OpenGL bindings, since system glew doesn't solve our requirements (dynamic load X11/GLX/GL1.x).

Thanks!
I'll follow up at some stage with thoughts/attempts towards Bullet/src/BulletRoboticsSimulator.

olegklimov:
Okay, as a random suggestion:
Why not integrate OpenGL rendering at pybullet.c level?

* It is possible, because external renderer is possible,
* Will force you to expose debug data through C API :smile:  especially collision shape,
* Unentangle import code from render,
* Provide natural line of separation between physics and use of physics (render not allowed to peek into any C++ structures, physics cannot use render),
* Make it possible to use C API as Bullet is traditionally used,
* Improve reload time for large models (only collision shapes analysed, render can keep old models),
* Provide ability to configure or not configure render at cmake, for more platforms,
* Render as plug-in becomes possible, because very few functions are needed for basic render to work.

The downside is no rendering on server side. But that is next to useless anyway (imagine using Bullet C API to explain server how to draw sky and where to place grass and fog and lights -- python interface is a natural place to do that, on client!).

> SHARED_MEMORY

Why should it make debug easier? It makes debug harder (two debuggers simultaneously?!), compile-run-edit cycle longer. Nothing compares to simply running a program under debugger, or does it?

olegklimov:
...and Erwin are you saying that Bullet will continue working if OpenGL is not present? (why else keep separation layer) How to test this? Will it help to improve reload speed (no need to viz shapes)?

benelot:
Before the bullet example browser and pybullet were built, Bullet had no
visualization at all. So I think that should be feasible again if required.

erwincoumans:
@olegklimov Yes, please check Bullet/test/SharedMemory/premake4.lua, the Test_PhysicsServerDirect has no OpenGL dependency, it includes a client and server.
See also attached picture. It looks like you are 'fighting' against the Bullet client-server infrastructure and don't use pybullet nor OpenGL debug visualization etc. If you support 'shared memory', and spawn a physics server separately, you can automatically use pybullet, in addition to your own graphics/python bindings.

physics client server components

olegklimov:
> Test_PhysicsServerDirect 

I just looked into Test_PhysicsServerDirect, it `links { "BulletInverseDynamicsUtils", "BulletInverseDynamics", "BulletFileLoader", "BulletWorldImporter", "Bullet3Common", "BulletDynamics",  "BulletCollision", "BussIK", "LinearMath" }` and a bunch of .cpp files... 

No Gwen? Seriously?!! Can't believe it! I'll be **so happy** if I can compile a library in the style of this PR. I'll try this ASAP.

erwincoumans:
Exactly, no GWEN and no OpenGL required when using DIRECT or Shared Memory. I'd like to merge the DIRECT option with 'shared memory', so that we always can have the option for pybullet scripting as well as VR,  and debug visualizer as a separate process with 1 line of code change: 'Shared memory' only adds a few files.

olegklimov:
Sorry for posting here, that was a terrible bug, visible only in release (not debug), ah tired as hell now :(

erwincoumans:
I already fixed those cylinder volume and capsule from/to issue a few hours ago here:
https://github.com/bulletphysics/bullet3/commit/e32debdca4a9bf6bf6a7dd1eb149a4fdd9254c41

olegklimov:
Had this remade.

PR #1057, opened 2017-04-05
refactor setup.py

cprogrammer1994:
tested on windows with python 3.5 and 3.6


setup.py
```diff
@@ -1,424 +1,228 @@
...
+}))
+
+sources.extend(rglob('examples/OpenGLWindow/**/*.cpp', exclude = {
+	'EGLOpenGLWindow.cpp',
+	'MacOpenGLWindow.cpp',
+	'Win32OpenGLWindow.cpp',
+	'Win32Window.cpp',
+	'X11OpenGLWindow.cpp',
+}))
+
```
cprogrammer1994: produces a list of 329 files (same as the previous ones)

cprogrammer1994:
I use [this method](https://github.com/cprogrammer1994/ModernGL/blob/master/setup.py#L4) to get the platform, but I am not able to test on Mac.

erwincoumans:
Thanks for some of the work. I'd rather use the original 
"from distutils.util import get_platform". Why does it need to be changed?

setup.py
```diff
@@ -1,424 +1,228 @@
-
-from setuptools import find_packages
+from setuptools import setup, Extension, find_packages
```
cprogrammer1994: [distutils_vs_setuptools](https://www.reddit.com/r/Python/comments/3882rh/distutils_vs_setuptools/)

cprogrammer1994: [PyPA sample uses setuptools as well](https://github.com/pypa/sampleproject/blob/master/setup.py#L9)

> setuptools provides enhancements to distutils, and is much more commonly used than just using distutils by itself. _[from distutils docs](https://packaging.python.org/key_projects/#distutils)_

erwincoumans:
I am maintaining setup.py and prefer to make the changes myself. Thanks for your contibution though, I'll have a look at the changes and may manually improve the setup.py eventually!

PR #1059, opened 2017-04-05
make the URDF and the minitaur.py consistent with the real robot

jietan:
I carefully checked the URDF file and found that for the left-half of the minitaur, the L and R motor is flipped. This is the cause that my minitaur_test.py does not run correctly.


erwincoumans:
Thanks for tracking this down! I'll check it soon and merge it.

PR #1075, opened 2017-04-17
forgotten setup.py build part

devnexen:


erwincoumans:
Thanks. Can you tell which BSD distribution + version you use?

Secondly, does pybullet build
(when using python setup.py build or python setup.py install, from the root of Bullet)

devnexen:
I use OpenBSD and FreeBSD. Yes.

erwincoumans:
thanks for the patch!

PR #1076, opened 2017-04-17
Modify the link index when computing Jacobian.

YunfeiBai:
Inverse dynamics Jacobian uses zero-based indexing of bodies, not starting from -1 for base.

Also add a test for end effector orientation IK. Can set joint damping to control using which joints to achieve the IK target.


erwincoumans:
Thanks a lot for tracking this down and the fix!

PR #1078, opened 2017-04-19
Random improvements 2

olegklimov:
1. Memory leaks. Not sure I've found all of them, but plenty.

2. ClientC_API installation. I argue it's a proper library, not duplicating existing libraries. Here is `ldd` on linux:

```bash
/usr/local/lib$ ldd libPhysicsClientC_API.so.2.87 
	libBulletDynamics.so.2.87 => /usr/local/lib/libBulletDynamics.so.2.87 (0x00007f415c93e000)
	libBulletCollision.so.2.87 => /usr/local/lib/libBulletCollision.so.2.87 (0x00007f415c60b000)
	libBulletInverseDynamics.so.2.87 => /usr/local/lib/libBulletInverseDynamics.so.2.87 (0x00007f415c3f2000)
	libLinearMath.so.2.87 => /usr/local/lib/libLinearMath.so.2.87 (0x00007f415c1d0000)
```
It contains things like URDF load, and for physics it depends on usual Bullet libraries. Size is 1.6M on linux, which is large (it contains tinyxml, tiny render, ...) but not too large. Also tested on Mac.

3. And more MJCF support ("axisangle") and little code fixups.

### DIFF
--- a/examples/Importers/ImportMJCFDemo/BulletMJCFImporter.cpp
+++ b/examples/Importers/ImportMJCFDemo/BulletMJCFImporter.cpp
@@ -630,11 +630,7 @@ struct BulletMJCFImporterInternalData
 		bool handledGeomType = false;
 		UrdfGeometry geom;
 
-		
-
-		const char* gType = link_xml->Attribute("type");
 		const char* sz = link_xml->Attribute("size");
-		const char* posS = link_xml->Attribute("pos");
 		int conDim = defaults.m_defaultConDim;
 
 		const char* conDimS = link_xml->Attribute("condim");
@@ -707,6 +703,7 @@ struct BulletMJCFImporterInternalData
 			geom.m_localMaterial.m_name = rgba;
 		}
 
+		const char* posS = link_xml->Attribute("pos");
 		if (posS)
 		{
 			btVector3 pos(0,0,0);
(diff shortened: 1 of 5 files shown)


erwincoumans:
Thanks for the contribution. The PhysicsClientC_API will be named BulletRobotics and all files related to this library need to be moved into the Bullet/src folder. So no CMakeList.txt that points to the examples or Extras folder. Can you remove the CMakeLists.txt, so I can accept the other changes? (github has no way to partially accept a pull request). 

We have zero memory leaks in our tests, but our tests assume zero issues with the URDF files (we don't test MJCF at the moment). It looks like there are leaks for URDF files with issues (duplicate links, duplicate joints etc), good to sort those indeed.

By the way, do you consider using pybullet at all, or do you re-create your own Python bindings?

olegklimov:
BulletRobotics -- sounds great!

pybullet -- yes roboschool has its own bindings, because pybullet can be a limiting factor in terms of what can be done. Obviously physics library should be at C++ level, with minimum work left to python (because speed), also more advanced features require python binding anyway.

memory leaks -- no, ordinary work with MJCF leaks memory, not error cases. You see I've deleted destructor from urdf-only code and placed it into header, that's the main source of the leak.

OK let me rebase this...

erwincoumans:
The code doesn't compile, will you replace the assert.h by btAssert?
Thanks!

olegklimov:
Fixed asserts, friction bug I've mentioned, also link names (to distinguish links by name on client, these names are visible.)

erwincoumans:
Great, thanks for your contribution!

Issue #1066, opened 2017-04-10
pybullet: How to get names of all links


benelot:
I am trying to a dictionary mapping from the name of a link to an object holding the link. The link class I already have (it basically runs methods of pybullet and stores the bodyid and the linkIndex of the link). However, I can not get proper names of the links using getBodyInfo() for all the bodyId, linkIndex pairs. How can I do this? If it does not exist, I will help to implement it in pybullet as it is very handy if you want to access links known to you by name.

erwincoumans:
See getJointInfo in the pybullet quickstart guide (http://pybullet.org), it has names

benelot:
Yes, for joints, but not for links. If you load a MJCF, there are link names (<body name="torso" ...) and joint names. (<joint name="rot2"...) I already found the joint names to access the joints via dictionary. But I need the link names too. That is pretty standard practice in MuJoCo to use the names.

benelot:
How would you like to make the link names accessible? I would suggest to introduce a getLinkInfo(bodyIndex, linkIndex) which could return all sorts of things like shapetype, scaling and my highly desired link name.

benelot:
Alternatively I could just append it to the jointInfo as another field. I just did this in a separate branch for my purposes since the linkName is already properly stored, but not accessible.

benelot:
Just give me a hint and I will make a PR.

PR #1082, opened 2017-04-25
Add link name to getJointInfo(...).

benelot:
PR to #1066. Just added the linkName to getJointInfo(...). Is this sufficient or should I create a new command to get link specific info such as getLinkInfo(...)?

### DIFF
--- a/examples/pybullet/pybullet.c
+++ b/examples/pybullet/pybullet.c
@@ -2083,7 +2083,7 @@ static PyObject* pybullet_getJointInfo(PyObject* self, PyObject* args, PyObject*
 
 	int bodyUniqueId = -1;
 	int jointIndex = -1;
-	int jointInfoSize = 12;  // size of struct b3JointInfo
+	int jointInfoSize = 13;  // size of struct b3JointInfo
 	b3PhysicsClientHandle sm = 0;
 	int physicsClientId = 0;
 	static char* kwlist[] = {"bodyUniqueId", "jointIndex", "physicsClientId", NULL};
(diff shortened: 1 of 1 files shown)


erwincoumans:
Thanks, looks good to me.

ShravanTata:
@benelot It would be much better to have getLinkInfo to extract link names rather than from getJointInfo. Could we reopen this issue to fix this? Thanks

erwincoumans:
Does the linkname returned by getJointInfo not work for you? If not, reopen and describe a repro case.

ShravanTata:
Found an important information that helped me solve this issue.
In bullet linkIndex == jointIndex

Using the above, I currently have the following solution below

_link_name_to_index = {p.getBodyInfo(model_id)[0].decode('UTF-8'):-1,}
        
for _id in range(p.getNumJoints(model_id)):
	_name = p.getJointInfo(model_id, _id)[12].decode('UTF-8')
	_link_name_to_index[_name] = _id

submagr:
`In bullet linkIndex == jointIndex`

But a robot can have more links than the joint, right? For example: l1---j1---l2. This robot has two links connected by a joint. The reason I ask this question is if I want to do calculateInverseKinematics for l2, pybullet raises index out of range error.

ignacioDeusto:
It misses the root link (at least in my case).
For instance, in your example:
l1 -> missed
j1 -> joint index 0
l2 -> link index 0
When checking for collisions, l1 is expressed as -1 link, but there's no link index -1, if you try to check the -1 index link state it crashes.
Missing the first link ensures that the number of joints equals the number of links. The main limitation with the current approach is that you can't get the root link name, also this one on one relation between joints and links feels weird for newcomers.

ShravanTata:
`p.getBodyInfo(model_id)[0].decode('UTF-8')` returns the base link name. [Code above](https://github.com/bulletphysics/bullet3/pull/1082#issuecomment-516545497) should return all link names including the base/root link

PR #1083, opened 2017-04-26
Use float when clipping the shadow map index, because the index befor…

YunfeiBai:
…e clipping can be out of the range for int.


erwincoumans:
great, thanks for tracking this down!

PR #1085, opened 2017-04-28
A simple trained DDPG agent for minitaur walking

jietan:


erwincoumans:
Thanks for the contribution! It would be great to have some example Youtube video of this too :-)

PR #1087, opened 2017-05-01
Remove pointless if

gregeryb:
"if (m_angularFactor)" test evaluates m_angularFactor as (float*)operator() {return &m_floats[0]} which is always non-zero


erwincoumans:
It is suppose to be an early out, probably not a big deal. The intention was to test if the factor was non-zero, assuming it was a scalar, but it is a vector indeed.  if (m_angularFactor.length2()) would keep the original intention.

AndrewMeadows:
I tested it and indeed: a `btVector3` always evaluates to 'true' even when it is of zero length.

The `if` really could be eliminated.  Most people aren't using a zero `m_angularFactor` for most of their rigid bodies, so it would probably be a net optimization to remove the check entirely.

PR #1090, opened 2017-05-03
Collider object not necessary

gregeryb:
Collider object not necessary without DBVT_BP_ACCURATESLEEPING defined


erwincoumans:
indeed, unused data. Thanks for the help!

PR #1091, opened 2017-05-04
Add APIs to reset object mass, lateral friction coefficient, and to get user constraint id.

YunfeiBai:


erwincoumans:
Thanks, look good to me!

PR #1099, opened 2017-05-06
getSolveSwingLimit() returns wrong value

gregeryb:
Should return m_solveSwingLimit instead of m_solveTwistLimit


erwincoumans:
Copy/paste bug indeed, thanks for the contribution.

PR #1101, opened 2017-05-06
Running humanoid

olegklimov:
Sorry for broken self-collision, but Humanoid is supposed to have self collision.

So it's running at about 300 fps, numSubSteps and numSolverIterations has little effect on fps. At numSubSteps=1 it crashes often. Please ask for suboptimal humanoid that self-collides every step and crashes more spectacular, if you need it.


erwincoumans:
Thanks for creating a clean reproduction of the humanoid using TensorFlow!

At a first glance, biggest issue is you are using torque control, which very easily leads to 'explosions' due to insufficient damping. Position control or velocity control will be much more stable and easier (which is what we are doing for most robots). Our MuJoCo importer doesn't load or apply damping. You can easily introduce some damping force by using a non-zero force when setting the VELOCITY_CONTROL line. This is similar to the MuJoCo implicit damping (although using a velocity motor instead of modifying the mass matrix).


dampingForce = 1
for j in range( p.getNumJoints(human) ):
		...
		p.setJointMotorControl2(human, j, controlMode=p.VELOCITY_CONTROL, force=dampingForce)

I'll spend a bit more time on it upcoming days.

erwincoumans:
Thanks for sharing, that is a great locomotion example with pybullet + TensorFlow .
I just implemented your 'don't collide with any parent in C-API and pybullet:
b3LoadMJCFCommandSetFlags  and loadMJCF("filename.xml", flags=pybullet.URDF_USE_SELF_COLLISION_EXCLUDE_ALL_PARENTS)

erwincoumans:
I added a new flag that uses your self-collision filtering and modified your humanoid_running.py to work with plain pybullet (unmodified). Pull request here:
https://github.com/bulletphysics/bullet3/pull/1115

See the demo here:
https://github.com/erwincoumans/bullet3/blob/master/examples/pybullet/tensorflow/humanoid_running.py

Also, I added a few extra humanoids. Since the ground plane is part of the model, having 3 times the same ground plane makes the simulation instable, so I removed 2 of them using removeBody.
Thanks a lot for your contribution!

PR #1102, opened 2017-05-07
write documentation for DBVH

AndySomogyi:
I've added a bunch of documentation for the dynamic bounding volume hierarchy. I've still got a bit more research to do to finish documenting the remaining methods, but I think these changes will really help out anyone wanting to use the DBVH. 

Is there any real rationale using the B3_DBVT_IPOLICY macro? The code would break horribly if the macro would be defined to anything other than 'ICollide& policy', and I think having the macro in there makes the code more confusing. I think it would be a fair bit clearer if this macro were replaced with 'ICollide& policy', and would make the Doxygen docs look better.

### DIFF
--- a/src/Bullet3Collision/BroadPhaseCollision/b3DynamicBvh.h
+++ b/src/Bullet3Collision/BroadPhaseCollision/b3DynamicBvh.h
@@ -11,7 +11,7 @@ subject to the following restrictions:
 1. The origin of this software must not be misrepresented; you must not claim that you wrote the original software. If you use this software in a product, an acknowledgment in the product documentation would be appreciated but is not required.
 2. Altered source versions must be plainly marked as such, and must not be misrepresented as being the original software.
 3. This notice may not be removed or altered from any source distribution.
-*/
+ */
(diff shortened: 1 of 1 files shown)


AndySomogyi:
Hey @erwincoumans, I'm starting to use Bullet in my position based dynamics research, and how I get to understand code is I read through it, and document my findings. I'm adding these as doxygen formatted comments in the headers, and I've made sure not to make any non-whitespace/comment changes. 

I'm also going through and documenting the time stepping procedure, so I'm going to write up documentation here also.

erwincoumans:
Most of the Bullet/src/Bullet3* folders are not really used yet, the active version is still Bullet/src/BulletCollision and Bullet/src/BulletDynamics and Bullet/src/LinearMath. Bullet rigid body dynamics is based on a velocity level LCP solver. Only the cloth/soft body dynamics in Bullet/src/BulletSoftBody is using a similar approach to position based dynamics.

Also, I don't have many cycles to accept many pull request, so you may just focus on adding documentation in your fork for now.

PR #1110, opened 2017-05-09
Fix the z-buffer fighting by adding back face culling. Optimize tinyrender by precompute matrix multiplication in vertex shader.

YunfeiBai:


erwincoumans:
Thanks for improving this!

PR #1116, opened 2017-05-11
Rewrite collision filtering code for C API

olegklimov:
The problem: user posts request with filter on BodyA and LinkA.

Existing code matches BodyA -or- BodyB against BodyA filter, then it filters LinkA.

Result is BodyB and LinkA, that is not what filter was supposed to do. Valid result is BodyA, LinkA or BodyB, LinkB (later swapped for user so it matches the filter).


olegklimov:
...and self-collision needs a fix too. For humanoid, legs collided fine, but hands penetrate the body. After some debugging, I came up with this.

erwincoumans:
Thanks again, I'll processed it soon.

PR #1119, opened 2017-05-12
fix typo in btHingeConstraint method name

AndrewMeadows:
I spotted a simple typo: Velosity --> Velocity, then decided to add a bit of minor cleanup.


erwincoumans:
Thanks for the help, all bits matter 👍

PR #1129, opened 2017-05-18
remove sonnet dependency

jietan:
remove sonnet dependency.


erwincoumans:
Great, thanks for the contribution!

ahundt:
As much as I like removing the sonnet dependency, loading a mysterious serialized metagraph is much less informative! Perhaps having a single SimpleAgent that uses sonnet if it is available and falls back on the serialized metagraph otherwise is reasonable, so people with sonnet can see how the setup works and people without can run it.

The way to handle fallbacks is something like the [following example from keras](https://github.com/fchollet/keras/blob/master/keras/utils/vis_utils.py):

```python

try:
    # pydot-ng is a fork of pydot that is better maintained.
    import pydot_ng as pydot
except ImportError:
    # Fall back on pydot if necessary.
    try:
        import pydot
    except ImportError:
        pydot = None
```

erwincoumans:
Thanks Andew, Jie addressed it here: https://github.com/bulletphysics/bullet3/pull/1142

Issue #1107, opened 2017-05-09
MacOS homebrew installer - request for update with support for pybullet


ahundt:
I'm trying to update the [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core) repository with [**this new bullet.rb install script**](https://github.com/ahundt/homebrew-core/blob/bullet/Formula/bullet.rb), which is modified to support pybullet [with these changes](https://github.com/ahundt/homebrew-core/commit/fd915407ba908b709f5d711a4529ac0bac8d9320) in a [ahundt/homebrew-core bullet branch](https://github.com/ahundt/homebrew-core/tree/bullet), but there appears to be bugs in bullet's own CMake script for setting Python and NumPy variables on OSX, triggering CMake build error and selecting the incorrect python library version.

I attempt to install with the following command:
```
brew uninstall bullet --ignore-dependencies; brew install bullet --with-demo --with-double-precision --with-framework -v
```

The formula (aka ruby install script) repository should be in the `/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula` folder if [homebrew](https://brew.sh/) is installed.

My updated version of the script generates this cmake command:
```
cmake
..
-DCMAKE_C_FLAGS_RELEASE=-DNDEBUG
-DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG
-DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/bullet/2.86.1_1
-DCMAKE_BUILD_TYPE=Release
-DCMAKE_FIND_FRAMEWORK=LAST
-DCMAKE_VERBOSE_MAKEFILE=ON
-Wno-dev
-DHAVE_CLOCK_GETTIME:INTERNAL=0
-DINSTALL_EXTRA_LIBS=ON
-DBUILD_UNIT_TESTS=OFF
-DUSE_DOUBLE_PRECISION=ON
-DBUILD_BULLET2_DEMOS=ON
-DBUILD_PYBULLET_MAC_USE_PYTHON_FRAMEWORK=ON
-DBUILD_PYBULLET=ON
-DBUILD_PYBULLET_NUMPY=ON
-DBUILD_PYBULLET_CLSOCKET=ON
-DBUILD_PYBULLET_ENET=ON
-DPYTHON_EXECUTABLE=/usr/local/opt/python/bin/python
[...]

PR #1131, opened 2017-05-21
cmake python config bugs fixed

ahundt:
- Python and pybullet remain optional with these changes.
- A specific python version can now be selected by setting PYTHON_VERSION_PYBULLET.
- The python library version must now match the interpreter version exactly.
- If all required python dependencies are found, pybullet is now enabled by default.

Changes incorporate the following BSD licensed cmake code:

https://github.com/BVLC/caffe/blob/32bf5c7ad804ad683aa5ea9382209e9284451e5f/CMakeScripts/FindNumPy.cmake
https://github.com/NikolausDemmel/CMake/pull/2/files
https://github.com/Kitware/CMake/blob/86578eccf2e82286248796bad1032cd0e3a5e1e2/Modules/SelectLibraryConfigurations.cmake

This addresses many of the cmake issues blocking https://github.com/bulletphysics/bullet3/issues/1107


erwincoumans:
Thanks for the work on this. I think it is better to disable the BUILD_PYBULLET by default, unless the user manually enables it, since I really want Bullet to build out-of-the-box, undisturbed by broken python installations. We can leave it switched on if Python versions match for now. If people report broken builds, we switch it off.

ahundt:
Okay, I tried to be aware of that concern by. It should try and ensure if any problems are detected python automatically switches off, standards warnings are printed, and the cmake configuration + build continues & succeeds without python.

erwincoumans:

erwincoumans:
People report issues with cmake and python, and the continuous integration also fails.
https://github.com/bulletphysics/bullet3/issues/1322
We need to have a more reliable way for cmake to locate python

PR #1133, opened 2017-05-21
fix a crash in example browser when enabling multi-threading

lunkhound:
By default RTTI is disabled, so enabling multi-threading in the example browser crashes on a dynamic cast.
Changed it to a static cast and moved the cast inside the dynamics world constructor (which in the parent class just created the object being cast, so the type is known).


erwincoumans:
Thanks, indeed the code should work without RRTI. Thanks for the contribution!

PR #1135, opened 2017-05-21
disable Grasp Soft Body/SoftBody coupling demo unless USE_SOFT_BODY_MULTI_BODY_DYNAMICS_WORLD is defined, fixed related hang in experimental 'loadBunny' command

erwincoumans:


ahundt:
I can confirm it doesn't crash now! Non crashing version is definitely experimental as expected. :-)

PR #1142, opened 2017-05-23
add back sonnet dependency. If sonnet is not installed, fall back to …

jietan:
add back sonnet dependency. If sonnet is not installed, fall back to simpleAgent that does not need sonnet.


erwincoumans:
Thanks, LGTM,

PR #1144, opened 2017-05-23
Bullet 2 threading refactor: moved parallel-for calls into core libs

lunkhound:
This PR basically moves code around to make it easier for users to integrate multi-threaded Bullet into a project and maintain it. There isn't too much new here.
What's new:
  - an interface class (btIParallelForBody) that can be subclassed to express work to be done in parallel
  - an interface class (btITaskScheduler) that can be subclassed to execute work in parallel

Using those interfaces all of the multi-threading that was being done in the example browser has been moved into the core libs (but only in classes with a suffix "Mt"). All of the usual single-threaded classes are untouched here.
With this change, it will be far easier to use bullet CPU multithreading, just use the 'Mt' version of the physics world, and call btSetTaskScheduler() to select which task scheduler to use.

I've tested this on Windows 10/Cmake/MSVC 2015/32-bit.


erwincoumans:
I'll have a look, but I'm not eager to move the multi-threading code in Bullet/src, unless we can make it work with the Bullet's own multithreading handling (b3PosixThreadSupport, b3Win32ThreadSupport), and possibly users own multi-threading handling. Once I have a bit of time, I'll look into it. Basically allow the user to execute jobs, using their own threading mechanism (in addition to allowing TBB, OpenMP etc).

lunkhound:
With this change, you can still use any threading system. All you need to do is subclass btITaskScheduler and implement the parallelFor virtual function. Then call btSetTaskScheduler() with an instance of the subclass, and that's it.

I'm currently using it in my personal project with my own custom job manager.

Would you prefer for me to move the TBB, OpenMP, etc related code out of btThreads.cpp and back into the example browser? It would be a pretty trivial change. I only included it in there as a convenience for users to get multithreading up and running without having to subclass anything.

ahundt:
Might be worth considering an alternative to subclassing. What if the other threading API you're using required subclassing too? Things get messy pretty quickly that way.

erwincoumans:
I'd like to create a bridge between your multi-threading abstraction and b3PosixThreadSupport, b3Win32ThreadSupport that will be enabled by default (since pthreads and Win32 threads are pretty much always available).

lunkhound:
@ahundt The ITaskScheduler interface is only called from Bullet internals so regardless of what the other threading API is it shouldn't require multiple inheritance or anything like that. I've implemented the ITaskScheduler interface for 4 different threading APIs and usually all that is required is to define an "Adapter" object to convert from the function signature the API expects to the signature of the passed in IParallelForBody object.  [For example](https://github.com/lunkhound/bullet3/blob/pr-threading-refactor/src/LinearMath/btThreads.cpp#L180)

lunkhound:
@erwincoumans I like that idea. Should I add this "bridge" to this PR?
From skimming the code in MultiThreadingExample.cpp it looks like it should be fairly easy.

ahundt:
ok sounds good, thanks for the details

erwincoumans:
>> @erwincoumans I like that idea. Should I add this "bridge" to this PR?
Yes, that would be nice. 

>> Would you prefer for me to move the TBB, OpenMP, etc related code out of 
>> btThreads.cpp and back into the example browser? 

As long as they are disabled using some #ifdef, it is probably fine to leave them there. I'll get back to you later once I have some time to look at the changes/pull request.

lunkhound:
OK I'll see what I can do.

One thing I'm wondering is where to put all the *ThreadSupport source files. LinearMath?

And I suppose I'll also have to change the "b3" prefixes back to "bt" etc.

lunkhound:
@erwincoumans So I've got a task scheduler implemented on the ThreadSupportInterface working pretty well now (only tested on Windows though). However, I've noticed a few issues with the ThreadSupport implementations:

- The win32 version was setting the thread affinity mask of all threads to 2. This resulted in terrible performance as all the threads were trying to run on the same processor. I've commented that out for now, and the performance improved greatly.
- Why is the setNumTasks() function there? It does nothing in the Posix version and almost nothing on the win32 side
- The win32 version uses WaitForMultipleObjects in a number of places which is limited to 64 objects. So it will break if it tries to use more than 64 threads -- and there are CPUs now with more hardware threads than that
- I find the naming of getNumTasks() confusing. Seems like it should be called getNumThreads()

It just seems like the interface could be cleaned up a bit. However I don't want to expand the scope of this PR too much, so I'm leaving the interface alone.

For now I've added the task scheduler source (btTaskScheduler.h/cpp) in the same directory as the thread support files (examples/Multithreading)

examples/Utils/ChromeTraceUtil.cpp
```diff
@@ -184,7 +184,7 @@ void MyEnterProfileZoneFunc(const char* msg)
 		return;
 #ifndef BT_NO_PROFILE
 	int threadId = btQuickprofGetCurrentThreadIndex2();
-	if (threadId<0)
+	if (threadId<0 || threadId >= BT_QUICKPROF_MAX_THREAD_COUNT)
```
erwincoumans: Did you run it on a CPU with >64 threads? Or does the thread index grow over 64 for other reasons?

lunkhound:
In the example browser with one of the multithreaded demos, if you select
certain task schedulers (TBB for one) and then play with the "thread count"
slider it ends up creating and destroying many threads. And this can cause
the thread index to exceed 64 even though only a small number of threads
are being used.
I think it can be fixed by having the example browser reset the thread
indices when switching schedulers or adjusting the thread count.

lunkhound:
I'm thinking it would be best to fix the thread index problem with the example browser in another PR since it isn't directly related to this PR. The problem has been around for a long time, and I think fixing it will muddle up this PR.
I didn't really intend to include that change in ChromeTraceUtil.cpp in this PR. Do you think I should back out the change to ChromeTraceUtil.cpp?

erwincoumans:
It is ok to leave the improvement to ChromeTraceUtil.cpp in there. Somehow I cannot apply this patch locally, making it a bit harder to review. Let's apply it and I make adjustments later, if need. 
Thanks!

erwincoumans:
Fixed some minor compile issue when using premake:
https://github.com/bulletphysics/bullet3/pull/1166

PR #1145, opened 2017-05-24
Improve debug text/line rendering, can be use to draw frames and text  in local coordinate of an object / link.

erwincoumans:


erwincoumans:
mmm, seems to have an issue with Mac OSX/retina, need to look into that.

PR #1153, opened 2017-05-28
Zeroed axis in setupTorsionalFrictionConstraint

gregeryb:
Not sure if this was the desired behavior. The zeroing of normalAxis1 seems to be the cause of spheres spinning in place while stopped. Also the two dot products for contactNormal1 and contactNormal2 would always be zero as well.


erwincoumans:
No, it is intentional to use normalAxis(0,0,0) for the linear component: the torsional friction works fine, and shouldn't have any linear effect. The angular axis is using normalAxis1.

PR #1166, opened 2017-06-02
btThreadsAreRunning wasn't defined if BT_THREADSAFE was not defined, …

erwincoumans:
…causing compile errors using premake


lunkhound:
By the way #1168 also fixes the issue of btThreadsAreRunning being undefined. May cause a merge conflict.

PR #1167, opened 2017-06-03
Swap yaw and pitch, and use setEulerZYX in camera computation. Add camera yaw, pitch, distance, and target to getDebugVisualizerCamera. Add rgba color reset for TinyRender.

YunfeiBai:


erwincoumans:
Thanks a lot for sorting out this SimpleCamera yaw/pitch mess!

PR #1168, opened 2017-06-04
fix various multithreading issues with thread indexes

lunkhound:
With multithreading enabled, there is a bug where if you resize the thread pool up and down repeatedly, the index returned by GetCurrentThreadIndex() can increase past BT_MAX_THREAD_COUNT which leads to a number of issues.
Added a compile option (disabled by default) called BT_DETECT_BAD_THREAD_INDEX which checks for thread-index collisions. There is some commentary added about how thread indexes are kept in a valid range and avoids collisions.
The profiling system has a similar mechanism for assigning thread indexes, only it isn't threadsafe and when increasing the thread count rapidly I observed some race-induced crashes. So now if multithreading is enabled it just uses the threadsafe one.
Also I reorganized btThreads.h to avoid having things declared/defined dependent on BT_THREADSAFE as that has caused a lot of compilation breakage. Now all symbols/functions/etc are always declared and defined.


erwincoumans:
thanks, looks good to me.

PR #1169, opened 2017-06-05
example browser: fix 'clamp-to-integer' on GUI-sliders

lunkhound:
Improves the feel on GUI sliders that are flagged with clamp-to-integer


erwincoumans:
looks good to me, thanks!

PR #1182, opened 2017-06-10
SequentialImpulseConstraintSolver: fix bugs with SOLVER_SIMD flag

lunkhound:
The SOLVER_SIMD flag should only control whether SIMD optimizations are used or not. However it has unexpected interactions with the SOLVER_INTERLEAVE_CONTACT_AND_FRICTION_CONSTRAINTS flag.

Bug 1: If both SIMD and INTERLEAVE flags are on, rolling friction is not applied
Bug 2: If SIMD flag is off, the INTERLEAVE flag is ignored

The reason for the discrepancies can be seen in the solveSingleIteration() function. There are 2 separate code paths, one for SIMD and one for non-SIMD. It looks like one path was copied and pasted from the other, then modified, and it seems that the SIMD path has acquired some later changes that weren't applied to the non-SIMD path.
With this PR, rather than having 2 paths through solveSingleIteration, there is just one path (with ~80 lines of redundant code removed), and the various calls go through function pointers that have been set to point to SIMD or non-SIMD implementations.

I tried to be as minimal as possible with the changes, avoiding any formatting changes that might confuse 'diff'.

### DIFF
--- a/src/BulletDynamics/ConstraintSolver/btSequentialImpulseConstraintSolver.cpp
+++ b/src/BulletDynamics/ConstraintSolver/btSequentialImpulseConstraintSolver.cpp
@@ -269,36 +269,28 @@ static btSimdScalar gResolveSingleConstraintRowLowerLimit_sse4_1_fma3(btSolverBo
 
 btSimdScalar btSequentialImpulseConstraintSolver::resolveSingleConstraintRowGenericSIMD(btSolverBody& body1,btSolverBody& body2,const btSolverConstraint& c)
 {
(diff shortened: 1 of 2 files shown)


erwincoumans:
Thanks! I agree, this is good to fix. Give me some more time to review it and merge it.

lunkhound:
Cool. If this passes review, I can do a similar fix for the NNCG solver too.

erwincoumans:
Thanks for your improvements, LGTM.

PR #1184, opened 2017-06-12
fix memcpy of depth and segmentation images to numpy arrays

mbreyer:
When the image data is copied to NumPy arrays, the wrong byte count was used for the depth and segmentation values.


erwincoumans:
That look like a bug indeed, thanks for the fix/pull request!

PR #1185, opened 2017-06-13
fix uninitialized erp in pybullet_setPhysicsEngineParameter

mbreyer:
Uninitialized error reduction parameter in pybullet_setPhysicsEngineParameter lead to unwanted calls to b3PhysicsParamSetDefaultNonContactERP (for example if erp=0) which produces unexpected simulation behavior.


erwincoumans:
well spotted, thanks for fixing the bug!

PR #1052, opened 2017-04-01
Fix some bullet build warnings

nicolaichuk:
fix some bullet build warnings:

```
linearmath\btidebugdraw.h(169): 'argument': truncation from 'double' to 'const btScalar'
linearmath\btidebugdraw.h(170): 'argument': truncation from 'double' to 'const btScalar'
linearmath\btidebugdraw.h(171): 'argument': truncation from 'double' to 'const btScalar'

linearmath\btvector3.cpp(179): '=': conversion from 'size_t' to 'long', possible loss of data
linearmath\btvector3.cpp(184): '-=': conversion from 'size_t' to 'unsigned long', possible loss of data
linearmath\btvector3.cpp(427): '=': conversion from 'size_t' to 'long', possible loss of data
linearmath\btvector3.cpp(564): '=': conversion from 'size_t' to 'long', possible loss of data
linearmath\btvector3.cpp(569): '-=': conversion from 'size_t' to 'unsigned long', possible loss of data
linearmath\btvector3.cpp(815): '=': conversion from 'size_t' to 'long', possible loss of data

linearmath\btquickprof.cpp(221): '*=': conversion from 'double' to 'LONGLONG', possible loss of data
linearmath\btquickprof.cpp(290): '=': conversion from 'unsigned __int64' to 'unsigned long', possible loss of data

bulletcollision\narrowphasecollision\btgjkcollisiondescription.h(33): 'initializing': truncation from 'double' to 'btScalar'
bulletcollision\narrowphasecollision\btmprpenetration.h(44): 'initializing': truncation from 'double' to 'btScalar'
bulletcollision\narrowphasecollision\btmprpenetration.h(152): '=': conversion from 'double' to 'float', possible loss of data
bulletcollision\narrowphasecollision\btmprpenetration.h(156): '=': conversion from 'double' to 'float', possible loss of data
bulletcollision\narrowphasecollision\btmprpenetration.h(157): '=': conversion from 'double' to 'float', possible loss of data
bulletcollision\narrowphasecollision\btmprpenetration.h(566): 'argument': conversion from 'size_t' to 'int', possible loss of data
bulletcollision\narrowphasecollision\btmprpenetration.h(570): 'argument': conversion from 'size_t' to 'int', possible loss of data

bulletcollision\narrowphasecollision\btgjkpairdetector.cpp(38): 'initializing': truncation from 'double' to 'btScalar'

[...]

PR #1191, opened 2017-06-16
Fix double to float conversion warning

1vanK:


erwincoumans:
Now compile the code using BT_USE_DOUBLE_PRECISION (that is the new default for pybullet and our usage) and you introduce a warning there...

1vanK:
Hm, in VS no warnings for conversion float to double

iSLC:
I suppose if you want to got the ugly route you could do something like:
```
#ifndef BT_USE_DOUBLE_PRECISION
    #define BT_F(v) v ## f
#else
    #define BT_F(v) v
#endif
```

And then:
```
btVector3(BT_F(1.0), BT_F(0.3), BT_F(0.3))
```

erwincoumans:
You can already do btVector3(btScalar(1),btScalar(0.3),btScalar(0.3)); 
Equally ugly and overly verbose indeed, hence the warnings are still there.

lunkhound:
I really don't see the point of such warnings for compile-time numerical constants. In fact, neither MSVC, GCC, or Clang are generating such spurious warnings. These warnings don't appear in the Appveyor (MSVC 2015), or Travis (GCC, Clang) logs.
If only some old version of MSVC generates this warning, perhaps a better fix is to just disable that warning for the offending compiler version.

1vanK:
Version of my VS is 2015...

lunkhound:
I'm using Cmake/MSVC 2015/Win32 and I'm not getting those warnings.
Also if you look at the appveyor logs: [https://ci.appveyor.com/project/erwincoumans/bullet3/build/1.0.1934](url)
which is using MSVC 2015, I'm not seeing those warnings. All I see is 16 occurances of C4838, and those all look like legitimate warnings.
Hmmm.. I see that in my Cmake built Bullet project, certain warnings are disabled. It has warning C4244 (and also C4267) disabled.
Are you using premake?

1vanK:
> Are you using premake?

cmake

lunkhound:
OK, now I can see where this is coming from. The warnings are coming from compiling external (non-bullet) source files that include that one bullet header.

In light of that, I agree this is a good change. I still think the warnings are pretty useless, but I also think it is kind of bad form for a third party library's headers to cause a spew of warnings.
Of all the ways to fix this, adding those 'f' suffixes is the least ugly option I can see.

1vanK:
Since when converting float to double there is no truncation, warning is also does not arise. So we can use float for all constants

nicolaichuk:
associated PR https://github.com/bulletphysics/bullet3/pull/1052

madebr:
This is fixed in current master.
This pr can be closed.

1vanK:
> You can already do btVector3(btScalar(1),btScalar(0.3),btScalar(0.3));
> Equally ugly and overly verbose indeed, hence the warnings are still there.

this is not same

https://stackoverflow.com/questions/55408191/is-literal-double-to-float-conversion-equal-to-float-literal

```
#include <utility>
#include <iostream>

using namespace std;

int main()
{
    float f1 =       1.0000000596046447753906250000000000000000000000000001f;
    float f2 = float(1.0000000596046447753906250000000000000000000000000001);

    if (f1 == f2)
        cout << "f1 == f2" << endl;
    else
        cout << "f1 != f2" << endl;
}
```

Output:

```
f1 != f2
```

PR #1194, opened 2017-06-17
ThreadSupportInterface: add platform independent create function

lunkhound:
This allows for creating a ThreadSupportInterface object without dealing with platform specific implementation details. Users can just call ThreadSupportInterface::create() and get the correct implementation for their platform.


lunkhound:
@erwincoumans a while back you had indicated you would like to see the btDefaultTaskScheduler class (which is layered on top of ThreadSupportInterface) integrated into the core bullet libs. Assuming that means moving ThreadSupportInterface into core libs also, I figured it would make sense to clean up the interface a bit. This PR is meant as a first step in that direction.
Are you still interested in moving TaskScheduler/ThreadSupport into core?

lunkhound:
Since #1579 has been merged, this PR is fairly pointless now. Closing.

PR #1203, opened 2017-06-21
add kukaCamGymEnv.py with camera observations (preliminary)

erwincoumans:
show camera position in example browser
disable per-vertex and per-fragment profile timings


ahundt:
Awesome! 👍 👍 👍

Issue #1209, opened 2017-06-24
Compiling bullet3 with python if your python setup is messed up


benelot:
Simply put, my cmake does not find python properly on its own. If you encounter this problem, just find the python interpreter which shows the appropriate python version on your console. For me this was 'python'. For you this might be python2.7 or similar. Then edit the following to the build_cmake_pybullet_double.sh file:

	cmake -DBUILD_PYBULLET=ON -DBUILD_PYBULLET_NUMPY=OFF -DUSE_DOUBLE_PRECISION=ON -DCMAKE_BUILD_TYPE=Release  ..

to

	cmake -DBUILD_PYBULLET=ON -DBUILD_PYBULLET_NUMPY=OFF -DUSE_DOUBLE_PRECISION=ON -DCMAKE_BUILD_TYPE=Release \
	-DPYTHON_INCLUDE_DIR=$(python -c "from distutils.sysconfig import get_python_inc; print(get_python_inc())")  \
	-DPYTHON_LIBRARY=$(python -c "import distutils.sysconfig as sysconfig; print(sysconfig.get_config_var('LIBDIR'))") ..

Replace the 'python' in '$(python -c' in the above lines with the interpreter name you found. This will the automatically resolve for the correct python directories and submit them to cmake. Best of luck!

cprogrammer1994:
Would not be better to provide wheels on PyPI?

cprogrammer1994:
I better say "easier" not "better" 😄

erwincoumans:
You should be able to use pip install pybullet indeed, that is recommended over using cmake etc.

https://pypi.python.org/pypi/pybullet

No need for wheels, pip should just install from source.

I upload some Mac OSX and Windows wheels sometimes. Would be nice to have an automatic way to generate wheels for most popular versions of python on Mac and Windows.

erwincoumans:
Benjamin, should we replace the line in build_cmake_pybullet_double.sh  file with your one? 

I would assume 'python' to be the default name in most cases.

benelot:
I can not yet say if this works in all cases, but in mine, it fixed the issue. We can change it and see how many other people's build system we break ;-) No, in fact it should work in general as it returns the correct path for that interpreter. It remains in question why the cmake search fails...

erwincoumans:
cmake fails to find packages so many times, especially on Windows. I spend a lot of time digging into some cmake files to see how it finds a particular package: cmake doesn't make it easy at all to find packages automagically.

[...]

PR #1211, opened 2017-06-25
Use explicit python paths when provided by python itself when using c…

benelot:
…make. This is the pull-request for #1209.

### DIFF
--- a/build_cmake_pybullet_double.sh
+++ b/build_cmake_pybullet_double.sh
@@ -2,7 +2,9 @@
 rm CMakeCache.txt
 mkdir build_cmake
 cd build_cmake
-cmake -DBUILD_PYBULLET=ON -DBUILD_PYBULLET_NUMPY=OFF -DUSE_DOUBLE_PRECISION=ON -DCMAKE_BUILD_TYPE=Release ..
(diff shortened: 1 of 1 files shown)


erwincoumans:
Thanks! I applied the patch, but it breaks the Mac OSX build. CMake finding libs is such a broken system, when the user provides the PYTHON_INCLUDES etc manually, it should NOT try to search for them again.

/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/System/Library/Frameworks/OpenGL.framework
-- Found PythonInterp: /usr/bin/python2.7 (found suitable version "2.7.10", minimum required is "2.7") 
-- Could NOT find PythonLibs: Found unsuitable version "3.6.0", but required is exact version "2.7.10" (found /Users/erwincoumans/anaconda3/lib)
Mac OSX Version is 10.12

benelot:
That is an interesting behaviour. So the same python environment delivers two python include paths that point to different versions of python? Then let us keep it the way it is and people that can not get it to run can just use my proposed fix.

erwincoumans:
Now travis is failing the build: https://travis-ci.org/bulletphysics/bullet3/jobs/273121440

erwincoumans:
And others also report failures in cmake python search, let's revert this:
https://github.com/bulletphysics/bullet3/issues/1322

PR #1219, opened 2017-07-03
NNCG solver: apply rolling friction consistently

lunkhound:
The NNCG solver has the same issues that were recently fixed in the SequentialImpulseConstraintSolver where the SOLVER_SIMD flag combined with the SOLVER_INTERLEAVE_FRICTION flag causes rolling friction to not be applied.
There are 2 commits here, the first adds some options to the MultiThreadedDemo which allows for easy testing of rolling friction -- a slider for rolling friction, as well as a slider for tilting the ground and a button to use spheres instead of boxes. The second commit fixes the issue.

So to verify the issue exists, follow these steps:
- apply the first commit
- Compile and run the example browser
- run the MultiThreadedDemo
- select the NNCG solver in the combobox
- set rolling friction to 0.2
- set ground tilt to 30 degrees
- enable the "spheres not boxes" button
- restart MultiThreadedDemo
- the spheres should spawn and come to rest
- enable the "Solver interleaved" button
- spheres should start rolling

Then to verify the issue is fixed, apply the second commit and redo those steps, and the spheres should stay put this time.
I tested it on Windows 10/Cmake/Win32/MSVC 2015.

### DIFF
--- a/examples/MultiThreadedDemo/MultiThreadedDemo.cpp
+++ b/examples/MultiThreadedDemo/MultiThreadedDemo.cpp
@@ -13,28 +13,35 @@ subject to the following restrictions:
 3. This notice may not be removed or altered from any source distribution.
 */
 
-#include "btBulletDynamicsCommon.h"
-#include "LinearMath/btQuickprof.h"
-#include "LinearMath/btIDebugDraw.h"
+#include "MultiThreadedDemo.h"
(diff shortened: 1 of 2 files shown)


erwincoumans:
Excuses for the huge delay, let's just merge this now.

lunkhound:
Thanks for the merge!

Issue #1213, opened 2017-06-29
btCreateCompoundFromGimpactShape leads to memory leak


Wugand:
The function btCreateCompoundFromGimpactShape is allocating much memory internally. (Which goes somewhere into the btCompoundShape). But this memory is not deallocated, when deleting the btCompundShape.

Analysis said the unreferenced heap space is allocated by:
`btBU_Simplex1to4* tet = new btBU_Simplex1to4(v0,v1,v2,rayTo);` (Line 77)
*It seams to me that this object is stored somewhere in "m_children" of btCompoundShape

Why is this memory doesn't get deallocated automatically?
Is there some way how I can deallocate this remains? 

Thanks!

(defined in: bullet3/src/BulletCollision/Gimpact/btCompoundFromGimpact.h)
https://github.com/bulletphysics/bullet3/blob/79d24db60f2c916ffcdaf8b074be622e2e5bc609/src/BulletCollision/Gimpact/btCompoundFromGimpact.h

erwincoumans:
Needs fixing indeed, pull request?

erwincoumans:
Thanks! Closing old issue.

PR #1221, opened 2017-07-03
Fix for memory leak at btCreateCompoundFromGimpactShape

Wugand:
This is an fix for issue #1213 

The btBU_Simplex1to4 ChildShapes are created internally, and don't get cleaned up.

Fix contains:
btCompoundFromGimpactShape inherited from btCompoundShape, which automatically delete this childShapes


erwincoumans:
Thanks for the contribution, that looks good to me.

PR #1223, opened 2017-07-04
Update Premake5 to current latest version 'premake-5.0.0-alpha11'

BillelAdam:
Updated Premake5 executable to the latest available version which is 'premake-5.0.0-alpha11' that supports Visual Studio 2017 projects directly instead of upgrading Visual Studio 2015 projects.


erwincoumans:
We just updated the premake5.exe, so let's close the issue.
(note we still use premake4.exe by default, the visual studio project files auto-convert to newer versions just fine)

Issue #1228, opened 2017-07-14
UBSan error with `getHash`


m-chaturvedi:
To reproduce, after instrumenting the library with UBSan, run the following example:

```
#include <bits/stdc++.h>
#include "LinearMath/btHashMap.h"
using namespace std;

int main() {
	btHashInt h1(34098);
	h1.getHash();
}
```

This gives the following output:
```
/home/chaturvedi/workspace/drake-distro/cmake-build-ubsan/install/include/bullet/LinearMath/btHashMap.h:106:58: runtime error: left shift of 1116265680 by 3 places cannot be represented in type 'int'
/home/chaturvedi/workspace/drake-distro/cmake-build-ubsan/install/include/bullet/LinearMath/btHashMap.h:106:98: runtime error: left shift of 1469381725 by 11 places cannot be represented in type 'int'
/home/chaturvedi/workspace/drake-distro/cmake-build-ubsan/install/include/bullet/LinearMath/btHashMap.h:106:89: runtime error: signed integer overflow: 1469381725 + 1478301695 cannot be represented in type 'int'
```
My guess is that the function knows what its doing, thus the suppression.
https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#disabling-instrumentation-with-attribute-no-sanitize-undefined

erwincoumans:
No volunteer shows up in fixing this hash function, so for now just suppress the warning.
If someone has a fix, re-open and provide a pull request.

PR #1229, opened 2017-07-14
Suppressing a Few UBSan Issues with getHash

m-chaturvedi:
re https://github.com/bulletphysics/bullet3/issues/1228


erwincoumans:
Thanks for the report of UBSan issues, I've never used UBSan but will try it out too.
For the fix, I prefer fixing the code instead of disabling checks for particular compilers. I'll have a look at it once I have time or you can suggest some code fixes for this.

m-chaturvedi:
Sure, thanks!
I'm not sure if there's any fix which will not make the code complicated.

m-chaturvedi:
Did you get a chance to have a look at it?  Thanks!

erwincoumans:
No volunteer shows up in fixing this hash function, so for now just suppress the warning.
If someone has a fix, re-open and provide a pull request.

PR #1230, opened 2017-07-14
Ignore .pyc everywhere.

benelot:
You do not want them in your repository. If you like this, I can add other python specific ignores.


erwincoumans:
It was already applied in another PR, thanks for the help!

PR #1235, opened 2017-07-18
Add init function and module for pybullet with EGL.

YunfeiBai:


erwincoumans:
Thanks for the PR. How can we enable EGL? Perhaps we need to dynamically load the EGL libraries (like we do for X11, GLX and OpenGL) and then we can add a new 'pybullet.connect(GUI_EGL)?

PR #1240, opened 2017-07-21
Provide sane default GwenUtil_* implementation

AMDmi3:
The linux/openbsd preprocessor chunk also applies to FreeBSD and probably much more systems, so instead of adding another condition lets's use it as default. Also merge with Apple chunk to reduce duplication.


AMDmi3:
Btw, where is the upstream of this Gwen library? I though these changes should be submitted there too.

PR #1247, opened 2017-07-30
Miscellaneous papercuts.

MakoEnergy:
I have recently updated the version of Bullet used in my engine (I was waiting on an updated Multi-threaded solution from LunkHound).  From that upgrade I found a few places where using Bullet can be made slightly easier.  Nothing ground breaking here.  

A couple extra changes seem to have made there way into the commits as well due to encoding.


erwincoumans:
Thanks for the contribution. Do you mind clearing up the conflict in Extras/CMakeLists.txt?
Bullet3Common is now always compiled, not behind the BUILD_BULLET3 anymore, so the Extras/CMakeLists.txt doesn't need to be changed (I think).

PR #1248, opened 2017-08-04
Remove hard-coded macro _BT_USE_CENTER_LIMIT_

mbeards:


erwincoumans:
It looks like your patch removes the ability to easily disable center limits, right? There is no change in functionality.

PR #1258, opened 2017-08-19
createPoseMarker example draws a 3 axis triad for visualization

ahundt:
I've implemented a basic version `createPoseMarker()` using what is available now in pure python which is for #1136. 

@erwincoumans Could this pure python function be integrated with pybullet or be reimplemented in C++? Hopefully it would be easy enough to separate and add to setup.py. That way it could be utilized like all other pybullet functions. If not, it might simply serve as useful boilerplate code that can stay in the examples.


erwincoumans:
See also pybullet/examples/debugDrawItems.py, it draws some reference frame in local link coordinates too, did you check it?

ahundt:
good point, I forgot about that one. I incorporated into this, the rename is because it includes the key words people are likely to search for the utility.

ahundt:
@erwincoumans any chance this can be merged or should I close the PR?

erwincoumans:
Probably best to close the PR. I plan on removing/cleaning up some of those pybullet examples and create a few more powerful ones. Thanks for your effort anyway!

PR #1260, opened 2017-08-19
Update minitaurGymEnv.py

hardmaru:
default code failed:

```
import minitaur_new
ModuleNotFoundError: No module named 'minitaur_new'
```

small fix to import library in the path, and to create the Minitaur object in the env.


erwincoumans:
I think this PR can be closed, since things moved around. Is that OK? Thanks for the help!

PR #1272, opened 2017-08-24
OpenGL window error handling updates

jonathantompson:
changed exit(0) to exit(EXIT_FAILURE) for window functions, and changed printf to fprintf(stderr for error string printing.


erwincoumans:
Thanks for the improvement!

PR #1273, opened 2017-08-25
Add torque sensing into GENERIC_LOGGING

JulianYG:
Now logging for generic objects also records the torques. Only for non real time simulation, otherwise values are all 0.


erwincoumans:
Thanks for the contribution. I think it is useful for adding it, but I'd like to add it behind some option for backwards compatibility. We have some project that could break if we add new fields right now.

erwincoumans:
I re-implemented it, so that also motor torques are logged. Motors will work in real-time simulation too. I'll implement some PD controller that works in real-time setting at some stage.
See new pull request here: https://github.com/bulletphysics/bullet3/pull/1288
Thanks!

PR #1286, opened 2017-08-30
Optimize Dbvt trees in place

mendsley:
Instead of allocating new vectors for each partition, simply partion the
nodes in place and pass the corresponding ranges to the next phase.


erwincoumans:
Very interesting, thanks for sharing!

Do you have any benchmark or numbers how this improves the results? 

I expected the current btDbvt not to make too many allocations: the resize(0) doesn't actually deallocated memory, so once the capacity is sufficient, the push_back should not require actual memory allocations.

It will likely take some time before I can make sure this patch doesn't break any of the existing code.

erwincoumans:
Thanks for the contribution, looks good to me.

PR #1289, opened 2017-09-01
Fix one-sided soft bodies.

samhocevar:
A coding mistake in btSoftBody::addAeroForceToNode() was applying forces
in the wrong direction for one-sided soft bodies.


erwincoumans:
Looks like an obvious bug indeed, thanks for the report and fix!

PR #1290, opened 2017-09-01
Make btHashString const-correct by changing a target pointer type.

samhocevar:


erwincoumans:
Looks reasonable, thanks for the contribution!

PR #1296, opened 2017-09-05
Gymperium: More gyms, agent registration and gym training framework.

benelot:
Atlas, Pusher, Thrower, Striker are the new gyms. The train scripts use my training framework (well, just one class) and trains a DDPG agent to perform the task. In some future commits, I can include welltrained versions of the agents.


erwincoumans:
Wow, thanks for the contribution, I'll have a look at it. By the way, there is some difference in your 'Ant' environment and the Roboschool Ant, even though they use the same xml assets. It could be the timestep, substep, iterations, default contact ERP etc. I'm looking into the differences now, if you have any insight, please let us know.

benelot:
There are still some differences between several of them and I have no clue
why. The Atlas robot does not walk at all for now, and in general, the
power constants of roboschool are too low. As if there was another scaling
factor or a different kind of control (oh, right, roboschool does not have
your motor control types...). I will check it out as soon as I have some
more spare time. As usual, just tell me if you are missing stuff you would
like to have, then I will work on that. If you like what you see, just
merge it.

If I had a VR device, I would build some VR trainable robots that can throw
balls with a VR player...Unfortunately I do not.

erwincoumans:
Before moving forward, I'd like to return to the enjoy_TF_AntBulletEnv_v0_2017may.py and make that work identically to the Roboschool Ant example. I find Python class hierarchies hard to digg through, and that makes it hard to figure out where things go wrong. Also, the pybullet.getBasePositionAndOrientation is called many times each frame for the Ant, it should only be called once. For a start, we can disable the randomization both in AntBulletEnv and in Roboschool, load the same assets, make sure the actions and observations are identical etc.

benelot:
Ok, I will have a look into it as soon as I have time. I suspect the
difference to be somewhere in the depths of the roboschool cpp code.

erwincoumans:
At a first glance some differences for Ant:
pybullet numSubSteps=2 , Roboschool  numSubSteps=4
pybullet  power=10.5, Roboschool power = 2.5
pybullet self_collision = False, Roboschool self_collision = True
pybullet loadMJCF flags = flags=p.URDF_USE_SELF_COLLISION_EXCLUDE_ALL_PARENTS
Roboschool loadMJCF flags = flags=p.URDF_USE_SELF_COLLISION+p.URDF_USE_SELF_COLLISION_EXCLUDE_ALL_PARENTS
pybullet timestep = timestep=0.0165/4, Roboschool timestep = timestep=0.0165
Roboschool sets the p.setDefaultContactERP(0.9), pybullet locomotors don't

For the Humanoid, the training/randomized reset is disable/todo/commented out.

benelot:
The power increase is what I did to make it work in the beginning. Since I
did not know that you intended to make it work exactly as they work in
roboschool, I did not track down all differences. I will fix that in the
next commit. Thanks for tracking these differences down. I realized that
roboschool also sets the kd and kp explicitly when running the motors. I
will also see if there is something done which is non-default.

erwincoumans:
So let's hold off on this CL and focus on making all the existing environments work well, with the same observations and rewards as Roboschool, given the same actions. So please create a clean pull request for those fixes (not merge it with new environments and refactoring).

With my latest changes, the Ant seems to walk well with pre-trained weights.

1) allow to switch off random values for comparison with Roboschool (same for Roboschool)
2) For each environment, check that observations and rewards are the same as Roboschool
3) Make sure each environment runs well on pre-trained weights. HalfCheetah, Walker2D and all of the pendulums don't work as well as in Roboschool
4) Make sure that each environment can be trained using OpenAI Baselines PPO (Humanoid for example doesn't have random weights at the moment): we need some script to do this.

I'll help with those items too.

Thanks for the help, this is going to be super useful!

benelot:
Just as a side-note: There is no difference in timestep between Ant pybullet and Ant roboschool:
pybullet timestep = timestep=0.0165/4, Roboschool timestep = timestep=0.0165/4
In roboschool it is set in gym_forward_walker.py.

benelot:
Sorry I did not have much time recently to look into the differences between roboschool and pybullet envs. I would love to fix it as soon as I find more time. I will check again at what point the roboschool execution differs from ours. Maybe, differences in the precision of python vs. c++ make the difference and probably it could be mitigated using numpy instead ( I had the experience of very different outputs based on slightly different inputs a lot with my simulations, in another ML project, we could fix imprecisions using numpy. Let us see what I can do. Let me know if there have been more insights on your side in the meantime. Furthermore, I really hope we can merge this PR soon, there are a lot of cool things in the pipeline (new envs not known from roboschool or openAI)!

benelot:
Hello Erwin

I might find some time after New Year's Eve to look again for differences between Roboschool and PyBullet Gym. I suspect the differences are in python vs. c++ floating-point precision. I read a lot about the issue recently, so I might find a fix for it. Do you still need a fix to be similar to roboschool or do you want to keep our own code and move on? I can create a new PR for the above improvements if you like.

erwincoumans:
Let's move on: Roboschool doesn't seem an active project.
It would be more useful to become compatible with the very popular MuJoCo locomotion tasks from Gym. Ideally allow models that are trained using PyBullet (for Reacher, Hopper, Humanoid, Ant etc) work on MuJoCo and vise versa.

The current object oriented code-structure is very confusing to me, and I keep on having to search where certain data/functions are (I made attached image).

image2

So for new MuJoCo compatible-versions, I rather have environments that are simple to understand and maintain, even it that means not sharing code.

benelot:
Ok, that sounds good. The changes in this PR make some parts a bit more
clearly separated compared to the state that is currently in the repo. I
will split this PR into some more digestible other PRs for you to accept in
several steps.

I will merge some inheritance structures a bit and duplicate some methods
to make it easier to maintain (merging the base classes into the more
specialized classes will do the trick I think). Any other suggestions?

Comparing to MuJoCo will be hard for me since I do not have a MuJoCo
license. But maybe I can get a trace of some MuJoCo env outputs using a
certain well known agent so that I can start to use the MuJoCo output as a
guideline and see if I can make our outputs more compatible. Do you have
some traces of MuJoCo or any other way for me to do this? I can even look
at videos of env performance if I have the repsective trained agent.

erwincoumans:
You could get a temporary free trial license (I think it is 1 month or so).

It is better to start from scratch for MuJoCo compatibile locomotion envs, preferably 1 file per env, keeping the code much simpler to navigate and understand.

Please don't start refactoring existing Roboschool-oriented code, and work in a separate repository, not main Bullet/PyBullet: it takes too much time for me to maintain backward compatibility, make sure nothing breaks etc. Also, I cannot properly maintain and support a large collection of new environments developed by others as main part of PyBullet and Bullet. So if you want to grow a large collection, it is better to grow it in a separate repository. Once environments are working well in the separate repository and we are happy with it, we can consider moving it to PyBullet.

Hence my suggestion to fork https://github.com/erwincoumans/pybullet_robots and create a subfolder for the environments.

benelot:
I started working in my own repository again (https://github.com/benelot/pybullet-gym) and for now I think I will keep it that way. I intend to make it an independently installable package that depends on your pybullet package, that way we can keep maintenance much simpler and you do no longer have to make sure nothing breaks when you do changes. I will continue with the Roboschool oriented code since I still think it contains a lot of helpful parts, but I will refactor is strongly and clean up all the parts not required. 

Then, I still intend to make it compatible to MuJoCo envs, the roadmap we discussed here stays the same. I just found some information on a prior attempt to make an open-source physics engine replace MuJoCo, but it seems to be hard (https://github.com/DartEnv/dart-env/wiki/OpenAI-Gym-Environments).

erwincoumans:
I'm currently working on a 'UrdfEditor.py' that makes it easier to
on-the-fly create URDF files and upload from client to server.
This would also allow dynamically changing things, add/remove limbs etc, so
you can re-create your Neural Network 3D Walkers.
Generally, I'd love to see ES creating creatures, and then train those
using RL and/or ES.

Small note I just encountered:
I recommend not to do this, because we always add new fields at the end
(right before physicsClientId field):
_,_,_,_,_,_,_,_,self.lowerLimit, self.upperLimit,_,_,_ =
p.getJointInfo(self.bodies[self.bodyIndex], self.jointIndex)
do this instead:
jointInfo = p.getJointInfo(self.bodies[self.bodyIndex], self.jointIndex)
self.lowerLimit = jointInfo[8]
self.upperLimit = jointInfo[9]
we don't reshuffle fields, only add at the end.

benelot:
I am totally up for that. The Walkers are still not deterministic and I
have no idea why. I worked way too much on that. Let us just recreate them
in pybullet. The setup of training evolved creatures is something I would
love to see too.

erwincoumans:
Sure, PyBullet is deterministic. In fact, you can save the state after n steps, restart the program (or reset the simulation) and restore the state (from .bullet file or in-memory state) and still be deterministic.

How far did you get with FlagRun and FlagRunHarder envs in PyBullet? And Atlas?
https://github.com/benelot/bullet-gym/blob/master/pybulletgym/envs/gym_humanoid_flagrun.py

benelot:
Oh that is great, if pybullet now supports all I need, I will rebuild the
NNWalkers.

The new repository for the gym things is benelot/pybullet-gym to comply
with the naming of pybullet. The python package will be named pybulletgym.
The bullet-gym repository just contains my previous state and some older
experiments.

In the pybullet-gym, I am currently working on making the envs
easier/cleaner/more maintainable. The Atlas environment works, however, I
have no trained agent for it. Same for flagrun and flagrun-harder. I will
keep you posted here.

erwincoumans:
The enjoy_TF_ReacherBulletEnv_v0_2017may.py (using your gymperium branch) doesn't seem to reach to the target. Do you have any idea why?

>> pybullet-gym
Naming is always a challenge. There is pybullet module includes pybullet_envs and pybullet_data already and pybullet_envs already has many gym envs. Perhaps pybullet_roboschool is a bit more informative? Or pybullet_keras? 

Also, some way to avoid naming conflicts of the actual environments would be useful.
Here is a spreadsheet of the current envs in pybullet_envs (assuming numbering increases v0, v1, v2 etc)
https://docs.google.com/spreadsheets/d/1MhnOE1H7JMolq6M93gVAlBakGHvHGpEvxoXM3VDGhes/edit#gid=0
It would be good to clarify for each env what the status is beyond just "working" (comments are enabled)
For "working" we could use "Do we have a way to train it successfully and enjoy it with our own trained model?"
Which of the environments can you successfully train+enjoy using your own trained model/weights?
Are there instructions how to do it?

benelot:
The reacher has a weird problem. I once looked into it, but I do not
remember what the problem was. So for now it is just a preview.

I can successfully train them and enjoy them using my own trained model (at
least a while ago I could). I will try this week and post here and update
the status in the spreadsheet. The instructions are basically in the train
scripts, which uses the Trainer class to combine an agent and an evironment.

erwincoumans:
I manually updated your Roboschool ports of PyBullet, see here: https://github.com/bulletphysics/bullet3/pull/1512
The reset was causing slowdown, since it was adding objects each time, now it uses saveState/restoreState etc. Also there are various other fixes, so you may want to include all this.

benelot:
Oh, thanks for working on this. I will check out the changes.
saveState/restoreState is definitely the better option.

erwincoumans:
After days tracking down some bug in the Humanoid/Flagrun/Harder envs, it works now.
https://github.com/bulletphysics/bullet3/pull/1524 and
https://www.youtube.com/watch?v=rMSWKdTiFAo

benelot:
Wow, thanks for working on this! I did not realize there were clamps on the
humanoids. But it actually makes sense, since it is much easier for larger
robots to control the motors with a simulated maximum output.

PR #1303, opened 2017-09-10
Export targets in CMake and add include directories to target properties

rickertm:
This patch exports CMake library targets in the namespace Bullet to enable target-based linking.

For this, the include directories are set as target properties via target_include_directories() in order to be included as dependencies. This would however require a minimum CMake version of 2.8.11. With GNUInstallDirs, the platform-specific include/lib/etc. directories can also be set accordingly. Required target-specific defines can also be added via target_compile_definitions().

With a specified Bullet_DIR or CMAKE_PREFIX_PATH environment variable, this enables linking against Bullet from applications based only on targets without further specification of include directories:

set CMAKE_PREFIX_PATH=C:\path-to-bullet\lib\cmake\bullet-2.87

find_package(Bullet CONFIG)
add_executable(foo foo.cpp)
target_link_libraries(foo Bullet::LinearMath Bullet::BulletCollision Bullet::BulletDynamics)


erwincoumans:
Thanks for the contribution. I'm not familiar with cmake namespace and GNUInstallDirs.
For our current robotics and reinforcement learning, we mostly use the new pybullet Python bindings (See http://pybullet.org if robotics are of your interest).

Would this change be backward compatible?
In other words, can you still use target_link_libraries(foo LinearMath BulletCollision BulletDynamics) without the Bullet:: namespace?

rickertm:
CMake currently has two ways of finding dependencies via find_package(): the classical one uses a find script, e.g., FindBullet.cmake, that defines variables such as ${BULLET_INCLUDE_DIRS} and ${BULLET_LIBRARIES} that can be used in a project. This is either written by the developer himself or ships with CMake in the Modules directory (the current CMake version includes a FindBullet.cmake). These can then be used in target_link_libraries(foo ${BULLET_LIBRARIES}) and include_directories(${BULLET_INCLUDE_DIRS}) or target_include_directories(foo PUBLIC ${BULLET_INCLUDE_DIRS}). ${BULLET_LIBRARIES} should contain absolute paths to the file and if available debug and optimized (release) configurations. Using link_directories() is not recommended any more (https://cmake.org/cmake/help/v3.9/command/link_directories.html).

The new approach lets projects ship a config with their installation that can be used by other projects without supplying the find script. Most of them also include targets that enable relocation of installations, as the absolute paths are no longer hard-coded but are stored in the config files. The one Bullet currently provides however does not provide absolute paths for the files, uses link_directories(), and is not relocatable. I just tried and could not compile an external project using this file on Windows with Visual Studio.

If a programmer uses the old config file or the CMake find script with target_link_libraries(foo ${BULLET_LIBRARIES}) instead of target_link_libraries(foo LinearMath BulletCollision BulletDynamics), the new config file would behave identical as it now contains "Bullet::LinearMath Bullet::BulletCollision Bullet::BulletDynamics" (and the other ones specified at the end of CMakeLists.txt). The other variables are not required any more, but can still be used. I set these to absolute paths however for compatibility with other CMake find scripts.

The namespace in combination with "::" is recommended to allow CMake to identify imported targets (see https://cmake.org/cmake/help/v3.9/manual/cmake-developer.7.html#a-sample-find-module).

GNUInstallDirs takes care of platform-specific directory names according to GNU coding standards, e.g., lib could also be lib64, lib/i386-linux-gnu, or lib/x86_64-linux-gnu.

Thank you for the link to pybullet, our main focus is however on C++ at the moment.

ahundt:
I took a quick look and these overall look like very good changes that make using bullet3 via cmake and C++ much easier and verification I'd suggest incorporating the improvments. However, two questions:

1. Will the GNUInstallDirs configuration break compatibility with windows?
2. What OSes and configurations has this been tested on?

erwincoumans:
OK, let's go for it. If there are issues on any platform, we either fix them or revert this patch until we know how to sort it out.

erwincoumans:
Thanks for the contribution!

erwincoumans:
There is some error on Windows that I need to check out, hence the revert:
https://ci.appveyor.com/project/erwincoumans/bullet3/build/1.0.2134

rickertm:
@ahundt GNUInstallDirs still uses the directory names "lib" and "include" on Windows, so this should be fine. I tested this on Windows with Visual Studio 2017 Win64 static-build and Ubuntu 16.04 64-bit shared-build. According to https://cmake.org/Wiki/CMake_Version_Compatibility_Matrix/Commands the minimum version should be 2.8.11 for target_include_directories() with the currently used options.

I just tested again with Ubuntu and a static build and got an issue regarding an undefined reference to CProfileSample. I was able to fix this by always adding the library dependencies via target_link_libraries(), e.g., target_link_libraries(BulletDynamics BulletCollision LinearMath), instead of only doing this for the BUILD_SHARED_LIBS=ON case. I'll add this to the patch.

@erwincoumans I'll have a look at the error and see if I can replicate it.

rickertm:
I rebased my changes (including the one regarding target_link_libraries mentioned above) to your commit c56c5d74e7955248af9ca0a8b3ff473ba1deef70 locally and tested with Visual Studio 2017 Win64, Python 3.5, and numpy 1.13.1 with the cmake settings from build_visual_studio_vr_pybullet_double_cmake.bat. This works fine including running ctest.
Error MSB3073 in the appveyor log seems to be related to a failed copy command for openvr_api.dll?

erwincoumans:
Yes, it copies those dll's in the right location, so the VR app runs out-of-the-box. Very odd that it failed in appveyor. It doesn't fail on my machine either. We can try to apply the PR again (need to figure out how) and see what is causing this problem.

rickertm:
Very strange indeed. I can create another pull request for the patch with the two commits or I can push the one where I rebased it to your latest commit and merged them into one.

ahundt:
@erwincoumans you should be able to click revert on your revert in #1306 :-)

Dwarfius:
@erwincoumans Could this be reverted? This would help people using CMake's ExternalProject or FetchContent that want to just include the target. Currently, one needs to patch cmakelists or manually collect include headers if he grabs the project via FetchContent, because the targets don't have exported include directories.

PR #1308, opened 2017-09-11
Fix gyms to make them work just like robotschool

benelot:
Basic purpose of this PR:
- Allow to switch off random values for comparison with Roboschool (same for Roboschool)
- For each environment, check that observations and rewards are the same as Roboschool
- Make sure each environment runs well on pre-trained weights. HalfCheetah, Walker2D and all of the pendulums don't work as well as in Roboschool
- Make sure that each environment can be trained using OpenAI Baselines PPO (Humanoid for example doesn't have random weights at the moment): we need some script to do this.

Thanks for the help, this is going to be super useful!

Notes to myself: 
At a first glance some differences for Ant:
pybullet numSubSteps=2 , Roboschool numSubSteps=4
pybullet power=10.5, Roboschool power = 2.5
pybullet self_collision = False, Roboschool self_collision = True
pybullet loadMJCF flags = flags=p.URDF_USE_SELF_COLLISION_EXCLUDE_ALL_PARENTS
Roboschool loadMJCF flags = flags=p.URDF_USE_SELF_COLLISION+p.URDF_USE_SELF_COLLISION_EXCLUDE_ALL_PARENTS
pybullet timestep = timestep=0.0165/4, Roboschool timestep = timestep=0.0165
Roboschool sets the p.setDefaultContactERP(0.9), pybullet locomotors don't

For the Humanoid, the training/randomized reset is disable/todo/commented out.

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/env_bases.py
+++ b/examples/pybullet/gym/pybullet_envs/env_bases.py
@@ -44,7 +44,7 @@ class MJCFBaseBulletEnv(gym.Env):
 			else:
 				self.physicsClientId = p.connect(p.DIRECT)
 		p.configureDebugVisualizer(p.COV_ENABLE_GUI,0)
-  
+
 		if self.scene is None:
 			self.scene = self.create_single_player_scene()
 		if not self.scene.multiplayer:
--- a/examples/pybullet/gym/pybullet_envs/robot_locomotors.py
+++ b/examples/pybullet/gym/pybullet_envs/robot_locomotors.py
@@ -152,23 +152,25 @@ class Humanoid(WalkerBase):
 		self.motor_names += ["left_shoulder1", "left_shoulder2", "left_elbow"]
 		self.motor_power += [75, 75, 75]
 		self.motors = [self.jdict[n] for n in self.motor_names]
-		# if self.random_yaw: # TODO: Make leaning work as soon as the rest works
-		# 	cpose = cpp_household.Pose()
-		# 	yaw = self.np_random.uniform(low=-3.14, high=3.14)
-		# 	if self.random_lean and self.np_random.randint(2)==0:
-		# 		cpose.set_xyz(0, 0, 1.4)
-		# 		if self.np_random.randint(2)==0:
-		# 			pitch = np.pi/2
-		# 			cpose.set_xyz(0, 0, 0.45)
-		# 		else:
-		# 			pitch = np.pi*3/2
-		# 			cpose.set_xyz(0, 0, 0.25)
-		# 		roll = 0
-		# 		cpose.set_rpy(roll, pitch, yaw)
-		# 	else:
-		# 		cpose.set_xyz(0, 0, 1.4)
-		# 		cpose.set_rpy(0, 0, yaw)  # just face random direction, but stay straight otherwise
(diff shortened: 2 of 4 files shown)


benelot:
First step is to clean up the stuff that are not required/roboschool specific visualization. Second is to remove all the tweaks that I added to make it work quickly.

benelot:
It is pretty normal that some of the envs now work even worse than before. But only from that point on we can get it to work exactly the same as roboschool.

erwincoumans:
I have been working on this as well, did you do a git pull? The Ant works very well in the current trunk version.

benelot:
Yes, I did a git pull. Ant still works the same. Until now I just removed
some force gains I introduced to make it work. But those gains have nothing
to do with how roboschool does it. This pull request is not yet meant to be
merged, this is just to show to avoid duplicate work by showing you what I
am doing. Next I will look how roboschool tunes and controls the joints.
Because somehow roboschool applies higher forces in general.

benelot:
There are several weird defaults coded into the physics-bullet.cpp. It will take a while to get our code to work the same.

benelot:
Humanoid works well too, also with randomized reset. I do not have roboschool installed as a reference (It is pretty cumbersome to install.), but I will install it as soon as I get most of the envs running pretty well. Until then I will improve them just by code comparison.

erwincoumans:
thanks! did you try training the fixed envs using TF Agents? I got a fork here, adding pybullet envs to it: https://github.com/erwincoumans/agents

erwincoumans:
I'd like to get incremental improvements faster. Shall we create individual issues for each agent (Ant, Hopper, Half-Cheetah, Double Pendulum etc) and solve them as-we-go? I'm going to merge this one.

benelot:
Sounds great!Let us open issues for each env separately and resolve them
separately. I just installed roboschool, now I can compare. Have you ever
had a look at the roboschool cpp code I mentioned? Maybe you understand
better what is different there from pybullet than what I get. A lot of
b3*code there I don't exactly know what it does.

erwincoumans:
The Roboschool cpp code is calling exactly the same C-API as pybullet does. There isn't happening that much different, is it?

I created a logger and replay for those commands, so it is easier to compare. I still need to convert them to ascii files, to make it easier to spot differences.

erwincoumans:
One way of comparison is: disable random in the reset. Then apply some standard actions in both Roboschool and pybullet env, and expect the same observation and reward.

benelot:
Ok, so from staring at the code and the actions and observations for 1.5 hours, I could find out that roboschool definitely causes a stronger push when applying the same amount of force. However, I could not find out why this happens. The actions and observations in inversePendulum quickly diverge since roboschool manages to balance the pendulum and pybullet does not manage to do so and continuously applies a similar force to make it happen. 
Example: 
Starting from Pendulum upright ( the first two values are x and vx):

Roboschool:
observation:  [ 0.  0.  1.  0.  0.]
action: 25.51864111260001
observation:  [  5.62117668e-04   3.40677351e-02   9.99999317e-01  -1.16903390e-03
  -7.08505586e-02]

pybullet:
observation: [ 0.,  0.,  1.,  0.,  0.]
action: 25.51864111260001
observation: [  0.00000000e+00,   0.00000000e+00,   1.00000000e+00, 9.32163201e-06,   5.64947395e-04]

What could be the cause of that behavior?

PR #1309, opened 2017-09-11
Conflict with x11 cpp 3 building

AndreaCatania:
The compiler x11 insert a header that define the macro "None". Is better to use the word Nothing instead.

### DIFF
--- a/src/BulletSoftBody/btSoftBody.cpp
+++ b/src/BulletSoftBody/btSoftBody.cpp
@@ -503,7 +503,7 @@ void			btSoftBody::addAeroForceToNode(const btVector3& windVelocity,int nodeInde
 					fDrag = 0.5f * kDG * medium.m_density * rel_v2 * tri_area * n_dot_v * (-rel_v_nrm);
 							
 					// Check angle of attack
-					// cos(10บ) = 0.98480
+					// cos(10ยบ) = 0.98480
 					if ( 0 < n_dot_v && n_dot_v < 0.98480f)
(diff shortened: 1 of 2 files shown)


erwincoumans:
Just #undef None before including the Bullet header, or move the Bullet header first.

AndreaCatania:
The problem is that "None" is used in the project code too, so after Bullet inclusion I've to redefine it. Since its value is defined outside the project can change and I'm not sure about its value. For this reason I want to avoid "undef"

AndresTraks:
Try refactoring the code in your project into separate files and include X11 and btSoftBody.h only in the files where you need them. You probably only need both definitions in some small part of the code, if at all. Then do the #undef hack where necessary.

The problem is with X11, because it's C that doesn't allow sane design or proper encapsulation of constants, not C++. It's expected that you need some hacks to interoperate between the two, but it's manageable if you keep the conflicting parts separate from the rest of the code.

erwincoumans:
Move the Bullet header before the X11 header and it should be fine. Can you point to this X11 header that defines the None macro?

AndreaCatania:
@AndresTraks The undef define is what I wanted to avoid, but it is the way how I did it.
@erwincoumans I'm implementing Bullet wrapper into Godot engine and I can't  do it. I've resolved the problem using undefine include define even if I don't like it.

It took me a lot of time to understand that "None" is causing this problem, for this reason I think change this name is a good thing. Even if it force to update the Bullet wrappers code that use this enumaration.

erwincoumans:
The None problem is caused by X11, not Bullet.

You could have solved it by #including Bullet headers BEFORE X11 . Why do you ignore this suggestion???

AndreaCatania:
In this moment I can't change the structure of inclusions to respect always the clang rules. I'll check if I can implement it in another way.

PR #1318, opened 2017-09-15
fix pybullet pendulums/pendula gym environments

erwincoumans:
some joins were not unlocked (main reason), removed extra gains that were applied

This partly fixes stuff for https://github.com/bulletphysics/bullet3/pull/1308

See 'set_servo_target' call, that unlocks the motor, inside the first_torque_call condition:

https://github.com/openai/roboschool/blob/master/roboschool/cpp-household/physics-bullet.cpp#L502

if (first_torque_call) {
		set_servo_target(0, 0.1, 0.1, 0);
		first_torque_call = false;
	}

### DIFF
--- a/Extras/Serialize/BulletFileLoader/btBulletFile.cpp
+++ b/Extras/Serialize/BulletFileLoader/btBulletFile.cpp
@@ -128,6 +128,9 @@ void btBulletFile::parseData()
 	mDataStart = 12;
 	remain-=12;
 	
+	//invalid/empty file?
+	if (remain < sizeof(bChunkInd))
+		return;
 
 	char *dataPtr = mFileBuffer+mDataStart;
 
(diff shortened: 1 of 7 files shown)


benelot:
Good observation! So I have to disable position control first to unlock the joint in order to torque control it. So I can torque control a joint like this:
p.setJointMotorControl2(bodyIndex=self.bodies[self.bodyIndex], jointIndex=self.jointIndex, controlMode=p.TORQUE_CONTROL, force=torque)

and still apply a position control the same joint like this at the same time?
p.setJointMotorControl2(self.bodies[self.bodyIndex],self.jointIndex,controlMode=p.POSITION_CONTROL, targetPosition=0, targetVelocity=0, positionGain=0.1, velocityGain=0.1, force=100)

If so, I think that should be mentioned in the docs. I know that this could be useful to simulate motor noise, but then it should be clearly stated. Because for now the API gives the impression that you **set** the joint motor control to a certain control mode. Modes can usually not be enabled at the same time.

erwincoumans:
>>If so, I think that should be mentioned in the docs. 

Yes, setJointMotorControl2 for position and velocity control will configure the existing motor/constraint.

When using setJointMotorControl2 for torque control, it directly applies a torque.

The idea is that the constraint can be used to mimic joint friction (set target to zero, friction = max force, when torque control is used.

PR #1320, opened 2017-09-16
[WIP] Fix hopper environment.

benelot:
[This is work in progress] I am working on the hopper environment. 2017may version work equally well in both roboschool and pybullet. However the 2017jul versions do not show the same performance. I am trying to find out why. I added a commit that applies your fix for the pendula.


erwincoumans:
Yes, it would be great if you can help tracking down (bisection?) what commit breaks the pybullet cheetah and/or hopper.

Not sure what changes you are still making to pendula code? And motors are already disabled right after XML loading, so no need to do it again later.

benelot:
I disable it later since I found that it makes the code easier to read as
there is a whole section of joint configurations. Also I removed the motor
disabling for the "ignore" joints since we disable motors for all by
default. I can drop my commit and replace with your as soon as you merge.
Basically I just needed your changes.

Cheetah and hopper worked with the other timestep I think. I will verify
that of course. But if you look at roboschool, at least the hopper does not
really work well with the 2017may version either. It just jumps for a while
and then falls. 2017jul version work much better in roboschool, but does
not in pybullet. That is how far I am.

erwincoumans:
It would be easier to limit the actual changes needed to the minimum, perhaps start a new PR: no new trained network data, checking in debug lines etc. That makes it a bit easier to digg through the essential changes to make pybullet envs similar to Roboschool (and the easier MuJoCo envs in gym soon).

1) the testEnv.py util in 'examples' can be used to compare observations and rewards for between environments

Some very minor changes could be due to compiler settings and use of python versus C++ to issue Bullet commands (pybullet versus Roboschool cpp-household). Still, the values should be similar enough within 1 step. Eventually this will drift indeed.

2) Roboschool RoboschoolHalfCheetah_v0_2017may.py works well with Oleg's fork of Bullet, with some modifications, such as a hard-coded restitution coefficient of 0.5 always.

We could move from that Fork, to the regular Bullet version of that data, apply the few needed changes (restitution 0.5 etc, but not the self-collision patch) and see if cheetah still works OK. Then move to newer versions of Bullet, and see what commit breaks compatibility.

benelot:
From what I saw from my debug outputs is indeed that initially the numbers are very similar then they drift. Yeah, I feel that my current approach does not help. Thanks for mentioning the test script, that might be helpful. On the note of the half-cheetah roboschool script, that is definitely another change we need to follow. We can close the PR, the only significant commit here is: Update stuck joints penalty and revert some inconclusive changes in WalkerBaseBulletEnv._step.
https://github.com/bulletphysics/bullet3/commit/9ab0c5b3beed7e0fe1c7c9cbd237170567ca5ff9

There is a penalty that needs updating and a change you made that I do not really understand why you did it. I think we should keep the code as close to the original roboschool code and only after that make changes/optimizations.

benelot:
I will do next what you proposed in 2. First I will look if it still works without the self-collision patch. Then I will check if it breaks with any additional commits.

PR #1331, opened 2017-09-22
Cmake pybullet fix

### LINKED ISSUE #1322: Could NOT find PythonLibs (missing:  PYTHON_LIBRARIES) (found suitable exact version "2.7.12")
Could NOT find PythonLibs (missing:  PYTHON_LIBRARIES) (found suitable exact version "2.7.12")

ahundt:
fix #1322, the problem is that python itself declares its own lib file to be in the wrong location in certain linux versions. Here I primarily add the extra missing path component to the set of search paths. I also restore changes from previous fix attempts that were incorrect.

https://github.com/bulletphysics/bullet3/pull/1131 is a past change of interest to this one

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,8 +301,25 @@ IF (APPLE)
 ENDIF()
 
 OPTION(BUILD_BULLET3 "Set when you want to build Bullet 3" ON)
-
-OPTION(BUILD_PYBULLET "Set when you want to build pybullet (Python bindings for Bullet)" OFF)
+# Optional Python configuration
+# builds pybullet automatically if all the requirements are met
+SET(PYTHON_VERSION_PYBULLET "2.7" CACHE STRING "Python version pybullet will use.")
+SET(Python_ADDITIONAL_VERSIONS 2.7 2.7.3 2.7.12 3 3.0 3.1 3.2 3.3 3.4 3.5 3.6)
+SET_PROPERTY(CACHE PYTHON_VERSION_PYBULLET PROPERTY STRINGS ${Python_ADDITIONAL_VERSIONS})
+SET(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/build3/cmake ${CMAKE_MODULE_PATH})
+OPTION(EXACT_PYTHON_VERSION "Require Python and match PYTHON_VERSION_PYBULLET exactly, e.g. 2.7.12" OFF)
+IF(EXACT_PYTHON_VERSION)
+  set(EXACT_PYTHON_VERSION_FLAG EXACT REQUIRED)
+ENDIF(EXACT_PYTHON_VERSION)
+# first find the python interpreter
+FIND_PACKAGE(PythonInterp ${PYTHON_VERSION_PYBULLET} ${EXACT_PYTHON_VERSION_FLAG})
+# python library should exactly match that of the interpreter
+FIND_PACKAGE(PythonLibs ${PYTHON_VERSION_STRING} EXACT)
+SET(DEFAULT_BUILD_PYBULLET OFF)
+IF(PYTHONLIBS_FOUND)
+	SET(DEFAULT_BUILD_PYBULLET ON)
+ENDIF(PYTHONLIBS_FOUND)
+OPTION(BUILD_PYBULLET "Set when you want to build pybullet (Python bindings for Bullet)" ${DEFAULT_BUILD_PYBULLET})
 
 OPTION(BUILD_ENET "Set when you want to build apps with enet UDP networking support" ON)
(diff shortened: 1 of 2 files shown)


erwincoumans:
Thanks for looking into this and the fix. Since it is all green, let's apply it!

erwincoumans:
The fix is not OK, it disables Python/pybullet again: See travis:
-- Could NOT find PythonLibs (missing: PYTHON_LIBRARIES) (found suitable exact version "2.7.13")
I'll have to revert this again.

ahundt:
I don't know how the configuration differs... I ran this on my ubuntu 16.04 machine and it worked ok, finding python, the libs, and numpy all correctly.

ahundt:
@erwincoumans https://travis-ci.org/bulletphysics/bullet3/jobs/278469626 worked 

```

OPENGL FOUND
/usr/lib/x86_64-linux-gnu/libGLU.so/usr/lib/x86_64-linux-gnu/libGL.so
-- Looking for versions: 2.7.6;2.7
-- Looking for python version '2.7.6' by checking executables: python;python2;python2.7.
-- Found executable /opt/pyenv/shims/python with suitable version 2.7.6
-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython2.7.so (found suitable exact version "2.7.6") 
-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython2.7.so (found version "2.7.6") 
-- Configuring done
-- Generating done
-- Build files have been written to: /home/travis/build/bulletphysics/bullet3
```

OH, I think you have to update travis for the other python versions so it finds the python version each travis build is supposed to test correctly. I don't think that's the fault of this cmake change, that's a travis change/cmake change. I set the default python to 2.7, so if you want to use some other version you need to specify it in ccmake!

ahundt:
specifically change https://github.com/bulletphysics/bullet3/blob/master/.travis.yml#L16 to `cmake .  . -DBUILD_PYBULLET=ON -G"Unix Makefiles" -DPYTHON_VERSION_PYBULLET=3.5` or whatever you want...

erwincoumans:
Why do we need a default python version? I'd like cmake to find 'some' python version, not the one I'm asking for. Perhaps some 'preferred' python version would be OK: if it finds multiple versions, pick the default (if it is among the versions found). Otherwise, cmake should just pick one.

ahundt:
It has a default, and the default is 2.7. Since that comes with ubuntu if I remember correctly it is always found. I don't know what travis run you are specifically referring to which failed so I cannot help without it.

erwincoumans:
FIND_PACKAGE(PythonLibs) works like that: you just find *SOME* python version. 
I don't want to be forced to specify what specific version cmake has to find, and I certainly don't want to be forced to tell where the include and library files are located: that is the whole point of FIND_PACKAGE to figure that out in a cross-platform way.

erwincoumans:
Today, it finds Python version "3.4.3" on linux and  "2.7.10" on OSX: https://travis-ci.org/bulletphysics/bullet3/jobs/279592908

ahundt:
Is https://travis-ci.org/bulletphysics/bullet3/jobs/279592908 with this patch or no? 

I think I'm seeing a chicken and egg problem here:
1. if it just finds some python version, and you're running 3.5, you could get 2.7 and 3.5 wouldn't work. 
2. If you don't make sure all the python versions are consistent, you could get 2.7 for the libs and 3.5 for the interpreter and then the install/execution breaks. 
3. When `n` python versions are installed with a correct auto-detect implementation, not specifying the python version will fail `n-1/n` of the time, but specifying the python version you want in advance will work 100% of the time.

Auto-detect across python 2 and 3 previously appeared to work in some cases but contained a bug, namely item 2 above. 

This patch resolves that by specifying the version in advance and choosing a reasonable default. Reverting this patch leaves it in the broken state 2. As I understand things, this PR works correctly to resolve that problem 2, as designed.

Is that description all consistent with your understanding of things? What would you find satisfactory?

erwincoumans:
>> Is https://travis-ci.org/bulletphysics/bullet3/jobs/279592908 with this patch or no?

No, it uses FIND_PACKAGE(PythonLibs) as is the current github bullet3 master.

I'd prefer to use cmake's built-in FIND_PACKAGE(PythonLibs) by default,
with an option called USE_IMPROVED_FIND_PYTHON to use the improved version (turned off by default).

Why doesn't this improved way of finding python (that you suggest) makes its way into cmake, replacing default FIND_PACKAGE(PythonLibs)?

ahundt:
Ok I think it is working in a way you will like with https://github.com/bulletphysics/bullet3/pull/1340

I set it to find any python, and I prioritized 3 over 2, so it will find 3.x first, then if it doesn't find that it will find 2.x. Additionally, it is easy to set the exact version you'd like to use as well.

Hopefully that is satisfactory?

PR #1333, opened 2017-09-24
[pybullet] Add calculateJacobian.

bingjeff:
* Add the calculateJacobian method to the pybullet API.
* Adjust the shared memory interface to handle fixed/floating bases
  in the calculateJacobian method.
* Fix a few comments.

### DIFF
--- a/examples/pybullet/pybullet.c
+++ b/examples/pybullet/pybullet.c
@@ -445,10 +445,10 @@ static PyObject* pybullet_connectPhysicsServer(PyObject* self, PyObject* args, P
 			{
 				printf("Connection terminated, couldn't get body info\n");
 				b3DisconnectSharedMemory(sm);
-                		sm = 0;
+						sm = 0;
 				sPhysicsClients1[freeIndex] = 0;
-                       		sPhysicsClientsGUI[freeIndex] = 0;
-                        	sNumPhysicsClients++;
+							sPhysicsClientsGUI[freeIndex] = 0;
+							sNumPhysicsClients++;
 				return  PyInt_FromLong(-1);
 			}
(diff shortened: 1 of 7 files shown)


examples/pybullet/pybullet.c
```diff
@@ -6947,6 +6953,147 @@ static PyObject* pybullet_calculateInverseDynamics(PyObject* self,
 	return Py_None;
 }
 
+/// Given an object id, joint positions, joint velocities and joint
+/// accelerations,
```
erwincoumans: I guess the comment was copied from the Inverse Dynamics API?

examples/pybullet/pybullet.c
```diff
@@ -6947,6 +6953,147 @@ static PyObject* pybullet_calculateInverseDynamics(PyObject* self,
...
+							double* linearJacobian = (double*)malloc(3 * byteSizeDofCount);
+							double* angularJacobian = (double*)malloc(3 * byteSizeDofCount);
+							b3GetStatusJacobian(statusHandle,
+												NULL,
+												linearJacobian,
+												angularJacobian);
+							if (linearJacobian)
+							{
+								PyObject* pymat = PyTuple_New(3);
+								for (int r = 0; r < 3; ++r) {
```
erwincoumans: you can move the 'int r' out of the loop, to satisfy CI?

AndresTraks: GCC 5 enables C99 features by default. Is it possible to upgrade to at least that instead?

erwincoumans: This is Visual Studio compiling C code, simple fix.

PR #1336, opened 2017-09-25
[sharedmemory] Fill-out calculateJacobian command.

bingjeff:
The server command processor actually didn't do anything with
the local point that was passed along with the calculateJacobian
command. Added in the necessary bit of math to return the
corresponding jacobian.

Also, fixed a typo in pybullet that was returning the same
jacobian for translation and rotation.


erwincoumans:
Thanks again. For most of the new functionality, I add some little example that shows how to use the command, in the Bullet/examples/pybullet/examples folder. If you have time, do you have some jacobian.py example?

bingjeff:
Perfect! Will do, that will also give me an excuse to test the code thoroughly.

PR #1344, opened 2017-09-27
Fix some IK related issues.

YunfeiBai:


erwincoumans:
Thanks for the help with the IK!

PR #1345, opened 2017-09-28
[pybullet] Fix bugs in calculateJacobian.

bingjeff:
* Added a small example for using calculateJacobian.
* Fixed translation jacobian to correctly use the local point.


erwincoumans:
Great, it looks like the numbers are matching. Thanks for the contribution!

PR #1349, opened 2017-09-29
Fix a shadow warning by renaming a parameter in btInfMaskConverter

piluke:
This warning has been annoying me for a couple months so I finally decided to send a pull request. The warning can be seen when compiling with `-Wshadow` in GCC.

I'm not very experienced with contributing to open source so please let me know if I need to change anything.


erwincoumans:
Thanks for the contribution!

PR #1350, opened 2017-10-01
[pybullet] Add calculateMassMatrix.

bingjeff:
* Fix compile warnings for GwenOpenGLTest and Test_enet_chat_client.
* Fix small typo in b3PluginArguments.
* Add calculateMassMatrix to the shared memory interface and pybullet.


erwincoumans:
Thanks again for the contribution, Jeff. Do you mind creating a small example .py file? Also, a small note why/how one would use the mass matrix (and jacobian) would be useful.

Some minor note: the CalculateMassMatrixResultArgs structure is really big, since it is double m_massMatrix[MAX_DEGREE_OF_FREEDOM * MAX_DEGREE_OF_FREEDOM];
For other commands using a lot of data, there is some 'streaming' happening, but I don't expect you to digg that deep. I can fix that in a follow-up commit. In a nutshell: the client (PhysicsDirect for example) sends multiple commands, and caches the result locally.

Also, it is best not to insert enums in the middle of avoid messing up compabitility between client and servers. If the enums (or command layout) changes, the SHARED_MEMORY_MAGIC_NUMBER in SharedMemoryPublic.h needs to be updated (year/month/day/rev).

And thanks for fixing the int->float copy/paste bug in b3PluginArguments!

Soon I'm going to clean up that gigantic switch statement in PhysicsServerCommandProcessor into 'plugins' with some command registry, similar to the b3PluginManager.

bingjeff:
Sure! I will make an example this week.

Would you like me to move the enums?

I was a little worried about the size of the mass matrix allocation.

Another thing that seems wasteful is the number of times that the inverse
dynamics routine might be being called from python. I rationalized this
that python should really only be used for prototypes, but if you think
there is a better way I would be happy to help fix it.

erwincoumans:
How/why do you want to use inverse dynamics and the mass matrix?

We use pybullet/Python for RL training for locomotion, grasping etc, and for this we don't need the mass matrix or inverse dynamics.

bingjeff:
I'm still doing things the old fashioned way ;)

I have a home project where I am doing task space control. Mass matrix is
useful for dof decoupling. Jacobians are useful for mapping task to joint.

ahundt:
@erwincoumans There is every reason to train an RL task on higher level aspects of a problem like grasping while low level control is delegated to well established traditional task space control methods.

@bingjeff thanks for working on this!

erwincoumans:
I'll merge it and make the matrix streaming in a separate commit. Thanks for the contribution!

erwincoumans:
If/when you create the mass matrix example, please update to latest to confirm that my changes (to move mass matrix in streaming part instead of status part of the shared memory) are OK.
See also https://github.com/bulletphysics/bullet3/pull/1358

Issue #1352, opened 2017-10-02
doubt regarding btCompund shape


srajangarg:
I modified the `SimpleBox.cpp` example to test out compound shapes.

```
{
	btCompoundShape* compoundShape = new btCompoundShape();
	btBoxShape* cube  = new btBoxShape(btVector3(0.5, 0.5, 0.5));
	m_collisionShapes.push_back(cube);
	btTransform transform;

	transform.setIdentity();
	transform.setOrigin(btVector3(0, 0, 0));
	compoundShape->addChildShape(transform, cube);

	transform.setIdentity();
	transform.setOrigin(btVector3(0, 1, 0));
	compoundShape->addChildShape(transform, cube);

	transform.setIdentity();
	transform.setOrigin(btVector3(0, 2, 0));
	compoundShape->addChildShape(transform, cube);

	transform.setIdentity();
	transform.setOrigin(btVector3(0, 1, 1));
	compoundShape->addChildShape(transform, cube);

	m_collisionShapes.push_back(compoundShape);
	transform.setIdentity();
	transform.setOrigin(btVector3(0, 10, 0));
	createRigidBody(1.0, transform, compoundShape);
}
```
^ is what i added, apart from the existing ground shape.

It render properly, but the physics are off 
![](https://i.imgur.com/zelXziC.png)

I cannot have it rest on diagonally on the ground. If I push it (using my mouse) so as to make it fall ttowards the protruding cube's side, it behaves as if it is hinged and rises pack to this upright position. Can someone point out what I am doing wrong?

erwincoumans:
The btRigidBody is aligned with the center of mass. You need to recompute the shift to make sure the compound shape is re-aligned. Here is a snippet, we should have this as a basic example indeed.


`
	{
		btCompoundShape* compoundShape = new btCompoundShape();
		btBoxShape* cube  = new btBoxShape(btVector3(0.5, 0.5, 0.5));
		m_collisionShapes.push_back(cube);
		btTransform transform;
	
		transform.setIdentity();
		transform.setOrigin(btVector3(0, 0, 0));
		compoundShape->addChildShape(transform, cube);

		transform.setIdentity();
		transform.setOrigin(btVector3(0, 1, 0));
		compoundShape->addChildShape(transform, cube);

		transform.setIdentity();
		transform.setOrigin(btVector3(0, 2, 0));
		compoundShape->addChildShape(transform, cube);

		transform.setIdentity();
[...]

PR #1353, opened 2017-10-02
add new example for btCompoundShape

srajangarg:
https://github.com/bulletphysics/bullet3/issues/1352

### DIFF
--- a/examples/ExampleBrowser/CMakeLists.txt
+++ b/examples/ExampleBrowser/CMakeLists.txt
@@ -122,6 +122,8 @@ SET(ExtendedTutorialsSources
 	../ExtendedTutorials/SimpleBox.h
 	../ExtendedTutorials/MultipleBoxes.cpp
 	../ExtendedTutorials/MultipleBoxes.h
+	../ExtendedTutorials/CompoundBoxes.cpp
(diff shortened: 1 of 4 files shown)


srajangarg:
ping @erwincoumans

erwincoumans:
Thanks! Could you also add the file to the premake build file (at least Bullet/examples/ExampleBrowser/premake4.lua)?

srajangarg:
I checked the premake build files, none of them have `SimpleBox.cpp` / `SimpleBox.h` in them. If premake builds SimpleBox, then CompoundBoxes will also be built. (via maybe the *.cpp / *.h directive) However, it might be the case none of them are being built, haven't tested out the premake build system.

Re : CI, maybe you can just add additional travis jobs which build using premake?

erwincoumans:
Ah, that's right, premake likely uses a wildcard *.cpp. 

Do you have time to add an additional travis job?
On Linux the procedure to build would be:
cd build3
./premake4_linux64 gmake
cd gmake
make -j12

On OSX , replace the command with ./premake4_osx

srajangarg:
I'll try and look at premake builds later. Mayeb after I add the VHACD convex decomposition example.

PR #1359, opened 2017-10-06
Pull request

jietan:
Add the new urdf for derpy. Created by Tingnan.


erwincoumans:
Nice one, thanks for the improved URDF file!

PR #1365, opened 2017-10-07
Fixed btVector3.h maxAxis4

elect86:


erwincoumans:
No, we shouldn't assign the value at the end, only the index is returned.

elect86:
and is it right a little after? At line 1191? 

https://github.com/bulletphysics/bullet3/pull/1365/files#diff-0e7e8456d3cb1704176af721986d7786R1191

PR #1371, opened 2017-10-09
Fixed setting of m_userConstraintResultArgs.

jonathantompson:
I'm not sure how it ever worked for anyone. It was filling up the userConstraintData with stale memory because serverCmd.m_userConstraintResultArgs wasn't ever initialized.

Please check the change though, because I'm not entirely familiar with the PhysicsServerCommandProcessor.cpp codebase (8589 lines of code! It's a monster ;-) ).

### DIFF
--- a/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
+++ b/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
@@ -7209,6 +7209,7 @@ bool PhysicsServerCommandProcessor::processCommand(const struct SharedMemoryComm
 												InteralUserConstraintData userConstraintData;
 												userConstraintData.m_mbConstraint = multibodyGear;
 												int uid = m_data->m_userConstraintUIDGenerator++;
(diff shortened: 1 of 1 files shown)


erwincoumans:
Well spotted, thanks for the fix.  It would corrupt the getConstraintInfo for gear and fixed user generated constraints indeed, so let's fix it.

It likely doesn't affect many, since most only load URDF/SDF/MJCF files instead of programmatic generation. In addition, the important return value for createConstraint is the integer constraint unique Id, through b3GetStatusUserConstraintUniqueId, which is valid.

>> 8589 lines of code! It's a monster

Yes, the monster will be smashed sooner or later, the enormous switch statement, will be refactored/moved into separate files, possibly some in plugins (such as logging and robot-specific functionality).

jonathantompson:
Yeah, I started looking around for uses of those constraints, and the vast majority of the time people are sending FIXED constraints first using the same parent and child objects. One my machine this fills the struct with mostly valid data from the previous call (which is why this bug was hard to find)...  So it could be that users were getting lucky.

Issue #1249, opened 2017-08-04
BulletIK


JulianYG:
**Problem 1: Parsing URDF**
It seems currently the Bullet IK can only parse arm urdf’s; that is, with the full robot model, it generates wrong IK solutions. The complete Sawyer robot model for example, has 20 joints, where 7 are activated joints and 13 are fixed joints. Bullet cannot construct a correct arm model from the full urdf, and the IK solutions are incorrect.

**Problem 2: Inverse Kinematics**
The robot with different joint states have different solutions for the same target pose. That is, the IK solutions are not state independent. 

It also needs multiple iterations to solve for one accurate solution. In bullet3/examples/pybullet/examples/inverse_kinematics.py, change use_real_time = 1 to 0, then the robot cannot reach to desired position at all.

The null space solution is very off. In bullet3/examples/pybullet/examples/inverse_kinematics.py, change use_null_space = 0 to 1, the actual and target lines are separated far away.

**Problem3: Rendering**
Depending on types of different hardware, the rendered color of the grippers are different. For example, a PR2 gripper image has red and green fingers, which is inconsistent with original yellow and green colors. This detriments the learning models based on visual inputs.

Pictures are attached in:
https://docs.google.com/document/d/1XH7GGaxPjahu-3z51t7aSw39DWgmjQ-NeFSvvyZYTdU/edit?usp=sharing

ahundt:
My project group has also noticed some problems with the IK. If you're willing to use external IK and only use bullet for the physics might I suggest https://github.com/jrl-umi3218/Tasks?

erwincoumans:
ahundt, JulianYG, please provide some URDF files that have issues with IK. That way we can resolve it, and people don't need to revert to an external IK solution.

Where can we find the Sawyer robot?

>>a PR2 gripper image has red and green fingers

The colors are not assigned in the URDF or SDF file, this can be easily fixed by the user.

Thanks for the feedback, this helps us improve pybullet etc.

JulianYG:
Yeah, I see some urdf files don't define colors in them, and they are related to the hardware that renders them. The sawyer urdf is here:
https://www.dropbox.com/s/aoxg0z0vheu02mp/sawyer_robot.zip?dl=0

Thanks to @ahundt for the link.

erwincoumans:
[...]

PR #1376, opened 2017-10-10
Modify inverse kinematics to also support robot model with fixed joint.

YunfeiBai:
#1249 Modify inverse kinematics to also support robot model with fixed joint. The attachment has the pybullet example code for sawyer arm IK, and its demo video.
[inverse_kinematics_sawyer.zip](https://github.com/bulletphysics/bullet3/files/1370273/inverse_kinematics_sawyer.zip)


erwincoumans:
Thanks for tracking this down and fixing it, Yunfei!

YunfeiBai:
Here is the updated pybullet example using qIndex from getJointInfo for mapping between joint index and DoF index.
[inverse_kinematics_sawyer.zip](https://github.com/bulletphysics/bullet3/files/1370394/inverse_kinematics_sawyer.zip)

PR #1382, opened 2017-10-14
Windows case sentitive fs build fix

jschueller:


erwincoumans:
Thanks for the fix/improvement!

PR #1383, opened 2017-10-14
Customize restitution and friction

AndreaCatania:
I've implemented bullet physics wrapper for Godot engine: https://github.com/godotengine/godot/pull/10013 and now I need to customize how is calculated combined the restitution and the friction.

To do this I saw that bullet is not at 100% ready, but there's this code: https://github.com/bulletphysics/bullet3/blob/4c2c2bf74b65038d3184a27888f013bb88dbf92e/src/BulletCollision/CollisionDispatch/btManifoldResult.h#L149-L155

So I've implemented it.

### DIFF
--- a/src/BulletCollision/CollisionDispatch/btManifoldResult.cpp
+++ b/src/BulletCollision/CollisionDispatch/btManifoldResult.cpp
@@ -22,7 +22,12 @@ subject to the following restrictions:
 ///This is to allow MaterialCombiner/Custom Friction/Restitution values
 ContactAddedCallback		gContactAddedCallback=0;
 
-
+CalculateCombinedCallback		gCalculateCombinedRestitutionCallback = &btManifoldResult::calculateCombinedRestitution;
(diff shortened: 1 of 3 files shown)


erwincoumans:
Thanks for the report. I'd like to avoid API changes and make this a small change. You should be able to modify/compute all the data by setting the gContactAddedCallback to your own. Did you try that?

AndreaCatania:
Thanks you for your response and disponibility. I thought that you didn't want to change it, I've edited my commit now it's using a callback for each function. In This way we should be able to customize it.

I can't use the callback gContactAddedCallback because it allow me to edit restitution of friction on the fly but not the algorithm how it's calculated. So to acchieve what I need I should create "an hack" that simulate the type of combinatin I need, and if it's possible I want avoid the adding of useless code or hack.

AndreaCatania:
Hi @erwincoumans, let me know if I need to change something or if it's fine because I need to implement it. Thanks you!

erwincoumans:
This looks good to me, thanks for the contribution!

Issue #1366, opened 2017-10-07
Backport 4ee09c2 to older bullet versions


ShFil119:
Hi, can you backport this fix (4ee09c2) to older version of bullet? With using z as up axis at least 2.86, 2.85 are corrupted. Thanks. ;)
https://github.com/rwengine/openrw/issues/318

erwincoumans:
Can't you just upgrade to 2.87?

PR #1388, opened 2017-10-18
Fix jump(without argument)

ShFil119:
While searching the reason of not working jump(when using Z as vertical axis). Lihis found that jump requires btVector3 filled with zeros. But actually default argument is empty.
Fix for openrw:
https://github.com/rwengine/openrw/commit/4bf1d3795b226c46ffd9c6873669b547047e453d
Issue to close:
https://github.com/bulletphysics/bullet3/issues/1366


erwincoumans:
Thanks for the contribution, let's merge it!

PR #1394, opened 2017-10-19
Optimize rayTest with btScaledBvhTriangleMeshShape

davilovick:
I've modified **btCollisionWorld::rayTestSingleInternal(...)** method to improve the performance in case that your collision object has a _btScaledBvhTriangleMeshShape_ shape.

The ray positions (from and to) are divided by the scale of the shape, and then I use this new ray positions to make a **performRay()** call with the underlying _btBvhTriangleMeshShape_.

That causes an incredible performance boost when using scaled meshes (10-20x in our project).

Best regards,
David Ávila, 
WaveEngine team member.


erwincoumans:
The appveyor CI is flaky, it fails in some copy file operation. We need to re-trigger CI, perhaps make a trivial change (white space etc)?

davilovick:
Thanks! We are working now to integrate Bullet as the default 3D Physic engine :)

erwincoumans:
Sounds good, let's merge it!

PR #1398, opened 2017-10-22
Added gear info to getConstraintInfo return.

jonathantompson:


jonathantompson:
Obviously, this will also require adding rows to the getConstraintInfo doc in the pybullet quickstart guide...

jonathantompson:
Ahhh, actually it looks like these gear fields are never set on the server side. They come back all zeros.

Debugging now.

jonathantompson:
OK, found the bug. I'll close this PR and start a new one with the code changes.

PR #1404, opened 2017-10-24
 perform IK in local body-fixed frame ,  add jointFrictionDamping.py example

erwincoumans:


erwincoumans:
See also this video: https://www.youtube.com/watch?v=qZiR6vNcmCU

PR #1408, opened 2017-10-27
adjust the minitaur_derpy urdf. Seperate the motor controller and the…

jietan:
… moving part of the motor. Fine tune the moment of inertia.


erwincoumans:
Thanks for the contribution. At some stage, we'll update the Minitaur scripts to use the updated URDF files (probably create a new environment -v1, to keep the old the same).

PR #1340, opened 2017-09-26
CMakeLists.txt search for any python version by default

ahundt:
Could https://github.com/bulletphysics/bullet3/pull/1334 be reverted so this can be applied?

See #1131 and https://github.com/bulletphysics/bullet3/issues/1322 for details on the evolution of these changes.

PR #1409, opened 2017-10-27
only probe for python if PYBULLET explicitly specified

jl:
On Ubuntu 16.04, despite using `-DBUILD_PYBULLET=OFF`, the build fails with the following error:
```
CMake Error at build3/cmake/FindPythonLibs.cmake:118 (message):
  Could not find required file 'FindLibPython.py'
Call Stack (most recent call first):
  CMakeLists.txt:317 (FIND_PACKAGE)
```

If we look at the relevant code in CMakeLists.txt:
```
FIND_PACKAGE(PythonLibs ${PYTHON_VERSION_STRING} EXACT)
```
which FATAL_ERRORs out when it encounters this logic in [FindPythonLibs.cmake](https://github.com/bulletphysics/bullet3/blob/master/build3/cmake/FindPythonLibs.cmake)
```
if(NOT _FIND_LIB_PYTHON_PY)
    message(FATAL_ERROR "Could not find required file 'FindLibPython.py'")
endif()
```

This pull request changes the build behavior so that the environment is not probed at all for any Python libraries unless `BUILD_PYBULLET` is explicitly set to `ON`. The previous behavior was to always probe the environment, then set the default value of `BUILD_PYBULLET` based on whether the probe succeeds.

With this change, developers not interested in PYBULLET can isolate themselves from problems related to Python build going forward. The downside is folks that want PYBULLET and previously relied on the automagic probing behavior will now need to explicitly turn it ON.

Previous work: #1340

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,25 +301,26 @@ IF (APPLE)
 ENDIF()
 
 OPTION(BUILD_BULLET3 "Set when you want to build Bullet 3" ON)
+
 # Optional Python configuration
-# builds pybullet automatically if all the requirements are met
-SET(PYTHON_VERSION_PYBULLET "" CACHE STRING "Python version pybullet will use.")
-SET(Python_ADDITIONAL_VERSIONS 3 3.6 3.5 3.4 3.3 3.2 3.1 3.0 2.7 2.7.12 2.7.10 2.7.3 )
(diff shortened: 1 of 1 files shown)


erwincoumans:
Thank you very much, indeed we shouldn't fail if people don't request Python/pybullet!

PR #1410, opened 2017-10-28
Composite fixes for mingw64 and msvc.

d3x0r:
Composite of fixed code for mingw64(on windows) and general fixes for msvc.

1) (win32window) don't convert char to wide, use char direct to window.
2) (CMakeLists) Don't link one library as static CPPLIB and no others (mismatched allocations)
3) (macros) Fix Gwen macros for mingw64 on windows build.  (changes are by compiler(msc_ver) not platform)
4) (FileUtils) sprintf_s reference by platform, not compiler (mingw64 support)
5) (b3OpenCLUtils) fix bad define name _MSVC_VER->_MSC_VER
6) (compoundCollision) remove unused variables, simplify operation.
7) (impulseconstraint) remove duplicated code block

### DIFF
--- a/examples/OpenGLWindow/Win32Window.cpp
+++ b/examples/OpenGLWindow/Win32Window.cpp
@@ -442,6 +442,7 @@ void Win32Window::setWindowTitle(const char* titleChar)
 #ifdef _WIN64
 		SetWindowText(m_data->m_hWnd, titleChar);
 #else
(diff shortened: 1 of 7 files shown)


erwincoumans:
Thanks for the effort and patience. I reviewed those changes and looks good to me.

PR #1412, opened 2017-10-30
use Ansi version of LoadLibrary

RicoP:
I got the error "C2664: 'HMODULE LoadLibraryW(LPCWSTR)': cannot convert argument 1 from 'const char *' to 'LPCWSTR'". 
Mapping B3_DYNLIB_OPEN to the Ansi version of LoadLibrary fixed the error for me.


erwincoumans:
Thanks for the contribution, let's apply it.

PR #1419, opened 2017-11-02
Update support for buggy android NDK

kwasimensah:
Expand logic for when __thread is supported to work around https://github.com/android-ndk/ndk/issues/8

### DIFF
--- a/src/LinearMath/btQuickprof.cpp
+++ b/src/LinearMath/btQuickprof.cpp
@@ -680,46 +680,60 @@ void	CProfileManager::dumpAll()
 	CProfileManager::Release_Iterator(profileIterator);
 }
 
+// BT_HAVE_TLS is defined to 1 when __thread should be supported.
+// We assume __thread is supported on Linux when compiled with Clang or compiled
+// against libstdc++ with _GLIBCXX_HAVE_TLS defined.
+#ifdef BT_HAVE_TLS
+#error BT_HAVE_TLS cannot be directly set
+#elif defined(__linux__) && (defined(__clang__) || defined(_GLIBCXX_HAVE_TLS))
+#define BT_HAVE_TLS 1
+#endif
 
-
+// There are platforms for which TLS should not be used even though the compiler
+// makes it seem like it's supported (Android NDK < r12b for example).
+// This is primarily because of linker problems and toolchain misconfiguration:
+// TLS isn't supported until NDK r12b per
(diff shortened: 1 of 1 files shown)


erwincoumans:
OK, let me have a look and merge it today, Thanks!

erwincoumans:
Mac OSX platform is affected by this change. Could you rewrite the patch, using  
```
#if defined(__ANDROID__) 
 //some new logic, 
#else 
unmodified existing logic 
#endif
```
then there is less chance it breaks other platforms.

kwasimensah:
Ok, pushed new version. Will wait for tests to pass

kwasimensah:
Tests have passed. I think this can be integrated now.

erwincoumans:
Since the logic is more complex than #if defined(__ANDROID__)  #else <unmodified> #endif I'll have to manually go through the code and test if this doesn't break on any platform.

erwincoumans:
Thanks for the pull request!

erwincoumans:
OMG, this patch seems to be broken indeed: https://github.com/bulletphysics/bullet3/commit/8aefa7e8d05118a09d9bb3a4c49db7678391b7b5#commitcomment-25547649

So it gets reverted and I'll have to manually change the logic to not break things....

kwasimensah:
Pull request is at https://github.com/bulletphysics/bullet3/pull/1431

AndresTraks:
The new patch fixes it.

BT_THREADSAFE is defined when BULLET2_USE_THREAD_LOCKS is set in CMake, so you'd have to build two configurations if you want to test this. It makes sense for it to be a preprocessor macro, because otherwise, with the way multi-threading is done now, an "isThreadSafe" variable would have to be checked a lot during simulation.

As an exception, I think btQuickProf should be made thread-safe by default or at least redesigned to not use static variables, because it crashes in the case where you want to simulate completely separate worlds in separate threads, which has nothing to do with Bullet's multi-threading functionality: http://bulletphysics.org/Bullet/phpBB3/viewtopic.php?t=5588

erwincoumans:
Good point. btQuickProf/BT_PROFILE is used for benchmarking, and we mainly use the newer and better b3ProfileZone/B3_PROFILE instead. At some stage, I'll have to retire LinearMath and move to Bullet3Common, have to figure out when/how to avoid breaking lots of code bases.

PR #1421, opened 2017-11-04
Update btTriangleIndexVertexArray.h

elect86:
typo


erwincoumans:
thanks for fixing it!

Issue #1197, opened 2017-06-19
Wrong depth values using Tiny Renderer


mbreyer:
In the small example below, Tiny Renderer returns incorrect depth values that do not match the values returned by the OpenGL renderer. The depth values also seem to vary strongly with different near and far planes.

```
import matplotlib.pyplot as plt
import numpy as np
import pybullet as p

p.connect(p.GUI)
p.loadURDF('plane.urdf')
p.loadURDF('cube_small.urdf', basePosition=[0.0, 0.0, 0.025])

width = 128
height = 128

fov = 60
aspect = width / height
near = 0.02
far = 1

view_matrix = p.computeViewMatrix([0, 0, 0.5], [0, 0, 0], [1, 0, 0])
projection_matrix = p.computeProjectionMatrixFOV(fov, aspect, near, far)

# Get depth values using the OpenGL renderer
images = p.getCameraImage(width, height, view_matrix, projection_matrix, renderer=p.ER_BULLET_HARDWARE_OPENGL)
depth_buffer_opengl = np.reshape(images[3], [width, height])
depth_opengl = far * near / (far - (far - near) * depth_buffer_opengl)

# Get depth values using Tiny renderer
images = p.getCameraImage(width, height, view_matrix, projection_matrix, renderer=p.ER_TINY_RENDERER)
depth_buffer_tiny = np.reshape(images[3], [width, height])
depth_tiny = far * near / (far - (far - near) * depth_buffer_tiny)

p.disconnect()

# Plot both images - should show depth values of 0.45 over the cube and 0.5 over the plane
plt.subplot(1, 2, 1)
plt.imshow(depth_opengl, cmap='gray', vmin=0, vmax=1)
plt.title('OpenGL Renderer')

plt.subplot(1, 2, 2)
plt.imshow(depth_tiny, cmap='gray', vmin=0, vmax=1)
plt.title('Tiny Renderer')

plt.show()
```

erwincoumans:
TinyRenderer depth values are converted to OpenGL style depth values here:
https://github.com/bulletphysics/bullet3/blob/master/examples/SharedMemory/TinyRendererVisualShapeConverter.cpp#L971
If you can, please share your improvements!

mbreyer:
That is why I use the same transform to get the original depth values for both OpenGL and TinyRenderer. However, the values are slightly different, for example 0.48 instead of 0.5.

erwincoumans:
[...]

PR #1439, opened 2017-11-17
Fix wrong depth values returned by TinyRenderer

mbreyer:
This pull request addresses Issue #1197 in which I observed that TinyRenderer depth values are incorrect. This can be seen in the following image generated with this [script](https://gist.github.com/mbreyer/ba7db03a3a2834d6b9ca1285dfe102fd), where images in the first row were rendered with OpenGL and images in the second row were rendered with TinyRenderer.

figure_1

The problem was not precision, but an error in the transformation pipeline. The part of the code that transforms depth values to OpenGL style depth buffer assumes that the input is the distance from camera eye to the object, see this [article](https://sjbaker.org/steve/omniv/love_your_z_buffer.html). However, the value stored in m_data->m_depthBuffer is actually given in clip coordinates. In my first commit, I transform these values to camera eye coordinates according to this [article](http://www.songho.ca/opengl/gl_projectionmatrix.html), before they are transformed to OpenGL style depth buffer.

Second, in our_gl, fragments were not discarded properly leading to objects disappearing, even though they were further away than the near_plane (also seen in the above figure). After addressing these issues, I get the same output with TinyRenderer and OpenGL

figure_2

Lastly, if all triangles get discarded because they are too close or too far away, OpenGL’s depth buffer contains the maximum value of 1. My third commit mimics this behavior in TinyRenderer by appropriately initializing the buffer.

### DIFF
--- a/examples/SharedMemory/TinyRendererVisualShapeConverter.cpp
+++ b/examples/SharedMemory/TinyRendererVisualShapeConverter.cpp
@@ -772,12 +772,13 @@ void TinyRendererVisualShapeConverter::resetCamera(float camDist, float yaw, flo
 
 void TinyRendererVisualShapeConverter::clearBuffers(TGAColor& clearColor)
 {
+	float farPlane = m_data->m_camera.getCameraFrustumFar();
     for(int y=0;y<m_data->m_swHeight;++y)
     {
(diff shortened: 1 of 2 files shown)


erwincoumans:
Great, thanks for sharing your improvements. Will look at it soon.

PR #1458, opened 2017-11-25
add const to masses for calculatePrincipalAxisTransform

jl:
`masses` is treated as read-only in `btCompoundShape::calculatePrincipalAxisTransform`, so it can be declared as a `const` parameter so that callers can also use read-only data to feed into this function.


erwincoumans:
Yes, those masses should be const indeed. Thanks for the contribution!

PR #1471, opened 2017-12-08
FindPythonLibs.cmake add missing quotes

ahundt:
This avoids problems in the case that the the path has a space in it.


erwincoumans:
Thanks, looks good to me.

Issue #1450, opened 2017-11-22
Building with NMake (VS2017) produces warning D9025 : overriding '/arch:SSE' with '/arch:SSE2'


avikde:
My system:
* Windows 10 with latest updates
* Installed the VS2017 build tools from https://www.visualstudio.com/downloads/#build-tools-for-visual-studio-2017
* cmake 3.9.6 from https://cmake.org/download/ (latest)

Steps:
* pulled the latest source, opened a developer command prompt for VS2017, and roughly followed the bat file:

```
> cd bullet3
> mkdir build_cmake
> cd build_cmake
> cmake .. -DBUILD_PYBULLET=OFF -DUSE_DOUBLE_PRECISION=ON -DCMAKE_BUILD_TYPE=Release -G "NMake Makefiles"
> nmake
```

Output (for each compilation line)
```
[ 20%] Building CXX object examples/ThirdPartyLibs/BussIK/CMakeFiles/BussIK.dir/Misc.obj
cl : Command line warning D9025 : overriding '/arch:SSE' with '/arch:SSE2'
Misc.cpp
```

The libraries build, and the examples run fine, only the warning is annoying.

I tried googling but don't understand enough about VS with cmake to know how to fix it. I also tried `cmake .. -DUSE_MSVC_SSE2=ON -DUSE_MSVC_AVX=OFF ....` but that had the same result.

erwincoumans:
You could try to use USE_MSVC_SSE=OFF?
It looks like a silly warning, https://stackoverflow.com/questions/1067630/sse2-option-in-visual-c-x64

avikde:
That fixed it! Not sure why I didn't try that before.  For reference,
```
> cmake .. -DUSE_MSVC_SSE=OFF <other options> -G "NMake Makefiles"
```

PR #1476, opened 2017-12-12
Removed default ON for USE_MSVC_SSE (warning on recent 64 bit windows)

avikde:
This is related to https://github.com/bulletphysics/bullet3/issues/1450.

SSE and SSE2 conflict with this [command line warning D9025](https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/command-line-warning-d9025), and as [observed here](https://stackoverflow.com/questions/1067630/sse2-option-in-visual-c-x64) recent 64-bit windows versions support SSE2.

Just to simplify building on most recent versions of Windows and remove warnings, it may be preferable to have this default be off.

It doesn't affect other platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,7 @@ IF(MSVC)
 	IF (CMAKE_CL_64)
 	  ADD_DEFINITIONS(-D_WIN64)
 	ELSE()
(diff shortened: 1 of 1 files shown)


erwincoumans:
Good point, thanks for the contribution!

PR #1478, opened 2017-12-14
add system id for rainbow dash (second generation minitaur with TX2 i…

jietan:
…nstalled)


erwincoumans:
Thanks for the contribution!

Issue #1480, opened 2017-12-19
Is it possible to use the discrete GPU for rendering instead of the integrated one?


avikde:
This is more of a query than a request: I have a laptop with integrated as well as discrete GPU (Windows 10) and bullet3 only uses the integrated one. Is there a way to make it use the discrete GPU when available? MATLAB (for e.g.) uses the dGPU when it opens so I think it's possible for the program to request or select it.

If you could point me to where in the OpenGL init code I should dig, I can look myself as well.

Thanks!

EDIT: I just looked through some files in ExampleBrowser - would it be there? Feel free to close if this is a GWEN or other library issue.

erwincoumans:
It is not enabled indeed, our home-grown implementation is pretty simple.
The OpenGLWindow implementation is all here:
https://github.com/bulletphysics/bullet3/blob/master/examples/OpenGLWindow/Win32OpenGLWindow.cpp


Recently I added some option to enable GLFW, it could be worth trying out.

It is only enabled with premake4, see the build3/premake4.lua file.

newoption
    {
trigger     = "glfw_include_dir",
value       = default_glfw_include_dir,
description = "GLFW 3.x include directory"
    }

newoption
    {
                        trigger     = "glfw_lib_name",
                        value       = default_glfw_lib_name,
                        description = "GLFW 3.x library name (glfw, glfw3)"
    }

    newoption
    {
trigger     = "glfw_lib_dir",
value       = default_glfw_lib_dir,
description = "(optional) GLFW 3.x library directory "
    }

    newoption
    {
trigger     = "enable_glfw",
value       = false,
description = "(optional) use GLFW 3.x library"
    }

if _OPTIONS["enable_glfw"] then
defines {"B3_USE_GLFW"}

So if GLFW has that feature, you may use that, or backport it to Bullet
OpenGL.

Let me know how it goes, or if you need some further help,
Thanks!
Erwin

avikde:
I found a solution for NVidia specifically https://docs.nvidia.com/gameworks/content/technologies/desktop/optimus.htm and created a pull request that makes the request on windows.

avikde:
Solution found within user code, so I'm closing this

PR #1481, opened 2017-12-20
Request nvidia discrete GPU if available (no effect if NVidia GPU is …

avikde:
This is related to #1480 

The NVidia drivers check for this global variable and request the discrete GPU if available. I tested with the GPU connected and not, and it runs fine in both cases (just uses the integrated if the nvidia GPU is not available). 

EDIT: more information: https://docs.nvidia.com/gameworks/content/technologies/desktop/optimus.htm

### DIFF
--- a/examples/OpenGLWindow/Win32OpenGLWindow.cpp
+++ b/examples/OpenGLWindow/Win32OpenGLWindow.cpp
@@ -30,6 +30,11 @@ static void printGLString(const char *name, GLenum s) {
   printf("%s = %s\n",name, v);
 }
 
+// Request NVidia discrete GPU if both are available
+extern "C" {
+__declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
+}
+
 bool sOpenGLVerbose = true;
 
 void Win32OpenGLWindow::enableOpenGL()


erwincoumans:
Thanks for looking into this. I'd like to add an API to disable it (someone may want to use integrated GPU to save battery life etc), and AMD support.

~~~~~
extern "C" {
__declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
}
~~~~~
(from https://www.gamedev.net/forums/topic/679029-selecting-the-actual-gpu/)

avikde:
I can't test the AMD option but it seems harmless to add.

I see your point about the API, but the only way I see to do it is with a preprocessor directive (maybe set using cmake -- let me know if I should try this)... I'm not sure of a run-time way to let the user choose.

erwincoumans:
~~~~~~~~
static int gDiscreteGpuEnabled = 1;
__declspec(dllexport) DWORD NvOptimusEnablement = gDiscreteGpuEnabled;
~~~~~~~~
If you set gDiscreteGpuEnabled to 0, before initializing the window, does it still select the discrete GPU?

erwincoumans:
Let's first just enable it, and deal with it if/when someone has the issue.

avikde:
I tried a few things but it seems like (unlike what https://docs.nvidia.com/gameworks/content/technologies/desktop/optimus.htm says), even declaring 

```
extern "C" {
__declspec(dllexport) DWORD NvOptimusEnablement = 0x00000000;
}
```

enables the dGPU, even though the assigned value is 0 (without anything even changing its value in the program).

I guess till there's a more robust solution I'll just declare the global var in my user code and we can discard this PR for now...

PR #1484, opened 2017-12-23
Support for reading <inertial> elements from MJCF

rhaps0dy:
I'm trying to replicate an analytic CartPole environment in Bullet, and to do that accurately I needed to specify the moments of inertia. This patch is part of the solution and adds support for that.

I'm not very sure about [this line](https://github.com/bulletphysics/bullet3/compare/master...rhaps0dy:MJCFImporter?expand=1#diff-3fcc3b2c48d9839736d2f098bc4ca97aR299), is it always the case that bodies without a set `localInertiaDiagonal` have it set to 0? Empirical observation suggests so but my tests haven't been very extense.

More information in the issue that I'm going to link.


erwincoumans:
Thanks for the contribution!

PR #1499, opened 2018-01-09
Tidy up cmake build script

wrosecrans:
Minor improvements to tidy up build_cmake_pybullet_double.sh

Existing version:
-started with an rm of the CMakeCache.txt, which won't exist on first build, printing an error
-did a mkdir build_cmake which already exists on teh second build, printing an error.
-make command used a hardcoded "-j12" to always try to use 12 processes regardless of the number of CPU's on the machine.  
-The symlink for pybullet.so would error after the first build because pybullet.so already exists.

This PR will
-Check for the existence of CMakeCache.txt before trying to delete it.
-Add a -p flag to mkdir so it won't error when repeatedly making the build_cmake directory.
-Use the nproc command if available to run as many build processes as there are CPU's in the system. (Or preserve old behavior and use -j12 if it isn't available)
- Tidy up the pybullet.so symlink so it is only attempted if a .dylib exists, and ln is set to force mode so it won't complain if pybullet.so already exists.
- Add a friendly message saying "Completed build of Bullet." to increase the joy of the user.  Message is displayed only if script didn't bail early because cmake or make failed.

Result is fewer distracting warning messages will be printed in a normal build, and build may be faster on machines with many cores and fast storage.

### DIFF
--- a/build_cmake_pybullet_double.sh
+++ b/build_cmake_pybullet_double.sh
@@ -6,7 +6,7 @@ fi
 mkdir -p build_cmake
 cd build_cmake
 cmake -DBUILD_PYBULLET=ON -DBUILD_PYBULLET_NUMPY=OFF -DUSE_DOUBLE_PRECISION=ON -DCMAKE_BUILD_TYPE=Release ..
-make -j $(command nproc || echo 12)
+make -j $(command nproc 2>/dev/null || echo 12)
 cd examples
 cd pybullet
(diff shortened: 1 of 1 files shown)


erwincoumans:
Good point, thanks for the contribution. Nowaways I mainly use pip (or python setup.py develop) to install pybullet, and premake4 for development, but good to improve cmake as well.

PR #1500, opened 2018-01-09
Load softbody from obj file.

YunfeiBai:


erwincoumans:
Great, thanks for the contribution. We may want to create a softbody from vertices/indices as well, and re-use the same command.

erwincoumans:

PR #1506, opened 2018-01-12
allow to use colors from MJCF file as option (default to random Googl…

erwincoumans:
…e colors), use p.loadMJCF(filename, flags=p.URDF_MJCF_COLORS_FROM_FILE

fix quadruped.py example.
add PyBullet.isConnected() API, more friendly than PyBullet.getConnectionInfo()["connected"]


GPaolo:
Hi, for me adding the flag URDF_MJCF_COLORS_FROM_FILE does not work. It says that it does not exists.

I am on version 2.5.3

GPaolo:
Hi, for me adding the flag URDF_MJCF_COLORS_FROM_FILE does not work. It says that it does not exists.

I am on version 2.5.3

anqixu:
Apparently it has been renamed.
This works for me: `pybullet.MJCF_COLORS_FROM_FILE`
(`pybullet==2.8.1`)

PR #1508, opened 2018-01-12
Added kuka environment with multiple diverse objects.

deq2:


erwincoumans:
Thanks for the contribution, it looks like a nice Gym environment! 

The large amount of random files (3000) is an issue, it degrades performance of git and pip etc.
We could add a zipfile of all, and implement some zip reading on the server side, or have a manual step to unzip the archive.

Can you re-create the pull request and remove all the random objects? I'll make sure to deal with them separately.

I added some maxVelocity  to avoid the abrupt/shaky motion: https://github.com/bulletphysics/bullet3/pull/1515

1) Change action repeat to 75 (from 50).
2) Twice add a if self._renders:    	time.sleep(self._timeStep) after p.stepSimulation
3) Change near plane from 0.1 to 0.01 make the z-buffer show up (even though the env doesn't seem to use it).

I'm happy to make those small changes, if you re-create the pull request without all the random files (I already made a zipfile of them)

deq2:
Hi Erwin, thanks for the quick feedback. I updated the pull request to not include the random urdfs- If you have time to make the small changes, maybe that would be easier for me (I'm not sure what you mean exactly by the near plane). Thanks so much!

deq2:
Let me know if there is anything else I can do to help!

erwincoumans:
This PR will improve the environment and add the zipfile. At the moment, the zipfile needs to be manually extracted first. When I have time, I add zip reading in PyBullet.
https://github.com/bulletphysics/bullet3/pull/1524
See also a Youtube video of your environment in action:
https://www.youtube.com/watch?v=y7scmgsGQdA&feature=youtu.be

kargarisaac:
Hi, 
thank you for great env.
I have some questions.
1. Is the state of KukaDiverseObjectEnv and KukaCamGymEnv only rgb image? how about depth image? 
2. Is there any example for training KukaDiverseObjectEnv? I get the following error when I use the script for KukaCamGymEnv:

```
  File "mine/kuka_arm_train.py", line 119, in <module>
    main()
  File "mine/kuka_arm_train.py", line 72, in main
    callback=callback
  File "/home/isaac/miniconda3/envs/rl/lib/python3.5/site-packages/baselines/deepq/simple.py", line 173, in learn
    observation_space_shape = env.observation_space.shape
AttributeError: 'NoneType' object has no attribute 'shape'
```

PR #1511, opened 2018-01-15
fix conversion from getCameraImage output to rgb_array

mwest1066:
This fixes a bug where `pybullet_envs.agents.visualize_ppo` fails with an error like:
```
  File "/Users/mwest/anaconda3/envs/DRL/lib/python3.6/site-packages/pybullet_envs/env_bases.py", line 101, in _render
    rgb_array = rgb_array[:, :, :3]
IndexError: too many indices for array
```
The array returned from `getCameraImage()` is flat and needs to be reshaped to `(height, width, 4)`.

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/env_bases.py
+++ b/examples/pybullet/gym/pybullet_envs/env_bases.py
@@ -96,7 +96,8 @@ class MJCFBaseBulletEnv(gym.Env):
 			projectionMatrix=proj_matrix,
 			renderer=p.ER_BULLET_HARDWARE_OPENGL
 			)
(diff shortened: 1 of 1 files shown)


erwincoumans:
Thanks for the report. I'm not getting that issue. Could it be you compiled PyBullet without numpy support? How did you build PyBullet?

erwincoumans:
When using numpy to compile PyBullet, it is not a flat array, see the code here:
https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/pybullet.c#L6595

erwincoumans:
this was previously fixed, closing.

PR #1519, opened 2018-01-17
Don't predictUnconstraintMotion for inactive objects

Niadb:
There doesn't seem to be a need to applyDamping and predictIntegratedTransform for bodies that are not active.

Profiling showed my app was spending 2/7 of physics update time in here despite most objects being inactive. 

If you prefer a flag to turn this on/off let me know


dyanikoglu:
@erwincoumans Any update on this PR?

I was able to track same problem in my own custom engine implementation with lots of sleeping objects laying on the map. Also, same change also can be applied into Multithreaded variant of the same class.

I'm not sure if those memory allocation changes are related with that issue btw @Niadb ? They seem useful, but not sure about possible consequences. Out of curiousity, I just gave them a quick try, and seems like they don't cause any problems, my own unit tests are all passing without problem.

Niadb:
@dyanikoglu 
You are correct the change to btAlignedObjectArray.h was not supposed to be in this changelist--whoops!

PR #1523, opened 2018-01-23
[fixed] gym_manipulator_envs typo

kimhc6028:
It seems there was typo: BaseBulletEnv -> MJCFBaseBulletEnv


benelot:
Right, that is a typo indeed. It seems that the renaming did not work there.

erwincoumans:
Thanks!

PR #1527, opened 2018-01-25
Removed warnings of unused variable in btThreads header

Breush:
I don't care too much about warnings when building the library,
but these where generated using it (it's in a, header file).


erwincoumans:
thanks for the fixes!

Issue #1525, opened 2018-01-25
No module name baseline enjoy_kuka_diverse_object_grasping


AndreaIncertiDelmonte:
Hi,
I was trying to run the demo showed in this video https://www.youtube.com/watch?v=y7scmgsGQdA

Inside a python 3.5 virtual env I ran the following commands:
```
pip3 install pybullet tensorflow gym
python -m pybullet_envs.baselines.enjoy_kuka_diverse_object_grasping
```
and I got this import error:
```
bullet3/examples/pybullet/gym/pybullet_envs/baselines/enjoy_kuka_grasping.py
from baseline import deepq
ImportError: No module name 'baselines'
```
I also tried to install pybulled from source same results.

It seems the module deepq is missing.

Thnks
Andrea

erwincoumans:
That baselines line shouldn't be there. For the time being, try use pip install baselines?

PR #1531, opened 2018-01-26
Override of the reset() public funciton in child classes

AndreaIncertiDelmonte:
Hi Erviw,
I followed your instruction about the issue https://github.com/bulletphysics/bullet3/issues/1525 
Then I discovered that the classes KukaGymEnv and KukaDiverseObjectEnv call the function reset() that calls the _reset() function from the Env class.

In this PR i just added the override of the reset() funciton in the child classes.

Andrea


erwincoumans:
Thanks, looks good to me. I'll update the pypi repo version soon.

PR #1532, opened 2018-01-26
avoid errors in some ide docs system

cedspam:
there are some error linked to repeated import of this module in spyder, an env with the same id cannot be registered again


erwincoumans:

PR #1535, opened 2018-01-26
no need to keep iterating if an object is found awake or unable to sleep

CuriousRubick:
title is wrong, should say "no need to keep iterating if an object is found awake or unable to sleep."


erwincoumans:

PR #1542, opened 2018-02-01
Unknown character causing syntax error

ishaanv:
After downloading, install and then running an example training file from the latest master, I got this message:
`
  File "enjoy_kuka_grasping.py", line 8, in <module>
    from pybullet_envs.bullet.kukaGymEnv import KukaGymEnv
  File "/Users/ishaanvarshney/bvn/bullet3/examples/pybullet/gym/pybullet_envs/__init__.py", line 4
    if id in  registry.env_specs:
            ^
SyntaxError: invalid character in identifier
`

This PR removes this addition unnecessary character.

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/__init__.py
+++ b/examples/pybullet/gym/pybullet_envs/__init__.py
@@ -1,7 +1,7 @@
 import gym
 from gym.envs.registration import registry, make, spec
 def register(id,*args,**kvargs):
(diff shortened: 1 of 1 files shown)


erwincoumans:
This was fixed yesterday (see https://github.com/bulletphysics/bullet3/pull/1540/commits/cb0e7404221b92cfb85ba2765a20342b9f69a026) can you do an update?

PR #1548, opened 2018-02-05
fix install bug

xiaoerlaigeid:
I install on win10 , finding bug "error: 'strlen' was not declared in this scope"


examples/ExampleBrowser/GwenGUISupport/GwenParameterInterface.cpp
```diff
@@ -1,6 +1,6 @@
 #include "GwenParameterInterface.h"
 #include "gwenInternalData.h"
-
+#incldue <cstring>
```
jslee02: This is a typo. `#include` should be `#include`.

PR #1549, opened 2018-02-06
test clang_format google-style

erwincoumans:
test running clang-format using latest clang-format 7 and unmodified google style.


nbelakovski:
Some tidying of the styling would be great! Do you plan to set this up as a commit hook of some sort?

PR #1553, opened 2018-02-07
createCloth python binding

JanMatas:
**Added python command `createCloth` that creates a 2D rectangular softBody.  **
Hi, 

I needed a way to create a 2D deformable object (cloth), so I created the python binding that exposes the cloth creation as well as most common params governing the cloth dynamics.

I understand that this might too specific to be present in the API but there is currently no other way to do it from Python. If you could recommend a better way to accomplish the same thing, I am happy to close this PR and implement it differently. Any feedback would be appreciated!

See example `examples/pybullet/examples/createCloth.py`.


erwincoumans:
Thanks for the contribution.  It is useful to create some 'experimental' APIs for soft body. I'll look at it soon and give some feedback.

PR #1555, opened 2018-02-08
Implemented support for softbodies in OpenGL renderer

JanMatas:
Hi,

I implemented the support for softbodies in OpenGL  renderer. The main use cases are:
- softbodies are now visible in p.getCameraImage images
- softbodies now get textures and shadows in example browser

Known issues:
- I believe this could be problematic with 2D softBodies (as cloth in PR https://github.com/bulletphysics/bullet3/pull/1553) because of Face Culling. I will address this in separate PR is this one gets accepted.

Performance considerations:
- I am a bit worried that we need to write each vertex position on each syncPhysicsToGraphics call, but I do not currently see a better option. 

Screen from load_soft_body example:
screen shot 2018-02-08 at 13 46 50

### DIFF
--- a/examples/ExampleBrowser/OpenGLGuiHelper.cpp
+++ b/examples/ExampleBrowser/OpenGLGuiHelper.cpp
@@ -1291,8 +1291,8 @@ void OpenGLGuiHelper::computeSoftBodyVertices(btCollisionShape* collisionShape,
 											  btAlignedObjectArray<GLInstanceVertex>& gfxVertices,
 											  btAlignedObjectArray<int>& indices)
 {
-	b3Assert(collisionShape->getUserPointer());
-	btSoftBody* psb = (btSoftBody*)collisionShape->getUserPointer();
+	b3Assert(collisionShape->getBodyPointer());
+	btSoftBody* psb = (btSoftBody*)collisionShape->getBodyPointer();
 	gfxVertices.resize(psb->m_faces.size() * 3);
 	int i, j, k;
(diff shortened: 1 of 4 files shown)


erwincoumans:
That's great, thanks for sharing your work!
It would be nice to keep using the getUserPointer/setUserPointer. Why did you add a new method?

JanMatas:
Hi Erwin,

I just felt that userPointer should be left free for bullet library users in their applications instead of populating it inside the library code. I do not have a strong opinion on it and I am very happy to roll back the second commit if you prefer to use userPtr.

erwincoumans:
PyBullet and BulletRobotics (the[ C-API](https://github.com/bulletphysics/bullet3/blob/master/examples/SharedMemory/PhysicsClientC_API.h) and [C++ API](https://github.com/bulletphysics/bullet3/blob/master/examples/RobotSimulator/b3RobotSimulatorClientAPI_NoGUI.h)) don't expose Bullet directly, so there is no need to leave it free. Please revert that commit and I'll look at merging your PR.

JanMatas:
Thanks for the insight! I reverted the commit in question. Please let me know if you have any other feedback I should address.

erwincoumans:
Let's merge it for now, and I add a couple of improvements.

erwincoumans:
Here is a follow up pull request with some performance improvements:
https://github.com/bulletphysics/bullet3/pull/1568

We still need to sync the vertices from the server thread using double-buffering, to avoid flickering in the rendering (graphics is running in a separate thread from physics, and no synchronization happens).

PR #1559, opened 2018-02-12
minor bugfix in image reshaping in kukaGymEnv.py

shubhtuls:
Reshaped image array should be of size (RENDER_HEIGHT, RENDER_WIDTH, 4) instead of (RENDER_WIDTH, RENDER_HEIGHT, 4).


erwincoumans:
Good point, thanks!

PR #1560, opened 2018-02-13
Add shader and example for projective texture.

YunfeiBai:


erwincoumans:
Great, thanks for the contribution. Let me have a look first and then merge it soon.
Thanks!

erwincoumans:
The patch reverts an improvement/fix in SortableTransparentInstance:
we need to pre-compute the dot product, not compute it during the sorting.

This is how it should be:
~~~~~~~~
B3_ATTRIBUTE_ALIGNED16(struct) SortableTransparentInstance
{
	b3Scalar m_projection;

	int m_shapeIndex;
	int m_instanceId;
};

B3_ATTRIBUTE_ALIGNED16(struct) TransparentDistanceSortPredicate
{

	inline bool operator() (const SortableTransparentInstance& a, const SortableTransparentInstance& b) const 
	{
		
		return (a.m_projection > b.m_projection);
	}
};
~~~~~~~~

erwincoumans:
The pull request adds many (247) files under build3/xcode etc. I'll see how to many apply the pull request without them.

YunfeiBai:
I have reverted the previous commit, and removed those unnecessary files.

erwincoumans:
Thanks, we just need to add some APIs to enable it (instead of hard-coding the change), right?

YunfeiBai:
Yes, I plan to work on exposing that and also other projector parameters in the API.

Issue #1563, opened 2018-02-16
Points generated by angular perturbation are wrong when object A is rotated


radarsat1:
The following gist demonstrates that the location of contact points depends on the order of calls to `btCollisionWorld::addCollisionObject`.

https://gist.github.com/radarsat1/2eb69623b46ee10dfdb52ac617bcc658

It can be run with,

    ./collision_points_object_order | python - &

The program just sets up a collision world with a smaller box as a btConvexHullShape sitting on top of a larger box as a btBoxShape.
With current master (e87060dcf4b445c5c041933b60b9deff9ef98568) it produces the following output:

![image](https://user-images.githubusercontent.com/66805/36311249-73a25246-1309-11e8-8805-8d2de7f52004.png)

Here, points on object A (i.e. getBody0) are blue, points on object B (getBody1) are red.  It seems that btConvexConvexAlgorithm correctly always selects to perturb the smaller object (the convex hull) but when it is A it generates the wrong points.  That is, they are not on the surface of either object.  I think the same points should be generated in both cases.  I think it is a wrong direction of the normal, or a wrong calculation.  I found a possible solution but am still verifying it.  I will submit a PR soon.

PR #1564, opened 2018-02-16
Fix calculation of startPt for perturbed object A.

radarsat1:
I believe this change in sign of the normal when calculating startPt will fix the calculation of contact points when object A is selected to be perturbed.  Indeed, in the object A case, first endPt (on object A surface) is calculated and then the startPt should be calculated following the normal **backwards** towards A surface, whereas currently the code moves along the normal forwards, into the B surface.

Since the contact point is then specified by the startPt location and normal, the result is a vector starting inside the B contour and pointing into it, instead of starting on the A surface and pointing to the B surface.

Regarding the test program mentioned in #1563, with this change, the following output is generated:

image

Similarly if the objects are switched position so that the convex hull is on the bottom, previous to this change in sign it was:

image

but now is:

image

Similar results can be seen if the objects are rotated slightly.

### DIFF
--- a/src/BulletCollision/CollisionDispatch/btConvexConvexAlgorithm.cpp
+++ b/src/BulletCollision/CollisionDispatch/btConvexConvexAlgorithm.cpp
@@ -259,7 +259,7 @@ struct btPerturbedContactResult : public btManifoldResult
 			btVector3 endPtOrg = pointInWorld + normalOnBInWorld*orgDepth;
 			endPt = (m_unPerturbedTransform*m_transformA.inverse())(endPtOrg);
 			newDepth = (endPt -  pointInWorld).dot(normalOnBInWorld);
-			startPt = endPt+normalOnBInWorld*newDepth;
+			startPt = endPt - normalOnBInWorld*newDepth;
 		} else
 		{
 			endPt = pointInWorld + normalOnBInWorld*orgDepth;


erwincoumans:
Thanks for tracking this down! The convex perturbation is not enabled by default but can be useful indeed, to compute multiple contact point, even when the algorithm only computes one.

I ran your test and confirm the sign swap, so let's fix it.

~~~~~~~~~~~~
///Use this method to allow to generate multiple contact points between at once, between two objects using the generic convex-convex algorithm.
	///By default, this feature is disabled for best performance.
	///@param numPerturbationIterations controls the number of collision queries. Set it to zero to disable the feature.
	///@param minimumPointsPerturbationThreshold is the minimum number of points in the contact cache, above which the feature is disabled
	///3 is a good value for both params, if you want to enable the feature. This is because the default contact cache contains a maximum of 4 points, and one collision query at the unperturbed orientation is performed first.
	///See Bullet/Demos/CollisionDemo for an example how this feature gathers multiple points.
	///@todo we could add a per-object setting of those parameters, for level-of-detail collision detection.
	void	setConvexConvexMultipointIterations(int numPerturbationIterations=3, int minimumPointsPerturbationThreshold = 3);

	void	setPlaneConvexMultipointIterations(int numPerturbationIterations=3, int minimumPointsPerturbationThreshold = 3);

~~~~~~~~~~~~~~

radarsat1:
Indeed, I'm using it for surface-surface contact between convex hulls.  It works very well!  But I had been confused for quite some time about why my contact point were occasionally not on the surface of the objects, so finally decided to dive in.  Thanks for the merge!

PR #1565, opened 2018-02-16
Render options

BlGene:
This PR modifies `OpenGLExampleBrowser` to add the variables `window_backend` and `render_device`. The `window_backend` parameter is set to `X11` by default and can also be set to `EGL`. In case EGL is chosen  the `render_device` can be used to specify a GPU 

1. I'm using these changes from the pybullet interface and have modified `setup.py` accordingly. I have not updated the CMake build.
2. The PR overwrites those parts of the code that invoke the static/thrid_party Glew library because this was causing segfaults for EGL. 
3. This code invokes std libraries in a few files which don't otherwise use it.
4. `rendertest.py` is not cleaned up.


erwincoumans:
Thanks for the work on this, but mind that if it requires too many changes it is unlikely to be merged.
1) Before introducing EGL by default, it would be good to have a way to dynamically loading EGL functions, instead of linking against them. So we can select EGL at runtime, without failing on systems where EGL is missing. This is why we modified glew, and it is probably best to switch to GLAD instead of glew.

2) why do you introduce Docker?

BlGene:
Hi @erwincoumans, 

1. Some EGL function calls that I was using, for optional device discovery, were not available on the version of `<EGL/eglext.h>` provided by Ubuntu 14 and travis-ci only really supports Ubuntu 16 via docker. 

2. I reduced the commit size, amongst others by removing the `rendertest.py` file.

I've commented out device discovery, I would suggest removing it completely and adding it again once these functions can be loaded dynamically using GLAD, this also solves the docker problem. That would reduce the size of the commit to ~100 lines. Device discovery can be added in another commit if its still desired.

erwincoumans:
Moving to glad (https://github.com/Dav1dde/glad) is probably a good thing to do. Can glad support both GL and EGL at the same time, or do we need to separate 'glad' loaders in one executable? Perhaps you want to look into/experiment with glad too?

.travis.yml
```diff
@@ -14,7 +14,7 @@ addons:
 script:
   - echo "CXX="$CXX
   - echo "CC="$CC
-  - if [[ "$TRAVIS_OS_NAME" == "linux" && "$CXX" = "g++" ]]; then sudo apt-get install python3-pip; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" && "$CXX" = "g++" ]]; then sudo apt-get install libegl1-mesa-dev libglew-dev python3-pip; fi
```
BlGene: I'll remove this.

BlGene:
I would separate this into two separate PRs. This one adding render options and another changing the loaders. This is seems like a natural factoring because glad doesn't change the OpenGL/EGL behavior, it just changes loading.

In this case we would need to be careful since it means committing a bit of EGL code that is not covered by CI, as EGL is not included in the current CI. Making CI cover EGL can be done by another PR if needed. The docker option is a pain, the most efficient option might just be to wait till Travis enables Xenial support, which should happen soon. This would also mean that EGL support is not available without changing `setup.py` till this point, but that should be OK.

BlGene:
@erwincoumans  This has finally converged on a version that I am happy with. It should address all the points you mentioned, except for commit size, but this is due to glad.

erwincoumans:
Thanks for working on this! I'm in the process of moving from glew to glad, so there will be likely some conflicts, we can deal with them.
My pull request is here: https://github.com/bulletphysics/bullet3/pull/1569

BlGene:
@erwincoumans: rebased and squashed, ready for review.

examples/OpenGLWindow/SimpleOpenGL3App.h
```diff
@@ -16,7 +17,7 @@ struct SimpleOpenGL3App : public CommonGraphicsApp
 	class GLInstancingRenderer* m_instancingRenderer;
 	virtual void setBackgroundColor(float red, float green, float blue);
 
-	SimpleOpenGL3App(const char* title, int width,int height, bool allowRetina=true);
+	SimpleOpenGL3App(const char* title, int width, int height, bool allowRetina=true, std::string windowType="default", int renderDevice=-1);
```
erwincoumans: what is the string windowType doing here?

examples/OpenGLWindow/EGLOpenGLWindow.cpp
```diff
@@ -101,29 +103,41 @@ void EGLOpenGLWindow::createWindow(const b3gWindowConstructionInfo& ci) {
...
-    (PFNEGLGETPLATFORMDISPLAYEXTPROC)eglGetProcAddress(
-                                                       "eglGetPlatformDisplayEXT");
-    if (eglGetPlatformDisplayEXT == nullptr) m_data->egl_display = EGL_NO_DISPLAY;
-    
     const int max_devices = 32;
     EGLDeviceEXT egl_devices[max_devices];
     EGLint num_devices = 0;
     EGLint egl_error = eglGetError();
-    if (!eglQueryDevicesEXT(max_devices, egl_devices, &num_devices) ||
+    if (!glad_eglQueryDevicesEXT(max_devices, egl_devices, &num_devices) ||
```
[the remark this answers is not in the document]
BlGene: Fixed.

setup.py
```diff
@@ -378,22 +378,30 @@
...
     +["examples/ThirdPartyLibs/glad/glad.c"]\
     +["examples/ThirdPartyLibs/glad/glad_glx.c"]
     include_dirs += ["examples/ThirdPartyLibs/optionalX11"]
+    if 'BT_USE_EGL' in CXX_FLAGS:
+        # linking with bullet's Glew libraries causes segfault
+        # for some reason.
+        CXX_FLAGS += '-std=c++11 '
+        sources += ['examples/ThirdPartyLibs/glad/glad_egl.c']
+        sources += ['examples/OpenGLWindow/EGLOpenGLWindow.cpp']
+        libraries += ['EGL', 'GL']
```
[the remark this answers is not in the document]
BlGene: I expected this too, but for me they weren't. I've asked the question here: https://github.com/Dav1dde/glad/issues/130

examples/OpenGLWindow/SimpleOpenGL3App.h
BlGene: This allows runtime selection of X11 vs. EGL, which allows using the ExampleViewer without recompiling.

BlGene:
@erwincoumans is there anything else I should still do for this PR?

erwincoumans:
The pull request is not acceptable: some devs use the EGL in the cloud and require original egl API calls, so renaming to glad is not an option.
EGL needs to be dynamically loaded, not linked. If glad cannot do that, then we need to wait for another solution.
There are small other nitpicks (windowType needs to be an int etc), but above needs to be sorted out first. I suggest just using your own fork for now.

PR #1577, opened 2018-02-24
Nbelakovski/determinism changes

nbelakovski:
These are all the changes I found to be necessary in order for serialization/deserialization to be deterministic. There are essentially 3 changes:

1) Serialize/deserialize localtime for btDiscreteDynamicsWorld (commit e4d2591)
2) Deserialization of linear and angular velocity, and modification of the btWorldImporter's createRigidBody function to use a MotionState (commit 22ddde3)
3) Serializing ConvexInternalShapeData as double precision when double precision is being used. This is the rest of the commits, and involves most of the code changes since a lot of things include ConvexInternalShapeData, and so they had to be updated so that the DNA could be properly regenerated.

I strongly recommend reviewing this one commit at a time.

I tested this with some changes to the BasicExample target and was able to verify that the sim ran deterministically through serialization/deserialization, although I must admit I only tested with the box shape. These changes should apply to any shape though.

Mentioning @dbacchet to subscribe him.


nbelakovski:
The commits seem to have gotten out of order in the display here. See https://github.com/nbelakovski/bullet3/commits/nbelakovski/determinism_changes for the proper list, and I'd recommend starting with the time one since that's where it first diverges from the official repo.

nbelakovski:
Ping @erwincoumans . Any interest? If not I will close the PR, as it will only get stale.

erwincoumans:
At the moment I don't have time to review and merge it.
This is going to break the serialization backward compatibility, not too fond of that.
Also, did you test loading a double precision file in a single precision build of Bullet and vise versa?
You also need to re-create the autogenerated [bullet.h](https://github.com/bulletphysics/bullet3/blob/master/Extras/Serialize/BulletFileLoader/autogenerated/bullet.h)
I would not create btDefaultMotionState by default, it is easy to create it afterwards if you need it.

nbelakovski:
I'm a little confused, you said you don't have time to review it, but then you provided some comments. Did you mean to say that if I made those changes and did that testing that you might be able to find the time to review and merge, or am I misinterpreting? I can certainly do those things, but I do have to budget my own time as well, and I can only budget time for these things if we're on a path to getting this merged in some fashion, even if it's a subset of these changes.

Regarding the breaking backwards compatibility, I guess that comes with the splitting of the ConvexInternalShape into the float and double variants (and all of the downstream impacts of that of course)? Yeah, it's unfortunate. Personally, I might categorize that under "can't make an omelet without breaking a few eggs", since the breaking comes with the benefit of determinism for future work, but obviously that's your call to make :)

erwincoumans:
At the moment I don't have time to FULLY review/test and merge it.
If you address the points in my previous post, you can just leave the PR open:

1) remove the creation of btDefaultMotionState by default
2) test loading a double precision file in a single precision build of Bullet and vise versa (in particular due to [this change](https://github.com/bulletphysics/bullet3/pull/1577/commits/6f8091d9e71e3bb74d803efa1ab73d5824930145#diff-797c275fa95423e4c3b2daf25c1742a8L376))
3) re-create the autogenerated bullet.h (if that is complicated, I'll do it myself)

nbelakovski:
I'm afraid that if you or any other maintainer of Bullet aren't in a position to commit resources to this PR in a reasonable timetable then I can't commit resources to it either :(

I will leave the branch as it is in my repo.

PR #1579, opened 2018-02-27
Multithreaded constraint solver

lunkhound:
Adds a new class SequentialImpulseConstraintSolverMt, which is basically a multithreaded version of SequentialImpulseConstraintSolver.

Supports all of the features of the normal sequential impulse solver such as:
    - split penetration impulse
    - rolling friction
    - interleaving constraints
    - warmstarting contact points
    - 2 friction directions
    - randomized constraint ordering
    - early termination of solver iterations when leastSquaresResidualThreshold is satisfied

A few changes were made to SequentialImpulseConstaintSolver to reduce code duplication. These changes are purely cosmetic in nature, and basically consist of moving certain chunks of code into functions so they can be called by a derived class.

So how does this integrate with the existing multithreading which solves multiple islands in parallel?
In my tests, dispatching multithreaded solvers in parallel resulted in poor performance with the task schedulers I tested it with. The default task scheduler does not support nested parallel-for, but some of the others do -- and they all performed rather poorly.

I tried a different approach: send large islands to the multithreaded solver one at a time, and then afterwards send the smaller islands to single-threaded solvers in parallel.
An island is considered "large" if the number of manifolds it has exceeds a certain threshold value (which is user settable).

Additionally, the default task scheduler (based on Win32 threads or pthreads) has been moved into the LinearMath project. This should make it much easier to get multithreading working out of the box for end users. And since this adds built-in multithreading, the Cmake config variable for multithreading has been renamed from "BULLET2_USE_THREAD_LOCKS" to just "BULLET2_MULTITHREADING"

This PR should have negligible effect on single-threaded Bullet, which is still the default.


erwincoumans:
Thanks a lot for the contribution. Does it maintain backward compatibility for single-threaded behavior case?

erwincoumans:
Thanks fo the contribution! I need to see if the changes don't break our internal builds.

erwincoumans:
It looks good to me, thanks!

PR #1587, opened 2018-03-05
Fix import of spherical joint state

a-anik:
The positions of MultiBody's spherical joints do not restore correctly when saving/restoring state from C-API. It happens because spherical joints in btMultiBody have 4 positional variables. Although C-API and pybullet do not yet fully support spherical joints (there is no way to create them), the issue could be fixed beforehand.


erwincoumans:
Good point, Eventually it would be good to expose spherical joints (and planar) and add motors/actuators etc. For now, we don't have robots in our lab with spherical joints, and not many cycles to add the feature.

erwincoumans:
oddly there is some conflict in the branch. I manually applied it following the steps,
new PR is here: https://github.com/bulletphysics/bullet3/pull/1592

Issue #1582, opened 2018-02-27
In btGeneric6DofConstraint the maximum motor force is used as an impulse despite the name


GaborPuhr:
In btGeneric6DofConstraint one can set the maximum motor force with the btRotationalLimitMotor::m_maxMotorForce and the btTranslationalLimitMotor::m_maxMotorForce.
Despite the name this value will be used as an Impulse (and not as a Force).
This is confusing and also the user should know the deltatime to set a correct value and that is not always easy to obtain.

Checking other constraints:
In case of btHingeConstraint it is correctly called setMaxMotorImpulse.
In case of btSliderConstraint it is called setMaxLinMotorForce and it is correctly used as force.

I suggest to change the implementation similar to btSliderConstraint by simply dividing with the FPS.

(The problem exist also in btGeneric6DofSpringConstraint and btGeneric6DofSpring2Constraint)

erwincoumans:
Thanks for bringing this up, it is best to fix this indeed (dividing by FPS when using a force).

GaborPuhr:
I will create a pull request shortly.

PR #1588, opened 2018-03-05
Fix for #1582

GaborPuhr:
I have also changed the actual values where maxMotorForce was used. However in a few places it seemed there were only arbitrary big numbers. I left those as they were.
These are unchanged:
ImportMJCFSetup.cpp
c->setMaxMotorForce(jointInfo->m_jointAxisIndex,10000);
ImportURDFSetup.cpp
c->setMaxMotorForce(jointInfo->m_jointAxisIndex,10000);
Hinge2Vehicle.cpp
pHinge2->setMaxMotorForce(motorAxis,1000);


erwincoumans:
Thanks for the fix!

Issue #1567, opened 2018-02-17
Spring2Constraint gets the dynamics wrong!


trashpirate:
Hi! 

First of all, thanks for an awesome physics engine! I've been using Bullet for my research and noticed strange behavior of the btGeneric6Dof**Spring2**Constraint, which seemed to ignore an important principle of physics. I investigated this with a simple horizontal pendulum simulation: one dof (rotation about vertical axis, spring enabled, no gravity). I did several simulations with different pendulum length (distance from pivot point to mass). To my surprise, the frequency of oscillation did not change, no matter the length of the pendulum -> this is dynamically incorrect! The frequency should be inversely correlated with pendulum length. 
I simulated the same thing with the "depricated" btGeneric6Dof**Spring**Constraint and all of a sudden, without making any other changes, I could see the correct behavior: the frequency increased with decreasing pendulum length!

![constraints_frequency_length](https://user-images.githubusercontent.com/36479450/36342410-c2451dec-13c3-11e8-8657-b5d3ee1b7bb0.png)

If there are any parameters I don't know of that would make Spring2Constraint to behave correctly, I would love to know! Otherwise, it would be great if you could fix the issue in the Spring2Constraint since it effects the accuracy of the physics simulations significantly. To my knowledge, the issue only occurs when the spring is enabled - the frequency of a pendulum driven only by gravity changes with the length.

Thanks!

erwincoumans:
I haven't never used or looked into the spring parts of the btGeneric6DofSpring2Constraint (for our robotics / deep reinforcement learning work, we use btMultiBody, and we don't use springs).

Can you please create a small reproduction case, perhaps modifying some Bullet example, that  shows the issue?

trashpirate:
I created an example which you can clone from here: https://github.com/trashpirate/bullet_spring_constraint_demo/tree/master

Let me know if you have issues compiling it. Thanks!

erwincoumans:
Could you try attached fix? (unzip and copy over the old version)

[btGeneric6DofSpring2Constraint.cpp.zip](https://github.com/bulletphysics/bullet3/files/1769179/btGeneric6DofSpring2Constraint.cpp.zip)

trashpirate:
Sorry about the late response. I can confirm that the simulated frequencies now correspond with those of the btGeneric6DofSpringConstraint, thank you!

benelot:
Any reason why the better version is not merged into master? Or how come it
is packed in a zip instead?

[...]

PR #1589, opened 2018-03-05
Fix for #1567

GaborPuhr:


erwincoumans:
Great, thanks a lot Gabor! 

It fixes the repro case in #1567.

PR #1595, opened 2018-03-10
add pybullet tm700 robot example

YuehChuan:
Hi, add [techman robot](http://tm-robot.com/)  tmRobot.py  under examples/pybullet/examples
tm-sim


erwincoumans:
It looks nice. What is the license of those files?
If the size is big, it may be better to move it into the https://github.com/erwincoumans/pybullet_robots repo

YuehChuan:
It's Apache License 2.0. 
from [here](https://github.com/kentsai0319/techman_robot)
ok, it's big compare to iiwa(18mb vs 1.8mb), I'll put it somewhere mentioned above.
Thank you for reviewing.

PR #1597, opened 2018-03-12
Render options glad2

BlGene:
This is an updated version of #1565:

This pull request addressed all of the issues that were raised last time. To recap, these were:
1. Not renaming EGL API calls to glad_*.
2. Keeping EGL dynamically loaded. This is done by using glad2, which also changed some filenames. 
3. windowType is an int.
4. (Additionally) X11OpenGLWindow.cpp now specifically requests at least a GL 3.3 context as GLInstancingRenderer.cpp uses glVertexAttribDivisor. 

Again this PR is quite big, but mostly because of chaining to the generated glad files.

Edit: this contains a modified version of glad/glx that does dynamic linking of X11, if this is not needed it could be removed.


erwincoumans:
Thanks a lot for the contribution.
Why is there a CXX_FLAGS += '-std=c++11 '?

BlGene:
That's there because the EGL code uses `nullptr`s
e.g. :https://github.com/bulletphysics/bullet3/blob/master/examples/OpenGLWindow/EGLOpenGLWindow.cpp#L107

erwincoumans:
Let's remove that nullptr then.

BlGene:
OK, I'll do it now.

setup.py
```diff
@@ -22,7 +22,7 @@
 CXX_FLAGS += '-DBT_ENABLE_ENET '
 CXX_FLAGS += '-DBT_ENABLE_CLSOCKET '
 CXX_FLAGS += '-DB3_DUMP_PYTHON_VERSION '
-
+#CXX_FLAGS += '-DBT_USE_EGL '  # uncomment for EGL (old EGL versions fail)
```
BlGene: Should compilation with EGL be deactivated or activated by default?

erwincoumans:
Thanks a lot. The premake build is broken. Can you try running build_visual_studio_vr_pybullet_double.bat and open/compile the solution? Premake is not included in the continuous integration (I didn't figure out how to enable BOTH premake and cmake)

BlGene:
...I'm on Linux, the easiest way might be switching Appveyor to premake and debugging over CI.

erwincoumans:
Premake works on Linux very well too:
cd build3
./premake4_linux64 gmake
cd gmake
make -j12

BlGene:
Thanks for the tip, this should be fixed now, though I've only tested on Linux.

BlGene:
Rendering with EGL, connection mode GUI, is faster for a small number of parallel environments, rendering with Tiny connection mode DIRECT, is faster for a larger number of parallel environments. Environments were stepped synchronously. 

I don't think this occurs for non-synchronous stepping. I couldn't get good numbers for Tiny-GUI as I cannot run this on the server, it opens a window.

@erwincoumans: Do you think this is due to the connection mode or the rendering back end? what would you suggest doing in this case?

profile_sync
(total = average fps over envs * number of envs)

BlGene:
@erwincoumans is there already/will there be an OpenGL version of `tinyRendererPlugin`?

erwincoumans:
@BlGene It would be nice to have an EGL/OpenGL plugin indeed. I'm working on other things at the moment. 

I still haven't been able to run EGL on any of my desktops (Linux, Mac, Windows).
What operating system (+version) do you use, and what NVIDIA driver version? Is there some hello world example that should work with plain EGL?

BlGene:
@erwincoumans 
I would have expected at least Linux to work. I am using Ubuntu 16.04.3 LTS, with NVIDIA 384.90. I have a cpp-only EGL test script (link below) that should work, the compile command is in the first line, but it needs the new glad files, a copy of these could be found in my branch. https://gist.github.com/BlGene/2b0d865bf93e9adc88946c248f315dcc

This is my first very cursory first stab at an EGL/OpenGL plugin. https://github.com/BlGene/bullet3/tree/egl-plugin. I'm not very familiar with bullets internal interfaces, so any tips would be appreciated. Two question I have are:

1. Would I need one instance of `GLInstancingRenderer` per object or is one instance for all objects sufficient? https://github.com/BlGene/bullet3/blob/egl-plugin/examples/SharedMemory/plugins/eglPlugin/eglRendererVisualShapeConverter.cpp#L80
2. How can I make `B3_PROFILE` to work for python plugins.

My to-do list would look something like:
- [ ] get python plugins profiling to work (to help debug)
- [ ] fix setting camera params for EGL
- [ ] don't copy `convertURDFToVisualShape`  function
- [ ] revert `glx_dyn.c` -> `glx.c` to fix make scripts.
- [ ] ...

erwincoumans:
Great, after purging the nvidia drivers and re-installing 396 your example works. I create a small repo with the files for reference: https://github.com/erwincoumans/egl_example


You can also test the plugin(s) using C and C++ API, so without Python, which makes debugging easier. I prefer statically linking the plugin by default once finished (while still keeping the option to dynamically load the plugin).

Why do you need to revert the naming glx_dyn.c -> glx.c ?

erwincoumans:
Excuses for not applying this patch yet. 

Could you re-apply/update your EGL patch to latest Bullet trunk, only for Linux (it seems Windows and Mac build is broken, see All checks have failed). Preferably keep the file/line changes as needed low.

examples/StandaloneMain/hellovr_opengl_main.cpp
```diff
@@ -596,12 +596,12 @@ bool CMainApplication::BInit()
 {
 	b3Printf( "GL Error: %s\n", message );
 }
-*/
 
 static void APIENTRY DebugCallback (GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length, const GLchar* message, GLvoid* userParam)
```
BlGene: This was complaining about Gsizei. As far as I cant tell it shouldn't be.

https://gist.github.com/BlGene/f062ee5166ff0acb02a5403657d2c057

BlGene:
@erwincoumans:
> Great, after purging the nvidia drivers and re-installing 396 your example works. I create a small repo with the files for reference: https://github.com/erwincoumans/egl_example

Good to hear that you could get it to work.


> Why do you need to revert the naming glx_dyn.c -> glx.c ?

Just so that I don't have to change all build files.


> Could you re-apply/update your EGL patch to latest Bullet trunk, only for Linux (it seems Windows and Mac build is broken, see All checks have failed). 

Yep, done.


> Preferably keep the file/line changes as needed low.

If you don't coun't the glad files its quite short actually. :grimacing:

BlGene:
I've moved the OpenGL plug in stuff to #1790

PR #1599, opened 2018-03-13
fix a bug that createVisualShape(Array) does not have the correct col…

jietan:
…or. Add urdfEditor.py to pybullet_utils.

My emacs editor automatically trims heading/trailing spaces (to match Google coding style), which results in a lot of unnecessary diff's. If you do not like them, I can revert this commit and merge my changes using another editor.


erwincoumans:
Thanks for the contribution! The file PhysicsServerCommandProcessor.cpp includes a lot of white-space changes. Could you please revert those, so it only shows the actual changes in code?

PR #1604, opened 2018-03-15
Revised URDFEditor

jietan:
Fix a bug that createVisualShape(Array) does not have the correct color.
Revise URDFEditor so that it uses createVisualShapeArray.
Add URDFEditor to pybullet_utils.


erwincoumans:
LGTM, Thanks for the contribution.

PR #1607, opened 2018-03-19
fix premake build instructions in README

lunkhound:
When testing the premake option I added in my other PR (--enable-multithreading) I discovered that the command line given in the bullet README doesn't work as expected.  When I tried

`./premake4_linux64 gmake --double`

It silently ignored the "--double", however putting the "gmake" after the options worked as expected.
I only tested this on linux64.


erwincoumans:
Good point! The multi-threading PR takes a bit longer, need to review the effect on our internal code/unit tests.

PR #1609, opened 2018-03-19
Add PyBullet API to set projective texture.

YunfeiBai:


erwincoumans:
Thanks for the contribution.
It is better to avoid adding this GLInstancingRenderer dependency to PhysicsServerCommandProcessor.cpp
#include "../OpenGLWindow/GLInstancingRenderer.h"

Those globals shouldn't be in GLInstancingRenderer.cpp, please expose some APIs instead.

YunfeiBai:
Hi Erwin, I have exposed an API to set projective texture instead of using the global variable, which does not require GLInstancingRenderer dependency.

erwincoumans:

PR #1615, opened 2018-03-26
Fix jacobian computation with fixed joints

haudren:
This PR fixes the `calculateJacobian` function. It should take inputs of size `numDoF` instead of `numJoints`.

This is a quick patch, let me know which improvements are needed before you can merge.

### DIFF
--- a/examples/pybullet/examples/jacobian.py
+++ b/examples/pybullet/examples/jacobian.py
@@ -8,6 +8,15 @@ def getJointStates(robot):
 	joint_torques = [state[3] for state in joint_states]
 	return joint_positions, joint_velocities, joint_torques
 
+def getMotorJointStates(robot):
+	joint_states = p.getJointStates(robot, range(p.getNumJoints(robot)))
+	joint_infos = [p.getJointInfo(robot, i) for i in range(p.getNumJoints(robot))]
+	joint_states = [j for j, i in zip(joint_states, joint_infos) if i[3] > -1]
+	joint_positions = [state[0] for state in joint_states]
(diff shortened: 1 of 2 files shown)


erwincoumans:
Great, thanks for the contribution. I haven't used the calculateJacobian myself yet: we mainly use deep reinforcement learning to control our virtual and real robots, which directly outputs torque without using inverse dynamics and jacobians.

erwincoumans:
Looks good to me, thanks a lot for the contribution!

haudren:
Thanks for merging this :) If you'd like I can also modify `jacobian.py` to become a real test on multiple robots.

erwincoumans:
You are welcome to add a some tests, by default travis continuous integration runs
~~~~~~
python3 examples/pybullet/unittests/unittests.py
~~~~~~

Do you have more info on this classic control baseline? If you can create it using pybullet, please consider sharing the baseline.

haudren:
I opened a new PR for the Jacobian test, #1617.

For the baseline, I cannot share it in its current state, but the idea revolves around a reaching task:
- Control the robot by IK, in this case a good control is simply the object position.
- Control the robot directly by joint position, in this case a good control is the pseudo-inverse of J multiplied by the position error.
- Control the robot directly by joint torque, in this case a good control is given by some inverse dynamics. However, bullet does not provide the derivative of the jacobian, so the desired acceleration cannot be computed purely in bullet.

Inverse dynamics:
```
qdd_d = argmin_qdd || J qdd + Jd qd - Kp e - Kd ed ||
```

PR #1617, opened 2018-03-28
Add a few jacobian tests

haudren:
This makes sure that the result of `Jacobian*vel` is equal to the velocity computed by the simulator. I added a `numpy` dependency, I guess it can be done by hand but it is slightly cumbersome.

### DIFF
--- a/examples/pybullet/unittests/unittests.py
+++ b/examples/pybullet/unittests/unittests.py
@@ -2,6 +2,8 @@ import unittest
 import pybullet
 import time
 
+from utils import allclose, dot
+
 class TestPybulletMethods(unittest.TestCase):
 
     def test_import(self):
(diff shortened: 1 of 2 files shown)


erwincoumans:
Thanks for the test. It would be good to remove the numpy dependency, a dot product is just one while loop with mul and add operations.

haudren:
Ok, I removed the numpy dependency !

erwincoumans:
By the way, I noticed you experiment with the[ j2s7s300 robot](https://github.com/Kinovarobotics/kinova-ros/issues/126). 

Could you provide the URDF file, already preprocessed with xacro, so I can check it out with PyBullet? Eventually we should update the Gazebo integration of Bullet with btMultiBody functionality, it should provide stiffer robot arms than ODE and btRigidBody.

PR #1622, opened 2018-04-04
Fix to support different kind of wheel axis

juhl:
btRaycastVehicle::setCoordinateSystem() is not sufficient to support different coordinate system.

This PR fixes to support different coordinate system for vehicle wheels.


erwincoumans:
LGTM, Thanks for the contribution!

PR #1630, opened 2018-04-10
Update the projective texture shader to solve the texture interpolati…

YunfeiBai:
…on artifact.


erwincoumans:
LGTM, thanks for the fixes!

PR #1637, opened 2018-04-16
PyBullet: add signed distance field support, with example/signedDista…

erwincoumans:
…nceField.py

See also data\toys\concave_box.urdf and data\toys\concave_box.cdf

data\toys\concave_box.cdf was generated from concave_box.obj using
//GenerateSDF.exe -r "32 32 32" -d "-1.6 -1.6 -.6 1.6 1.6 .6" concave_box.obj
//SDF is based on code from DiscreGrid, https://github.com/InteractiveComputerGraphics/Discregrid


Levi-Armstrong:
@jmeyer

PR #1640, opened 2018-04-17
Fixing include line of btMultibodyWorldImporter.h to work with instal…

nbelakovski:
…led bullet files

When running make install for Bullet, no "Extras/" folder is created, so the include line as is will fail. This modification should work since the relevant header is in the same folder as this header.


erwincoumans:

PR #1647, opened 2018-04-24
BugFix for compile in ARM architecture

cdominguezm:
Have to be explicit cast for compile in ARM v7 (Raspberry pi)


madebr:
What is the compile error message you're fixing here?

PR #1648, opened 2018-04-25
make sure that the pre-trained galloping and trotting policies work f…

jietan:
…or the minitaur_reactive_env and minitaur_trotting_env environments.


erwincoumans:

PR #1651, opened 2018-04-25
Expose more physics engine parameters

mbreyer:
Exposed more parameters of the engine in pybullet.

While reading the values, I noticed that the default value for contactERP (m_erp2) of 0.2 defined in [btContactSolverInfo](https://github.com/bulletphysics/bullet3/blob/master/src/BulletDynamics/ConstraintSolver/btContactSolverInfo.h#L85) gets overwritten to 0.08 inside the creation of a new world in [PhysicsServerCommandProcessor](https://github.com/bulletphysics/bullet3/blob/master/examples/SharedMemory/PhysicsServerCommandProcessor.cpp#L2296). I was just wondering if there was any particular reason for the choice of the lower value?

### DIFF
--- a/examples/pybullet/pybullet.c
+++ b/examples/pybullet/pybullet.c
@@ -1069,23 +1069,31 @@ static PyObject* pybullet_getPhysicsEngineParameters(PyObject* self, PyObject* a
 
 		statusHandle = b3SubmitClientCommandAndWaitStatus(sm, command);
 		statusType = b3GetStatusType(statusHandle);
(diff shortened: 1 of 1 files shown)


erwincoumans:
Ah, good point, expose them to the PyBullet.getPhysicsEngineParameters API.

Our error reduction parameter is related to the number of solver iterations and timestep. The default erp (0.2) is suitable for 60Hz, 10 iterations. For robotics, our default is 240Hz and 50 iterations, so I picked a more suitable erp default of 0.08. It is

mbreyer:
I updated the docstring to match the new function signature. If this gets merged into master, I can also adapt the google doc appropriately.

PR #1658, opened 2018-04-30
Expand aabb with CollisionObject getContactProcessingThreshold()

Levi-Armstrong:
I was working on a way to get distance information using the broadphase collision checking. I noticed that even if you set the contactProcessingThreshold for the collision object to a positive distance it would not return contact information for objects within this distance. These are the changes that were the changes required to get contact information for objects within the contactProcesssingThreshold. I also had to make a change to the ConvexConcaveCollisionAlgorithm to get correct information and if this is the correct approach this change will most likely need to be made to some of the other algorithms.

I wanted to get this PR out early to get the maintainer engage early in case I am headed down the wrong path.

### DIFF
--- a/src/BulletCollision/CollisionDispatch/btCollisionWorld.cpp
+++ b/src/BulletCollision/CollisionDispatch/btCollisionWorld.cpp
@@ -128,6 +128,11 @@ void	btCollisionWorld::addCollisionObject(btCollisionObject* collisionObject, in
 	btVector3	maxAabb;
 	collisionObject->getCollisionShape()->getAabb(trans,minAabb,maxAabb);
 
+	//need to increase the aabb for contact thresholds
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
@erwincoumans is this the correct approach?

erwincoumans:
Yes, this goes in the right direction indeed. Unfortunately this breaks a lot of internal tests here (they rely on deterministic physics, and even small changes can add up).

Levi-Armstrong:
Thank you for the feedback. I noticed that the test below failed on travis-ci but and I did see several failures on appveyor but I am new to Bullet and not sure where to start on resolving the errors shown appveyor. Can you provide any guidance?

FAIL: test_rolling_friction (__main__.TestPybulletMethods)

PR #1662, opened 2018-05-02
Fixed a collision check bug with CCD

### LINKED ISSUE #691: setIgnoreCollisionCheck and CCD
Seems like setIgnoreCollisionCheck doesn't work with CCD. The rigid bodies move through each other when they go slower, but when there's enough motion they collide. Is this a bug or am I just doing something silly?

sponji:
This should fix #691


erwincoumans:
Thanks, indeed the filter should be applied for CCD as well.

PR #1663, opened 2018-05-02
Fix some Doxygen comments in btMatrix3x3.

vsaulue:


erwincoumans:
Ah, thanks for the contribution!

Issue #1673, opened 2018-05-05
Crash for btCollisionObject in dynamic world when multi-threading is enabled


przemir:
1. My program using single-threaded btDiscreteDynamicsWorld crashed because of btSequentialImpulseConstraintSolver::getOrInitSolverBody method returned -1
It happened because of line:

`if ( btRigidBody* rb = btRigidBody::upcast( &body ) )`

I use CollisionObjects aside from RigidBodies. After I commented line above, single-threaded simulation started to work again.
Similar line is in btSequentialImpulseConstraintSolverMt::getOrInitSolverBodyThreadsafe though.

2. After configuring btDiscreteDynamicsWorldMt with specified btSequentialImpulseConstraintSolverMt program crashes as well.
For btDiscreteDynamicsWorldMt without btSequentialImpulseConstraintSolverMt it works.
I guess (like in the case above) it may have problems when there are pure CollisionObjects inside dynamic world aside from RigidBodies.
To reproduce problem with a small amount of code, I modified sample 'MultiThreadedDemo' a bit, i.e replaced fragment of method MultiThreadedDemo::createStack

from: 
```
				btRigidBody* body = localCreateRigidBody( mass, trans, boxShape );
				body->setFriction( 1.0f );
				body->setRollingFriction( gSliderRollingFriction );
```

to: 
```
				if(iX % 2 == 0) {
					btRigidBody* body = localCreateRigidBody( mass, trans, boxShape );
					body->setFriction( 1.0f );
					body->setRollingFriction( gSliderRollingFriction );
				}
				else {
					btCollisionObject* collisionObject = new btCollisionObject();
					collisionObject->setWorldTransform(trans);
					collisionObject->setCollisionShape(boxShape);
					m_dynamicsWorld->addCollisionObject(collisionObject);
				}
```

Now it causes crash.
There is btAssert: btBatchedConstraints.cpp:215 (conInfo.numConstraintRows <= 6)

* CMAKE configuration: 32-bit, Debug

lunkhound:
That assert in btBatchedConstraints should be removed. I'll make a PR.

lunkhound:
With the bad assert removed, I'm not seeing any crash related to btSequentialImpulseConstraintSolver::getOrInitSolverBody using the reproduction steps you provided.
Do you have repro steps for that crash?

przemir:
With bad assert removed demo works.

[...]

PR #1674, opened 2018-05-07
parallel solver: remove bad assert

lunkhound:
This assert serves no useful purpose. It will cause an erroneous assert when there are more than 6 points of contact between a pair of bodies.
See issue #1673


erwincoumans:
Thanks for the contribution. Do you mind looking into the btCollisionObject related issue(s) in btSequentialImpulseConstraintSolver and Mt version?

https://github.com/bulletphysics/bullet3/issues/1673

lunkhound:
I will fix it if I can get steps to reproduce it.
The bad assert was the only issue I found from the given repro steps. I'll inquire about getting a repro case for the other problems that were mentioned.

PR #1677, opened 2018-05-08
remove explicit cast from float to double

AndrewMeadows:
- remove explicit cast from float to double


AndrewMeadows:
Bah.  Don't merge this yet.  I optimistically thought I had solved the problem but it I'm not done yet.

AndrewMeadows:
Nevermind.  Our problem appears to be caused by VisualStudio being stupid.  We'll workaround it for now.

PR #1682, opened 2018-05-12
constraint solvers: fix crash for collision-bodies with incorrect flags

lunkhound:
This change only affects multithreaded bullet. Fixes issue #1673 which occurs when using non-rigid body collision objects with neither CF_STATIC_OBJECT nor CF_KINEMATIC_OBJECT flag set.

With this change, all non-rigid-body collision objects are treated as static, regardless of flags (instead of crashing).

### DIFF
--- a/src/BulletDynamics/ConstraintSolver/btSequentialImpulseConstraintSolver.cpp
+++ b/src/BulletDynamics/ConstraintSolver/btSequentialImpulseConstraintSolver.cpp
@@ -780,6 +780,8 @@ int	btSequentialImpulseConstraintSolver::getOrInitSolverBody(btCollisionObject& 
     }
     else
     {
+        // Incorrectly set collision object flags can degrade performance in various ways.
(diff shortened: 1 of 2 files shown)


lunkhound:
I tested these changes with MSVC 2015 x64 in Debug and RelWithDebInfo configurations (with the multithreading compile option on).
For testing purposes I modified the MultiThreadedDemo::createStack() function so that half of the boxes created are static:
```
                btRigidBody* body = localCreateRigidBody( mass, trans, boxShape );
                if ( ( iX & 1 ) == 0 )
                {
                    body->setMassProps(0.0, btVector3(0,0,0));
                    body->setCollisionFlags(0);  // incorrect flags
                }
```

I was able to verify that:
 - in a Debug build with incorrect flags the expected asserts fire
 - in a Debug build with correct flags, no asserts fire
 - in a RelWithDebInfo build with incorrect or correct flags, it does not crash

erwincoumans:
Thanks for the fixes!

PR #1686, opened 2018-05-16
Expose CFM parameters

yuchenericwu2:


erwincoumans:
Thanks for the PR. 

We moved to the generic 'setPhysicsEngineParameter' and 'getPhysicsEngineParameter' instead of individual functions in the PyBullet API, and those parameters are already exposed in the 'setPhysicsEngineParameter' API, see the [PyBullet Quickstart Guide](https://docs.google.com/document/d/10sXEhzFRSnvFcl3XxNGhnD4N2SedqwdAvK3dsihxVUA/edit#) in that section.


optional | erp | float | constraint error reduction parameter (non-contact, non-friction)
-- | -- | -- | --
optional | contactERP | float | contact error reduction parameter
optional | frictionERP | float | friction error reduction parameter (when positional friction anchors are enabled)

So let's not blow up the number of separate APIs.

PR #1699, opened 2018-05-22
Fix pendulum & manipulator envs with pybullet client

benelot:
You forgot to add the pybullet client to the pendulum environments. Thus they were no longer functional.


benelot:
Awkwardly, there seem to be more commits on the list, but as you can see, I only changed some lines in 3 files, that is all.

benelot:
There is the same problem with the manipulators too. That is why I appended the fix here too.

benelot:
There are some other problems with all manipulators except the reacher. I will look into that soon.

erwincoumans:
Thanks, I'll trust your PR improves the Gym environments.

Issue #1691, opened 2018-05-20
Failed to init some gym environments


ShangtongZhang:
```
    import pybullet_envs
    game = 'ReacherBulletEnv-v0'
    game = 'ThrowerBulletEnv-v0'
    game = 'HalfCheetahBulletEnv-v0'
    env = gym.make(game)
    env.reset()
```

```
Traceback (most recent call last):
  File "/opt/project/examples.py", line 449, in <module>
    env.reset()
  File "/usr/local/lib/python3.5/dist-packages/gym/wrappers/time_limit.py", line 44, in reset
    return self.env.reset()
  File "/usr/local/lib/python3.5/dist-packages/pybullet_envs/gym_locomotion_envs.py", line 28, in _reset
    r = MJCFBaseBulletEnv._reset(self)
  File "/usr/local/lib/python3.5/dist-packages/pybullet_envs/env_bases.py", line 67, in _reset
    s = self.robot.reset(self._p)
  File "/usr/local/lib/python3.5/dist-packages/pybullet_envs/robot_bases.py", line 124, in reset
    self.robot_specific_reset(self._p)
TypeError: robot_specific_reset() takes 1 positional argument but 2 were given
```

```
root@0afadd8d2e39:/workspace# pip freeze | grep 'bullet'
pybullet==1.9.8
```

Some environments work well (e.g., Ant) but some failed. Maybe it has sth to do with https://github.com/bulletphysics/bullet3/pull/1690

erwincoumans:
It is fixed in the pull request: https://github.com/bulletphysics/bullet3/pull/1692
will upload pybullet 1.9.9 soon
Thanks for the report!

ShangtongZhang:
Thanks for your quick response, however some games are still wrong in the new version.
```
    import pybullet_envs
    games = [spec.id for spec in gym.envs.registry.all() if spec.id.find('Bullet') >= 0]
    bugs = []
    for game in games:
        try:
            env = gym.make(game)
            env.reset()
            env.step(np.random.random(env.action_space.shape))
        except Exception as e:
            print(e)
            bugs.append(game)
    print(bugs)
```
Most are
```
episode_restart() takes 1 positional argument but 2 were given
```

erwincoumans:
More fixes in this PR: https://github.com/bulletphysics/bullet3/pull/1698
Thanks for the report!

PR #1702, opened 2018-05-23
Make a basic sanity check out of #1691. Could serve as unit test.

benelot:
I found the code in #1691 a pretty neat sanity check for the envs. Could serve as a unit test, but it depends on how much you want to keep the envs sane ;-)

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/gym_sanity_check.py
+++ b/examples/pybullet/gym/pybullet_envs/gym_sanity_check.py
@@ -1,0 +1,18 @@
+import gym
+import numpy as np
+import pybullet_envs
+envs = [spec.id for spec in gym.envs.registry.all() if spec.id.find('Bullet') >= 0]
+bugged_envs = []
+for env_name in envs:
+    try:
+        print('[TESTING] ENV', env_name, '...')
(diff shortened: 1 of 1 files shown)


examples/pybullet/gym/pybullet_envs/gym_sanity_check.py
```diff
@@ -0,0 +1,18 @@
...
+import numpy as np
+import pybullet_envs
+envs = [spec.id for spec in gym.envs.registry.all() if spec.id.find('Bullet') >= 0]
+bugged_envs = []
+for env_name in envs:
+    try:
+        print('[TESTING] ENV', env_name, '...')
+        env = gym.make(env_name)
+        env.reset()
+        env.step(np.random.random(env.action_space.shape))
```
erwincoumans: Better to use env.action_space.sample()
See also https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/examples/testEnv.py
~~~~~~~~~~~~
sample = env.action_space.sample()
action = sample*0.0
obs,rewards,done,_ =env.step(action)
~~~~~~~

benelot:
Sounds good, will do.

PR #1730, opened 2018-06-04
Adds unit tests for userData plus some bugfixes.

RanTig:
- Clears the user data cache when syncUserData is called.
- Adds unit tests for the UserData functons.
- Changes the char pointer in btHashString to std::string. There were problems where the object owning the string memory would deallocate the string, making the btHashString object invalid.

### DIFF
--- a/examples/pybullet/unittests/userDataTest.py
+++ b/examples/pybullet/unittests/userDataTest.py
@@ -1,141 +1,147 @@
 import unittest
 import pybullet
 
+from pybullet_utils import bullet_client
+
 PLANE_PATH = "plane.urdf"
 ROBOT_PATH = "r2d2.urdf"
 
+
 class TestUserDataMethods(unittest.TestCase):
 
+	@classmethod
+	def setUpClass(cls):
+		TestUserDataMethods.server = bullet_client.BulletClient(connection_mode=pybullet.GUI_SERVER)
+
+	@classmethod
+	def tearDownClass(cls):
+		del TestUserDataMethods.server
+
 	def setUp(self):
-		self.client_id = pybullet.connect(pybullet.SHARED_MEMORY)
+		self.client = bullet_client.BulletClient(pybullet.SHARED_MEMORY)
 
 	def tearDown(self):
-		pybullet.resetSimulation()
-		pybullet.disconnect()
+		self.client.resetSimulation()
+		del self.client
 
 	def testAddUserData(self):
-		plane_id = pybullet.loadURDF(PLANE_PATH)
-		uid1 = pybullet.addUserData(plane_id, 0, "MyKey1", "MyValue1")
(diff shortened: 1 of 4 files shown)


RanTig:
The problem occurred when I stored a std::string in a btAlignedObjectArray, then pointed the btHashString pointer to the c_str() pointer of the std::string. the btAlignedObjectArray copies over the elements when it resizes. Usually a string does a shallow copy (retaining the underlying buffer) if it can, but I guess sometimes it does a deep copy, de-allocating the old underlying buffer.

If the idea of letting btHashString owning the memory sounds OK to you, how about using strdup() instead?

examples/pybullet/unittests/userDataTest.py
```diff
@@ -0,0 +1,237 @@
+import unittest
+import pybullet
+
+PLANE_PATH = "plane.urdf"
+ROBOT_PATH = "r2d2.urdf"
+
+class TestUserDataMethods(unittest.TestCase):
+
+	def setUp(self):
+		self.client_id = pybullet.connect(pybullet.SHARED_MEMORY)
```
erwincoumans: You cannot connect to SHARED_MEMORY without starting a server. This is done in the userData.py example, https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/examples/userData.py Can you check it out?

RanTig: Thanks, starts a server now. Also looks a bit cleaner using the BulletClient helper class, especially the multiclient test case.
Is there a way to start a server without a GUI?

erwincoumans: Yes, you can use SHARED_MEMORY_SERVER for a non-gui version.

examples/pybullet/unittests/userDataTest.py
```diff
@@ -0,0 +1,239 @@
...
+
+PLANE_PATH = "plane.urdf"
+ROBOT_PATH = "r2d2.urdf"
+
+
+class TestUserDataMethods(unittest.TestCase):
+
+	@classmethod
+	def setUpClass(cls):
+		TestUserDataMethods.server = bullet_client.BulletClient(connection_mode=pybullet.GUI_SERVER)
```
erwincoumans: SHARED_MEMORY_SERVER to allow running with GUI/headless

RanTig: Done.

erwincoumans:
LGTM. The main todo is to make userDataId globally unique, and simplify the API to
getUserData(dataUid) and removeUserData(dataUid), with optionally a user-friendly version getUserData(obUid, linkIndex, keyStr) (using the conversion from (ObjUid, linkIndex, keyStr)->dataUid under the hood.

PR #1741, opened 2018-06-09
removed needsCollision calls from collision algorithms constructors. …

LordOfDragons:
…some contain this

call into the dispatcher which causes problems if the collision algorithm is used for
intersection collision testing (for example btCollisionWorld::contactPairTest).

Bug leading to this fix commit is the following situation:

ObjA: Touch Sensor with GhostObject and CompoundShape with one BoxShape
ObjB: Actor with RigidBody and CompoundShape with one BoxShape.
Callsing btCollisionWorld::contactPairTest() with these two objects.

This chooses in the end the BoxBox collision algorithm to test the two boxes. Without this fix the collision algorithm tries to call btDispatch::needsCollision() inside the algorithm contructor which produces "false". This in turn causes no m_manifoldPtr to be created and thus the collision test always fails, which is the wrong result.

The Sphere-Sphere collision algorithm does not use needsResponse thus it is a bug. With this fix the problem vanishes.


erwincoumans:
No plan to change the API, it is best to keep the changes in your own fork.

PR #1759, opened 2018-06-14
Forced Exploration

hadelin2p:
We force the exploration by setting done to False during the whole episode.

Forced exploration allows to give more information to the AI, which leads to finding the best muscle impulsions / actions even after it fails within one same full episode.

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/gym_locomotion_envs.py
+++ b/examples/pybullet/gym/pybullet_envs/gym_locomotion_envs.py
@@ -61,7 +61,7 @@ class WalkerBaseBulletEnv(MJCFBaseBulletEnv):
 		state = self.robot.calc_state()  # also calculates self.joints_at_limit
 
 		alive = float(self.robot.alive_bonus(state[0]+self.robot.initial_z, self.robot.body_rpy[1]))   # state[0] is body height above ground, body_rpy[1] is pitch
-		done = alive < 0
+		done = False
 		if not np.isfinite(state).all():
 			print("~INF~", state)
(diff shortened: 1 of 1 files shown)


benelot:
That does not sound right. Especially because you do not add it as a flag,
but you just break the code.

erwincoumans:
Thanks for the contribution! For Half-Cheetah, MuJoCo gym also uses done=False, and I'd like to use that too in ours. 

For the other envs (Ant, Humanoid etc), ideally we can use the original MuJoCo versions of 'done', would that work?

https://github.com/openai/gym/tree/master/gym/envs/mujoco

I want those changes by default, flags don't work well in Gym (but you could add some optional flag to use roboschool behavior of 'done', but default should be more like MuJoCo)

hadelin2p:
Glad to help!

Indeed, it would be amazing to have the original MuJoCo versions of 'done' for the other environments, especially for Ant and Humanoid where I am still observing some room for improvement. That would definitely work with ARS, as it did with this Mujoco-like update of 'done' in pybullet_envs for Half-Cheetah.

erwincoumans:
The other pull request replaces this, right?

benelot:
Hi Hadelin2p

It seems that you have some training resources at hand. Do you have
pretrained agents for several of the OpenAI mujoco gyms at hand by chance?
Mine take very long to finish training to a decent performance level as I
am just training on my laptop with a mediocre dedicated GPU. And until
then, I can not tune my pybullet mujoco-env implementations to perform
similar to mujoco.

PR #1760, opened 2018-06-15
 Adds multithreading support for batch ray casts.

RanTig:
To enable the feature, enable the BULLET2_MULTITHREADING option or use the flag:
-DBT_THREADSAFE=1.

- Increases the number of rays that can go in a batch request by storing
them in the shared memory stream instead of the shared memory command.
- Adds the API b3RaycastBatchSetNumThreads to specify the number of
threads to use for the raycast batch, also adds the argument numThreads
to the pybullet function rayTestBatch.
- Rays are distributed among the threads in a greedy fashion there's a shared
queue of work, once a thread finishes its task, it picks the next
available ray from the task. This works better than pre-distributing the
rays among threads, since there's a large variance in computation time per ray.

Some controversial changes:
- Added a pointer to PhysicsClient to the SharedMemoryCommand struct, this
was necessary to keep the C-API the same for b3RaycastBatchAddRay, while
adding the ray to the shared memory stream instead of the command
struct. I think this may be useful to simplify other APIs as well, that
take both a client handle and a command handle.
- Moved #define SHARED_MEMORY_MAX_STREAM_CHUNK_SIZE to
SharedMemoryPublic. This was necessary for the definition of
MAX_RAY_INTERSECTION_BATCH_SIZE.

### DIFF
--- a/examples/SharedMemory/PhysicsClient.h
+++ b/examples/SharedMemory/PhysicsClient.h
@@ -43,6 +43,7 @@ public:
     virtual void setSharedMemoryKey(int key) = 0;
 
     virtual void uploadBulletFileToSharedMemory(const char* data, int len) = 0;
+    virtual char* getSharedMemoryStreamBuffer() = 0;
 
     virtual int getNumDebugLines() const = 0;
 
--- a/examples/SharedMemory/PhysicsClientC_API.cpp
+++ b/examples/SharedMemory/PhysicsClientC_API.cpp
@@ -2729,12 +2729,14 @@ B3_SHARED_API	b3SharedMemoryCommandHandle b3CreateRaycastCommandInit(b3PhysicsCl
     b3Assert(command);
     command->m_type = CMD_REQUEST_RAY_CAST_INTERSECTIONS;
 	command->m_requestRaycastIntersections.m_numRays = 1;
-	command->m_requestRaycastIntersections.m_rayFromPositions[0][0] = rayFromWorldX;
-	command->m_requestRaycastIntersections.m_rayFromPositions[0][1] = rayFromWorldY;
-	command->m_requestRaycastIntersections.m_rayFromPositions[0][2] = rayFromWorldZ;
-	command->m_requestRaycastIntersections.m_rayToPositions[0][0] = rayToWorldX;
-	command->m_requestRaycastIntersections.m_rayToPositions[0][1] = rayToWorldY;
-	command->m_requestRaycastIntersections.m_rayToPositions[0][2] = rayToWorldZ;
+	command->m_requestRaycastIntersections.m_numThreads = 1;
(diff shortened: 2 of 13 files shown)


erwincoumans:
Great, thanks for the contribution! I've tried it out and it works well.
Please fix the controversial changes: we don't need to maximize the #MAX_RAY_INTERSECTION_BATCH_SIZE to maximize the SHARED_MEMORY_MAX_STREAM_CHUNK_SIZE. How many rays do we want at most in batch? 1024? 10k? Also, please implement the upload/copy from rays to shared memory inside the client code, similar to uploadBulletFileToSharedMemory (or just use that function).

erwincoumans:
If you don't mind, I can also make those changes, right after merging your pull request, that saves roundtrip time. In fact, I already made them, together with some other small fixes.

erwincoumans:
screenshot of my simple Python benchmark, running from 1 to 8 threads:

image

~~~~~~~~~~~~~
import pybullet as p
import time
import math

p.connect(p.DIRECT)#GUI)

p.configureDebugVisualizer(p.COV_ENABLE_RENDERING,0)
p.loadURDF("samurai.urdf")
p.loadURDF("r2d2.urdf",[3,3,1])

rayFrom=[]
rayTo=[]
numRays = 2048
rayLen = 13

for i in range (numRays):
	rayFrom.append([0,0,1])
	rayTo.append([rayLen*math.sin(2.*math.pi*float(i)/numRays), rayLen*math.cos(2.*math.pi*float(i)/numRays),1])

timingLog = p.startStateLogging(p.STATE_LOGGING_PROFILE_TIMINGS,"rayCastBench.json")
for i in range (10):
	p.stepSimulation()
	for j in range (8):
		results = p.rayTestBatch(rayFrom,rayTo,j+1)
	
	rayHitColor = [1,0,0]
	rayMissColor = [0,1,0]
	#for i in range (numRays):
	#	if (results[i][0]<0):
	#		p.addUserDebugLine(rayFrom[i],rayTo[i], rayMissColor)
	#	else:
	#		p.addUserDebugLine(rayFrom[i],rayTo[i], rayHitColor)
		
	#time.sleep(1./240.)
p.stopStateLogging(timingLog)
~~~~~~~~~~~~~

erwincoumans:
I'm going to merge it and make some changes, to avoid slow round-trip time. Let's discuss next week about details.

erwincoumans:
This pull request undoes the "Added a pointer to PhysicsClient " / exposing SHARED_MEMORY_MAX_STREAM_CHUNK_SIZE (they should be internal details).https://github.com/bulletphysics/bullet3/pull/1761

erwincoumans:
The TaskScheduler was allocated but never freed, leading to large memory leaks. Fixed in 
https://github.com/bulletphysics/bullet3/pull/1761

erwincoumans:
Follow up to achieve proper API backward compatibility: https://github.com/bulletphysics/bullet3/pull/1765

PR #1762, opened 2018-06-15
isAlive method forcing done=False for halfcheetah

alexis-jacq:
As suggested in https://github.com/bulletphysics/bullet3/pull/1759. 

I had to pass the 'alive' condition as an additive parameter of WalkerBaseBulletEnv.

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/gym_locomotion_envs.py
+++ b/examples/pybullet/gym/pybullet_envs/gym_locomotion_envs.py
@@ -14,7 +14,6 @@ class WalkerBaseBulletEnv(MJCFBaseBulletEnv):
 		self.walk_target_x = 1e3  # kilometer away
 		self.walk_target_y = 0
 		self.stateId=-1
-		self.alive = None
 		
 
 	def create_single_player_scene(self, bullet_client):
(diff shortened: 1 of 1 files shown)


erwincoumans:
Is only Half-Cheetah affected? Do the other environments (Humanoid, Ant etc) train with with ARS with the original PyBullet reward/done?

alexis-jacq:
Half-Cheetah is the only one that overwrites the method, so it should be the only one affected. 
With ARS, Half-Cheetah seems ok, Humanoid and Ant are longer to train. I tell you asap if they worked.

hadelin2p:
We need to have three different 'done' for Half-Cheetah, Ant and Humanoid, if we want to use the same as in Mujoco (which I think we should since replacing 'done = alive < 0' by 'done = False' made Half-Cheetah work successfully). The 'done' in MuJoCo are as follows:

**Half-Cheetah:**
`done = False`

**Ant:**
```
notdone = np.isfinite(state).all() and state[2] >= 0.2 and state[2] <= 1.0
done = not notdone
```

**Humanoid:**
```
qpos = self.sim.data.qpos
done = bool((qpos[2] < 1.0) or (qpos[2] > 2.0))
```

erwincoumans:
can you implement those?

hadelin2p:
Sure we will do that this evening

erwincoumans:
isDone method is more clear than 'isAlive', since it is passed through Gym interface as 'done' boolean.
Thanks for the help!

erwincoumans:
Thanks, LGTM.

PR #1763, opened 2018-06-16
Add the pdControlPlugin to the joint control C API, and add the PD co…

YunfeiBai:
…ntrol mode (also available in pybullet). Modify the pdControl pybullet example to use the PD control mode with setJointMotorControl API.


erwincoumans:
Great, thanks for the contribution. Would it be possible to remove/disable the pdController if maxForce = 0?

YunfeiBai:
I can add that to the API. Right now, if the maxForce=0, the controller is not removed, but it has no effect.

PR #1767, opened 2018-06-17
Increase number of maximum dofs in shared memory.

bingjeff:
Recently started working with a model that has 132 dofs. Most are
fixed; however, using getJointState fails if the max num of dofs
is not large enough.


erwincoumans:
test

PR #1769, opened 2018-06-19
Several fixes for the parallel raycasts

RanTig:
- Limits the maximum number of threads to 64, since btThreadSupportPosix
and btThreadsupportWin32 don't support more than 64 bits at this moment,
due to the use of UINT64 bitmasks. This could be fixed by using
std::bitset or some other alternative.
- Introduces a threadpool class, b3ThreadPool, which is a simple wrapper
around btThreadSupportInterface and uses this instead of the global task
scheduler for parallel raycasting. This is actually quite a bit faster
than the task scheduler (~10-15% in my tests for parallel raycasts),
since the advanced features (parallelFor) are not necessary for the
parallel raycasts.
- Puts 16*1024 of MAX_RAY_INTERSECTION_MAX_SIZE_STREAMING in
parentheses, since it otherwise causes problems with other operators
of equal precedence and introduces a smaller constant for Apple targets.
- Refactors the parallel raycasts code and adds some more profiling.


erwincoumans:
Thanks for the contribution, I'll check it out (close/reopen to trigger rebuild, the appveyor build sometimes fails)

examples/SharedMemory/PhysicsServerCommandProcessor.cpp
```diff
@@ -109,6 +110,47 @@ struct UrdfLinkNameMapUtil
...
+	}
+
+	~b3ThreadPool() {
+		delete m_threadSupportInterface;
+	}
+
+	const int numWorkers() const { return m_threadSupportInterface->getNumWorkerThreads(); }
+
+	void runTask(int threadIdx, btThreadSupportInterface::ThreadFunc func, void *arg) {
+		FunctionContext ctx = m_functionContexts[threadIdx];
```
erwincoumans: ctx is a local copy of the m_functionContexts[.], and this local variable goes out-of-scope while the threads are using it.
Please use a reference instead:
~~~~~~~~~~~~
	FunctionContext& ctx = m_functionContexts[threadIdx];
~~~~~~~~~~~~

examples/SharedMemory/PhysicsServerCommandProcessor.cpp
```diff
@@ -4716,16 +4758,17 @@ struct CastSyncInfo {
...
 	const btCollisionWorld *m_world;
 	const b3RayData *m_rayInputBuffer;
 	b3RayHitInfo *m_hitInfoOutputBuffer;
 	int m_numRays;
 
-	BatchRayCaster(const btCollisionWorld* world, const b3RayData *rayInputBuffer, b3RayHitInfo *hitInfoOutputBuffer, int numRays)
-		: m_world(world), m_rayInputBuffer(rayInputBuffer), m_hitInfoOutputBuffer(hitInfoOutputBuffer), m_numRays(numRays) {
+	BatchRayCaster(b3ThreadPool *threadPool, const btCollisionWorld* world, const b3RayData *rayInputBuffer, b3RayHitInfo *hitInfoOutputBuffer, int numRays)
+		: m_threadPool(threadPool), m_world(world), m_rayInputBuffer(rayInputBuffer), m_hitInfoOutputBuffer(hitInfoOutputBuffer), m_numRays(numRays) {
 		m_syncInfo = new CastSyncInfo;
```
erwincoumans: Where is this deleted?

examples/SharedMemory/PhysicsServerCommandProcessor.cpp
```diff
@@ -4832,9 +4895,17 @@ bool PhysicsServerCommandProcessor::processRequestRaycastIntersectionsCommand(co
 
 	const int numCommandRays = clientCmd.m_requestRaycastIntersections.m_numCommandRays;
 	const int numStreamingRays = clientCmd.m_requestRaycastIntersections.m_numStreamingRays;
-	const int numThreads = clientCmd.m_requestRaycastIntersections.m_numThreads;
+	const int totalRays = numCommandRays+numStreamingRays;
+	int numThreads = clientCmd.m_requestRaycastIntersections.m_numThreads;
+	if (numThreads == 0) {
+		// When 0 is specified, Bullet can decide how many threads to use.
+		// About 16 rays per thread seems to work reasonably well.
+		numThreads = btMax(1, totalRays / 16);
```
erwincoumans: Is there any limit on the number of threads? We don't want too many threads, I think a limit of say 4 to 32 threads would be good.

examples/SharedMemory/PhysicsServerCommandProcessor.cpp
RanTig: Thanks for catching this!

examples/SharedMemory/PhysicsServerCommandProcessor.cpp
RanTig: This is already limited to min(BT_NUM_THREADS, number of virtual cores) where BT_NUM_THREADS is 64.
Happy to add an additional limit if you still want it

erwincoumans:

PR #1771, opened 2018-06-22
Fix failing CMake due to OpenGL

crewmatt:
Fixes the failing CMake due to older OpenGL build rules.


erwincoumans:
Perhaps start removing removing FindOpenGL on Linux? I think we can also do Windows , but a bit work left. Mac may be more tricky.

Qix-:
Mac is easy. Just add `-framework OpenGL` to the compile/link options. Either that command line works, or the user somehow doesn't have OpenGL capabilities on their machine.

PR #1775, opened 2018-06-24
remove some debug globals from third_party/bullet

bingjeff:
There are some debug global variables that prevent using bullet safely on multi-threaded environments (tsan failures).

PATCH from marioprats@


erwincoumans:
Thanks for creating the pull request!

PR #1776, opened 2018-06-25
Remove the pdController when maxForce = 0

YunfeiBai:


erwincoumans:
Thanks for the contribution. We may also want to remove the PD controllers at 'resetSimulation' in the future.

PR #1784, opened 2018-07-07
Removed lib prefix from Win32 dlls.

mattparks:
Removed the "lib" postfix from Win32 builds when building shared libs, most other cmake projects do this and will not effect release mode or linking.

**Edit:**
Can also be achieved by adding this code to your cmake file, this patch makes bullet match other libraries:
```
SET(CMAKE_SHARED_LIBRARY_PREFIX "" CACHE STRING "The prefix for shared libraries.")
```


madebr:
CMake, by default, does this correct.
On MSVC, *no prefix* is set.
On MINGW, a *lib* prefix is used.

So your patch is incorrect for mingw.

erwincoumans:
I leave cmake changes up to the user and in forks.

PR #1787, opened 2018-07-10
Minor code optimization in multibody forward dynamics

jslee02:
This PR slightly optimizes the multibody forward dynamics computation hopefully not changing the functionality.


erwincoumans:
Thanks, I would expect the compiler to be smart enough to optimize this without const too, but making those variable const doesn't hurt. LGTM, thanks!

jslee02:
Yeah, I believe the modern compiler does the job. I actually expect this change improves the performance by merging two for-loops into one, which is somewhat not easy to see in the diff. 😄

PR #1790, opened 2018-07-13
OpenGL render plugin

BlGene:
### Motivation

Faster rendering could be achieved by writing a plug in that renders with OpenGL(EGL) analogous to the `tinyRendererPlugin.cpp`. 

>  It would be nice to have an EGL/OpenGL plugin indeed. I'm working on other things at the moment. ~erwin

### Performance

GUI+OpenGL(EGL) render fast with single environments, but for several environments (stepped  synchronously) performance deteriorates dramatically to the extent that DIRECT+Tiny rendering becomes quicker. See plot below. This happens even on servers with multiple cores and big GPUs. Can anybody else verify this?

profile_sync
(total = average fps over envs * number of envs)

I don't think this occurs for non-synchronous stepping. I couldn't get good numbers for Tiny-GUI as I cannot run this on the server, it opens a window.

### Progress
The current state can be seen in the video below:

debugb

- [ ] Clear screen?
- [ ] Several render instances (see question below)?
- [ ] Update object poses.
- [ ] ...

As I'd be working on this intermittently, anybody is free to contribute and I'll keep this PR very up to date.

### Questions

My current questions are:
1. Would I need one instance of GLInstancingRenderer per object or is one instance for all objects sufficient? https://github.com/BlGene/bullet3/blob/egl-plugin/examples/SharedMemory/plugins/eglPlugin/eglRendererVisualShapeConverter.cpp#L80


erwincoumans:
GLInstancingRenderer for all objects (in case you figured it out yet)

BlGene:
...I'm assuming that means *one* instance of GLInstancingRenderer for all objects.

dchichkov:
>> This happens even on servers with multiple cores and big GPUs. Can anybody else verify this?

I'd be glad to try it on g3.8xlarge, if you'd tell what branch should I pick. You have "OpenGL render plugin #1790" and "Render options glad2 #1597".  Which one should I try?

On g3.8xlarge I've seen similar performance degradation of the standard pybullet GUI/OpenGL backend, if I'm running 30 parallel processes. And so I'm also forced to use DIRECT+Tiny.

BlGene:
Can you try installing pybullet using distutils, e.g.  `pip install -e .`?

I can't reproduce this error on my machine, and I don't see why its happening. Did you change anything else, or are you using an Apple?

When I build using CMake all the example Apps stick with X11 by default and so does pybullet.

BlGene:
@dchichkov 
If can you get it to work, the best thing to test would by a gym version of `rendertest.py` that I added to the repo: https://github.com/BlGene/bullet3/blob/egl-plugin/examples/pybullet/examples/rendertest_sync.py

dchichkov:
>> Can you try installing pybullet using distutils, e.g. pip install -e .?

Ok.

>>  When I build using CMake all the example Apps stick with X11 by default and so does pybullet. Did you change anything else, or are you using an Apple?

Ubuntu, 16.04.  No changes, this PR and:
export CXXFLAGS=-DBT_USE_EGL
export CFLAGS=-DBT_USE_EGL
./build_cmake_pybullet_double.sh
....
....
[ 42%] Linking CXX executable AppBasicExampleGui
../OpenGLWindow/libOpenGLWindow.so: undefined reference to glad_eglQueryDevicesEXT' ../OpenGLWindow/libOpenGLWindow.so: undefined reference togladLoaderLoadEGL'
../OpenGLWindow/libOpenGLWindow.so: undefined reference to `glad_eglDestroyContext'

I've tried the EGL sample from https://github.com/erwincoumans/egl_example and that one built and ran.

dchichkov:
Similar trend with the MP test, at about 10 renderers FPS maxes out on Tesla M60, DIRECT/eglPlugin.
  Processes   total FPS
    1                 519
    5                 2747
    7                2800
    10              2857
    20              2409

Something is very broken with DIRECT/eglPlugin for me with a more complex model which contains meshes - I'm getting full utilization of Tesla M60, 5FPS just a single process running.  
Overhead  Shared Object                                Symbol                   
  33.12%  [kernel]                                     [k] entry_SYSCALL_64_afte
  15.79%  libc-2.23.so                                 [.] syscall
  11.53%  [kernel]                                     [k] __schedule
  11.23%  [kernel]                                     [k] pvclock_clocksource_r
   4.24%  libc-2.23.so                                 [.] __sched_yield
   3.46%  [kernel]                                     [k] sys_clock_gettime
   2.15%  [kernel]                                     [k] ktime_get_ts64
   1.88%  [kernel]                                     [k] xen_clocksource_get_c
   1.71%  [kernel]                                     [k] posix_ktime_get_ts
   0.74%  libnvidia-eglcore.so.396.37                  [.] 0x0000000000df508d
   0.72%  [kernel]                                     [k] _nv028699rm
   0.63%  [kernel]                                     [k] finish_task_switch
   0.59%  libnvidia-eglcore.so.396.37                  [.] 0x0000000000cc8b55
   0.59%  [kernel]                                     [k] sys_sched_yield
   0.46%  libnvidia-eglcore.so.396.37                  [.] 0x0000000000d115ae
   0.45%  eglRenderer.cpython-35m-x86_64-linux-gnu.so  [.] _ZN32TinyRendererVisu
   0.44%  [kernel]                                     [k] entry_SYSCALL_64_fast
   0.40%  libnvidia-eglcore.so.396.37                  [.] 0x0000000000cc8803

While the same model is fine with  GUI/--window_backend=2 --render_device=0 / renderer=p.ER_BULLET_HARDWARE_OPENGL.  It gives ~450FPS with 10 processes, which is good enough for the moment :)

BlGene:
@dchichkov:
Can you please try the following:
1. try cmake with the latest commit, remember to `rm -r cmake_build`
2. run the MP test with DIRECT/Tiny too, as comparison.
3. (if you have time) check if DIRECT/eglPlugin + complex mesh increase GPU memory consumption for each step.

Also, from looking at profiling timings, I have the suspicion GUI/EGL might be rendering asynchronously to the physics updates. This might be an issue for RL in case this is what you are pursuing, so it might be worth looking into that.

dchichkov:
On g3.4xlarge / NVIDIA 396.37 / Ubuntu 16.04 / 4.4.0-1062-aws

 ```
Processes    FPS/TINY    FPS/eglPlugiin
    1                 374                674
    5                 1861              3267
    7                 2606              3330
    10              2578              5925
    20              3630              6577

```

My test is very crude, but I think the trend still can be seen - 
```
import pybullet as p, os, sys, timeit
p.connect(p.DIRECT)
plane = p.loadURDF("plane100.urdf")
cube = p.loadURDF("cube.urdf",[0,0,1])
N = 1000
print(N*int(sys.argv[1]) / timeit.timeit(lambda:p.getCameraImage(100,100), number=N))
if p.loadPlugin(p.__file__.replace("pybullet", "eglRenderer"), "_tinyRendererPlugin"):
    print(N*int(sys.argv[1]) / timeit.timeit(lambda:p.getCameraImage(100,100), number=N))
```

And I'm just running it with parralel and looking at the last line, assuming that the last to print was running longest, so it would give some estimate on FPS :)  
parallel -j 30 python3 testplugin.py ::: 10 10 10 10 10 10 10 10 10 10

**Comment by dchichkov** [comment]
I'm actually getting a _different_ rendering from the plugin, if it is loaded correctly before loadURDF.  Rendering looks very similar (and correct) for GLX,EGL,Tiny but the picture is different for plugin.
Code to reproduce is below:

```
import pybullet as p, os, sys

if sys.argv[1] == "glx":
    p.connect(p.GUI)
elif sys.argv[1] == "egl":
    p.connect(p.GUI, options='--window_backend=2 --render_device=0')
elif sys.argv[1] == "tiny":
    p.connect(p.DIRECT)
elif sys.argv[1] == "plugin":
    p.connect(p.DIRECT)
    p.loadPlugin(p.__file__.replace("pybullet", "eglRenderer"), "_tinyRendererPlugin")
else:
    print("args should be [glx|egl|plugin|tiny]")
    sys.exit(1)

p.setAdditionalSearchPath("data")

width,height,viewMatrix,projectionMatrix = 128, 128, (0.9999997019767761, 0.00036114765680395067, -0.0006804274744354188, 0.0, -0.00036628328962251544, 0.9999714493751526, -0.007562750019133091, 0.0, 0.0006776766385883093, 0.007562996819615364, 0.9999712705612183, 0.0, 0.0007347110076807439, 0.09311465173959732, -0.22162719070911407, 1.0), (3.732050895690918, 0.0, 0.0, 0.0, 0.0, 3.732050895690918, 0.0, 0.0, 0.0, 0.0, -1.040816307067871, -1.0, 0.0, 0.0, -0.040816325694322586, 0.0)


print(p.getQuaternionFromEuler([0.0, 0.0, 0.0]))
plane = p.loadURDF("plane.urdf")
cube = p.loadURDF("cube.urdf",(-0.2, -0.6, 0.05), (0.0, 0.0, 0.0, 1.0))

if sys.argv[1] in ("egl", "glx"):
    _,_,rgba,_,_ = p.getCameraImage(width=width,height=height,viewMatrix=viewMatrix,projectionMatrix=projectionMatrix, renderer=p.ER_BULLET_HARDWARE_OPENGL)
else:
    _,_,rgba,_,_ = p.getCameraImage(width=width,height=height,viewMatrix=viewMatrix,projectionMatrix=projectionMatrix)

from PIL import Image
Image.frombuffer('RGBA', (width,height), rgba, 'raw').save("0." + sys.argv[1] + ".png")

```

python3 giu.py egl ; python3 giu.py glx ; python3 giu.py plugin ; python3 giu.py tiny

**Comment by dchichkov** [comment]
Tried building with your latest commit - builds.

And another attempt to roughly measure average FPS.  This time much more complex scene, g3.4xlarge, and I've actually verified that the renderings are correct.  plugin is also measured, but the rendering is wrong (see comment above).  
```
Processes   GUI/GLX             GUI/EGL     DIRECT/Tiny   DIRECT/eglPlugin
1               120              113             5               53      
2               255              212             9               70      
5               386              372             18              71      
7               442              417             29              71      
10              444              433             36              70      
20              404              414             45              72      
30              388              407             44              75      
40              368              432             45              76    
```

```
import pybullet as p, os, sys, timeit, multiprocessing
M, N = int(sys.argv[2]), int(sys.argv[3])

def render(n):
  if sys.argv[1] == "glx":
    p.connect(p.GUI)
  elif sys.argv[1] == "egl":
    p.connect(p.GUI, options='--window_backend=2 --render_device=0')
  elif sys.argv[1] == "tiny":
    p.connect(p.DIRECT)
  elif sys.argv[1] == "plugin":
    p.connect(p.DIRECT)
    p.loadPlugin(p.__file__.replace("pybullet", "eglRenderer"), "_tinyRendererPlugin")
  else:
    print("args should be [glx|egl|plugin|tiny]")
    sys.exit(1)

  if sys.argv[1] in ("egl", "glx"):
    p.configureDebugVisualizer(p.COV_ENABLE_RENDERING,0)
    p.configureDebugVisualizer(p.COV_ENABLE_GUI,0)
    p.configureDebugVisualizer(p.COV_ENABLE_SEGMENTATION_MARK_PREVIEW,0)
    p.configureDebugVisualizer(p.COV_ENABLE_DEPTH_BUFFER_PREVIEW,0)
    p.configureDebugVisualizer(p.COV_ENABLE_RGB_BUFFER_PREVIEW,0)

  p.setAdditionalSearchPath("data")
  width,height,viewMatrix,projectionMatrix = 128, 128, (0.9999997019767761, 0.00036114765680395067, -0.0006804274744354188, 0.0, -0.00036628328962251544, 0.9999714493751526, -0.007562750019133091, 0.0, 0.0006776766385883093, 0.007562996819615364, 0.9999712705612183, 0.0, 0.0007347110076807439, 0.09311465173959732, -0.22162719070911407, 1.0), (3.732050895690918, 0.0, 0.0, 0.0, 0.0, 3.732050895690918, 0.0, 0.0, 0.0, 0.0, -1.040816307067871, -1.0, 0.0, 0.0, -0.040816325694322586, 0.0)
  plane = p.loadURDF("plane.urdf")
  #cube = p.loadURDF("cube.urdf",(-0.2, -0.6, 0.05), (0.0, 0.0, 0.0, 1.0))
  cube = p.loadURDF("minis.urdf", [0.0, 0.0, .23], p.getQuaternionFromEuler([3.14157 / 2, 0, 0]))

  if sys.argv[1] in ("egl", "glx"):
    f = lambda:p.getCameraImage(width=width,height=height,viewMatrix=viewMatrix,projectionMatrix=projectionMatrix, renderer=p.ER_BULLET_HARDWARE_OPENGL)
  else:
    f = lambda:p.getCameraImage(width=width,height=height,viewMatrix=viewMatrix,projectionMatrix=projectionMatrix)

  return timeit.timeit(f, number=N)
  #_,_,rgba,_,_ = f()
  #from PIL import Image
  #Image.frombuffer('RGBA', (width,height), rgba, 'raw').save("0." + sys.argv[1] + ".png")  

p = multiprocessing.Pool(processes=M)
FPS = N * M * M / sum(p.imap_unordered(render, range(M)))
open("results.txt", 'at').write("%s %s %.0f" % (sys.argv[1], sys.argv[2], FPS))

```

BlGene:
to summarize:
1. The plugin is broken. These tests were preliminary to see how worth it is investing time in fixing the plugin. I'm assuming that the plugin is merely visually broken and not broken is such a way that computation takes longer, but this is not clear, especially since large meshes cause stalling.
2. DIRECT/Tiny seems to be the best choice for small meshes, GUI/EGL or GUI/GLX for large ones, But it's not clear that GUI runs synchronously with physics. 
3. Its easy to make the mistake of calling `loadPlugin` after `loadURDF`.

next steps:
1. make sure there is no memory leak-type error in plugin, by testing large meshes
2. look at plugin profiling to see what takes time.

dchichkov:
To me, the images looked like the projection or camera matrix were passed to opengl incorrectly. Which impacts FPS, by the way.

dchichkov:
I've looked a bit more into it, it seems that:
 * major performance gap between the plugin and GUI disappears once START_WIDTH / START_HEIGHT are hardcoded to correct values (although displayed GPU load is still much higher).  
 * the difference in the rendered image is in the direction of the axis.  In tiny/GUI the camera is pointing down, but in the rendering by the plugin, it is pointing up.  
* It looked like m_data->m_instancingRenderer->updateCamera() calls were lacking the m_upAxis parameter.  And the default in the eglRendererVisualShapeConverter::m_upAxis(2) while in the 
 examples/OpenGLWindow/GLInstancingRenderer.h/cpp the defaults are 1, for example updateCamera(int upAxis=1).  Tried tweaking it - no effect.  Any clues?

dchichkov:
The code in the original Tiny plugin - TinyRendererVisualShapeConverter::render(const float viewMat[16], const float projMat[16]) is interesting.  Instead of just passing the view and projection matrices to the renderer it is doing something complicated.  This could also be a source of mismatch.

erwincoumans:
I'm looking at this PR again. It would be best to avoid renaming glad/glad_gl.c and glad/glad_glx.c to glad/gl.c and glad/glx.c (same for header). That would reduce the amount of change in this PR a lot, and avoid having to change all the build files (I also have to make changes to build files in some private/internal repositories).

Do you have time to rename the glad files back to original (and only add the new glad/egl.c as glad/glad_egl.c)? Otherwise, if I accept this PR, I'll have to do it myself.

dchichkov:
Note, there are issues with rendering. In particular, writeSingleInstanceTransformToCPU / writeTransforms are never being called. 

It seems, this can be addressed, see:  
https://github.com/BlGene/bullet3/commit/30b57b04e4242d83a8638573bc34e9ac7ed5a23f

But it seems like some parts of the rendering still mismatch to the baseline, for example the default plane.  

It also seems like, there is a score of ways to call sensor rendering (plugin, GUI, GUI_MAIN_THREAD, etc, etc) and at least to me, all of them are broken in their unique ways.  Choosing one particular way and eliminating the rest might improve usability.  As it limits the amount of time a user can spend on attempts to use various ways of rendering sensor output.

erwincoumans:
Let's merge the PR and if someone has time we can address the remaining conversion issues with the plugin. @dchichkov There are 2 renderers: the CPU tiny renderer and the OpenGL 3 GLInstancingRenderer. The connection modes (GUI, DIRECT, SHARED_MEMORY etc) are not directly related to the underlying rendering implementation, aside from some conversion parts.

erwincoumans:
created an issue to track progress of the eglPlugin here:
https://github.com/bulletphysics/bullet3/issues/1867

PR #1801, opened 2018-07-24
Notify plugins on certain events (e.g. body added).

RanTig:
~~This also adds a periodic tick function to plugins, giving them periodically control over the simulation thread, even when the simulation is paused.~~

Hi Erwin, this is the initial implementation of the idea we discussed to notify plugins on changes in Bullet.
It would be great if you could give me some feedback on this change, particularly on:
- The function names tickPlugin and tickPlugins can cause some confusion, since the function calling onPreTick and onPostTick is also called tickPlugins.
- In PhysicsServerExample.cpp, I'm calling the tickPlugins function on every iteration, which might be wasteful, depending on how long the thread sleeps after every iteration. Do you think we should make this less frequent and if so, how? One option would be to always call tickPlugins when a step has been simulated or a command has been processed and otherwise, calling it every X milliseconds. WDYT?

### DIFF
--- a/examples/SharedMemory/PhysicsCommandProcessorInterface.h
+++ b/examples/SharedMemory/PhysicsCommandProcessorInterface.h
@@ -44,7 +44,7 @@ public:
 	virtual void enableRealTimeSimulation(bool enableRealTimeSim)=0;
 	virtual bool isRealTimeSimulationEnabled() const=0;
 
-	virtual void tickPlugins() = 0;
+	virtual void reportNotifications() = 0;
 
 	virtual void enableCommandLogging(bool enable, const char* fileName)=0;
 	virtual void replayFromLogFile(const char* fileName)=0;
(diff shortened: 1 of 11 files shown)


erwincoumans:
Please rename the function 'reportNotification'. 

Separately, we can add a function to regularly add a 'stepPlugin' function. Let me also look at how to regularly call the stepPlugin. Shall we do that as a separate CL/PR?

RanTig:
Changed tickPlugin (the plugin callback) to processNotifications
Changed tickPlugins to reportNotifications
Only calls processNotifications when there are notifications available.

erwincoumans:
Thanks for the PR. When are the reportNotifications/processNotification callback called exactly?
After each command that is processed, as well as in each postTickCallback (if a step simulation is executed)?

erwincoumans:
LGTM. In the future we may want to only add/process notifications if there is any plugin that implements the processNotification callback (as an optimization).

PR #1803, opened 2018-07-24
Renamed error_message and warning_message macros.

mbeards:
error_message and warning_message are common strings that are likely to
collide. Renamed to bt_id_{error,warning}_message to more strongly
uniquify.

grep -r error_message -l src/BulletInverseDynamics | \
  xargs sed -i -e "s/error_message/bt_id_error_message/g"

grep -r warning_message -l src/BulletInverseDynamics | \
  xargs sed -i -e "s/warning_message/bt_id_warning_message/g"


erwincoumans:
Good idea, will merge once the CI is finished ok.

erwincoumans:
some 'internal server error' in CI? Let's close/re-open to trigger CI.

Issue #1800, opened 2018-07-23
Memory leak in getClosestPoints?


haudren:
I am using pybullet to compute distances between a bunch of objects in a scene, but my RAM gets full very quickly... I tried to profile using `memory_profiler` and it seems that the memory allocated by `getClosestPoints` is never freed.

I wrote the following repro program:
```python
import pybullet as p
import pybullet_data
import os
import numpy as np

nr_ducks = 100
nr_steps = 1000

p.connect(p.DIRECT)

root = pybullet_data.getDataPath()

ducks = [p.loadURDF(os.path.join(root, 'duck_vhacd.urdf')) for i in range(nr_ducks)]

for duck in ducks:
    p.resetBasePositionAndOrientation(duck, 0.05*np.random.rand(3), [0, 0, 0, 1])

for i in range(nr_steps):
    distances = [p.getClosestPoints(ducks[0], duck, 0.01) for duck in ducks[1:]]
    del distances  # Should not even be needed

p.disconnect()
```
The resulting memory usage (`mprof run leak_pybullet.py`):
![leak_pybullet](https://user-images.githubusercontent.com/4352562/43060933-017bec26-8e8e-11e8-9827-8a0e09d75c0b.png)

erwincoumans:
Thanks for the report, we'll check it out.

erwincoumans:
Thanks for the report, should be fixed here: https://github.com/bulletphysics/bullet3/pull/1804

PR #1804, opened 2018-07-25
fix memory leak reported in Issue #1800

erwincoumans:
avoid access to array elements outside of range

### DIFF
--- a/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
+++ b/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
@@ -5022,8 +5022,10 @@ bool PhysicsServerCommandProcessor::processSyncUserDataCommand(const struct Shar
 
 	b3AlignedObjectArray<int> userDataHandles;
 	m_data->m_userDataHandles.getUsedHandles(userDataHandles);
-	memcpy(bufferServerToClient, &userDataHandles[0], sizeof(int) * userDataHandles.size());
-
+	if (userDataHandles.size())
+	{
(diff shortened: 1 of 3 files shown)


examples/SharedMemory/PhysicsServerCommandProcessor.cpp
```diff
@@ -5022,8 +5022,10 @@ bool PhysicsServerCommandProcessor::processSyncUserDataCommand(const struct Shar
 
 	b3AlignedObjectArray<int> userDataHandles;
 	m_data->m_userDataHandles.getUsedHandles(userDataHandles);
-	memcpy(bufferServerToClient, &userDataHandles[0], sizeof(int) * userDataHandles.size());
-
+	if (userDataHandles.size())
```
haudren: Is the check really necessary? I was under the impression that if the size is zero, `memcpy` would copy nothing...

haudren:
Thanks for the fix, it seems like it's no longer leaking !

haudren:
Actually, my app seems to be still leaking albeit less seriously than before whereas the duck example is fixed. I'm investigating.

examples/SharedMemory/PhysicsServerCommandProcessor.cpp
erwincoumans: The check is to avoid accessing non-existing elements in the array (userDataHandles[0], while the size == 0). It causes asserts in debug mode.

erwincoumans:
Thanks, I added your leak fix in the PR.

PR #1808, opened 2018-07-28
Fixed shared libs being enabled when not building pybullet.

mattparks:
Partial fix for #1794 when not looking to compile pybullet, shared libs still need to be fixed for win32.


erwincoumans:
I haven't looked into the pr yet, you want to close it?

mattparks:
It's pointless

PR #1811, opened 2018-08-01
Adds a simulation stepped notification

RanTig:
This notification is called when the simulation is stepped, regardless
of whether any objects have moved.

Without this notification, there was no way for a plugin to tell when the simulation had been stepped but nothing had moved, since processNotifications would not be called in that case. The workaround would've been requesting the state and comparing with a previous snapshot after onPostTick, which would be undesirable.


erwincoumans:

PR #1814, opened 2018-08-02
added additional parameter to btConcaveShape::processAllTriangles()

gjaegy:
Added an additional parameter to btConcaveShape::processAllTriangles() giving information about the other colliding object when appropriate.

In our case, we need to access the rigid body AABB (in addition to the sub-shapes AABB given in the original parameters) in order to have a consistant level-of-detail selection of the terrain chunks we use to perform the collision with (we perform that selection according to the size of the object, this allows us to have a more refined terrain collision mesh resolution when using small objects such as cars, and rougher resolution when colliding big objects such as aircraft - otherwise we end up testing way too many triangles, or we encounter LOD switches in subsequent frames, which is not good either since not giving a consistant collision mesh for the terrain).

### DIFF
--- a/Extras/GIMPACTUtils/btGImpactConvexDecompositionShape.cpp
+++ b/Extras/GIMPACTUtils/btGImpactConvexDecompositionShape.cpp
@@ -210,7 +210,7 @@ btGImpactConvexDecompositionShape::~btGImpactConvexDecompositionShape()
 {
 	delete m_decomposition;
 }
-void btGImpactConvexDecompositionShape::processAllTriangles(btTriangleCallback* callback,const btVector3& aabbMin,const btVector3& aabbMax) const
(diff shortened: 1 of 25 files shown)


gjaegy:
Any news on that one ?

erwincoumans:
It is best to just maintain a modified/forked version of Bullet, I don't want to change/break the API at this stage. Thanks for sharing the patch though.

gjaegy:
OK, it's a shame (for us :)) but I can understand that. Thanks for the quick answer !

gjaegy:
Erwin, just had a thought about your answer and your will to not change the API at this stage: I've added a single pointer parameter at the end of the processAllTriangles() mehod:

`void btGImpactConvexDecompositionShape::processAllTriangles(btTriangleCallback* callback,const btVector3& aabbMin,const btVector3& aabbMax,const btCollisionObject* otherObject) const`

Actually, setting the default value of that parameter to NULL won't break any existing code at all, so the impact of that API change will likely be invisible to all users in my opinion (but I might have missed some cases of course).

Just trying to defend my case (I prefer using unmodified third-party libraries) :)

PR #1818, opened 2018-08-03
Add MLCP constraint solver for multibody

jslee02:
**Overview**

This PR adds `btMultiBodyMLCPConstraintSolver`, an MLCP based constraint solver for multi-body. The class intended to support multibody, but it also works with rigid-body.

The class inherits `btMultiBodyConstraintSolver` to be compatible with `btMultiBodyDynamicsWorld`, but it has many things in common with `btMLCPSolver`, an MLCP constraint solver for rigid-body. We could consider refactoring shared code parts of `btMLCPSolver` as suggested by @erwincoumans.

It internally maintains two sets of methods: (a) methods for only rigid-bodies and (b) methods for the mix of rigid-bodies and multi-bodies. As (b) works for any case, we could use it for every case. The motivation behind having (a) is that we can decouple the constraint sets into rigid-bodies and mix of rigid/multi-bodies so that they can be solved independently. (a) can solve rigid-body constraints more efficiently than (b) because (b) has some overhead to be general for both cases.

**Tests**

This change is qualitatively tested on an existing demo (`MultiDofCreateFunc`; maybe this should be renamed to `MultiDof`?) and a new demo (`Serial Chains`). The two demo uses different constraints solvers per every new instantiation: SI, MLCP + PGS, MLCP + Dantzig, and MLCP + Lemke.

All constraint solvers visually equal solutions

image
image

**Possible Future Improvements**

MLCP based constraint solving doesn't scale well due to the algorithm complexity of the MLCP solvers, which is not surprising. We may consider introducing maximum total constraint number so that it fallbacks to using the sequential impulse method (SI) when the number of constraints exceeds the threshold. As a note, it fallbacks to using SI when the MLCP solver fails to solve, like `btMLCPSolver`.

There are many places can be parallelized such as constructing A matrix (each row and column can be computed in parallel), solving two distinct MLCPs, and so on.

**Minor Note**

The code is formatted by clang-format format with `_clang-format` configuration.


erwincoumans:
Great, thanks a lot for this contribution!
At a first glance, the direct solver(s) such as Dantzig and Lemke require some CFM:
~~~~~~~~
m_dynamicsWorld->getSolverInfo().m_globalCfm = 1e-4; //todo: what value is good?
~~~~~~~~
Let's look at it more in detail next week.

jslee02:
The instability was actually due to [this mistake](https://github.com/bulletphysics/bullet3/pull/1818/commits/340236ee29472aae9c13ec901ac91bb1e07a9d2c). Now all the LCP solvers except Lemke works (without blowing up) even with zero CFM. I added some CFM just in case.

As we discussed, Lemke implementation needs some work, so I removed from the demo for now.

erwincoumans:
Great, thanks a lot of tracking it down. Let me review and LGTM it asap.

erwincoumans:
Thanks! I'll also add it to the premake4.lua file after merge. Also, we should add it to the pybullet and to setup.py setup tools build (in the root of this repo)

PR #1822, opened 2018-08-07
Add `format` target to format code using ClangFormat 3.8

jslee02:
This PR adds `format` target that is only generated when Bullet is configured by CMake and `clang-format-3.8` is installed. `format` formats the code according to the configuration file (i.e., `_clang-format` in the root directory). You can format C/C++ files that are specified as:
```cmake
format_add_sources(btScalar.h)  // single file
format_add_sources(btScalar.h btScalar.cpp ...)  // multiple files
```
Note that this PR doesn't set any files to be formatted.

In addition, `check-format` is added to only check whether all the (specified) code is compliance to the code style.

I specified the ClangFormat version different versions result in different code formatting for the same configuration. ClangFormat 3.8 is chosen since it's the latest version supported by Ubuntu Trusty. I'm open to using other version though.

Also, I had to replace `SUBDIRS()` with `add_subdirectory()`. `SUBDIRS()` doesn't work with CMake properties for some reason. [`SUBDIRS()` is deprecated](https://cmake.org/cmake/help/v2.8.12/cmake.html#command:subdirs) anyway. I replaced only necessary uses of `SUBDIRS()` to make this PR small, but it would be worth to replace others.

Related: #1549


erwincoumans:
Thanks for the contribution, will look at it soon.

aaronfranke:
Related to #2689

erwincoumans:
trigger CI.

PR #1833, opened 2018-08-10
Add blocked Gauss-Seidel constraint solver for rigid body

jslee02:
This PR adds `btBGSSolver`, an implementation of the blocked Gauss-Seidel (BGS) constraint solver.

Loosely speaking this class is a mix of `btSequentialImpulseConstraintSolver` and `btMLCPSolver`. Specifically, BSG forms many small MLCPs of subsets of the constraints and solves each MLCP using solvers suitable for small-sized problems (e.g., Dantzig), and then applies the Gauss-Seidel splitting to the blocked MLCPs unlike the regular Gauss-Seidel applies it to every single constraint. A block is usually defined as all constraints associated with a single contact point or any set of constraints better to be solved together at once. 

Currently, `btBGSSolver` constructs MLCP blocks only for contact constraints, but we can extend the class to be able to form an MLCP for an arbitrary set of constraints.

The expected performance is somewhere between `btSequentialImpulseConstraintSolver` and `btMLCPSolver` in terms of speed and accuracy. From my testing, the performance was much closer to SI.

The new constraint solver was qualitatively tested on `BasicExample`: https://youtu.be/tojxQnd_RQI

As a note, more comprehensive tests will be done once BGS for multibody and mixed constraint solver is implemented, which will be followed. This is because to see the performance differences in one simulation.

### DIFF
--- a/examples/BasicDemo/BasicExample.cpp
+++ b/examples/BasicDemo/BasicExample.cpp
@@ -23,9 +23,16 @@ subject to the following restrictions:
 
 #include "LinearMath/btVector3.h"
 #include "LinearMath/btAlignedObjectArray.h"
+#include "BulletDynamics/MLCPSolvers/btBGSSolver.h"
+#include "BulletDynamics/MLCPSolvers/btMLCPSolver.h"
+#include "BulletDynamics/MLCPSolvers/btDantzigSolver.h"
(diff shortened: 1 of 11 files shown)


erwincoumans:
Thanks for the PR! Could you rename it to btBlockGSSolver?

jslee02:
Let me close this PR and create a revised version later.

PR #1842, opened 2018-08-23
Allow current positions for Calculate IK.

crewmatt:
Allows the user to specify the current positions of joints when
calculating Inverse Kinematics.


erwincoumans:
Thanks for the contribution!

PR #1843, opened 2018-08-23
preserve constness to avoid compiler warnings

rhaschke:
Minor fix to avoid compiler warning when more strict checking (`-Wcast-qual`) is used.


erwincoumans:
Thanks, that looks good to me!

Issue #1827, opened 2018-08-09
Cannot build on Visual Studio 2017 (v141) with Treat Warnings as Errors (/WX)


cmbasnett:
```
1>------ Build started: Project: naga, Configuration: Debug Win32 ------
1>boilerplate
1>physics_simulation.cpp
1>d:\github\mandala\ext\bullet3-2.87\src\linearmath\btidebugdraw.h(169): error C2220: warning treated as error - no 'object' file generated
1>d:\github\mandala\ext\bullet3-2.87\src\linearmath\btidebugdraw.h(169): warning C4305: 'argument': truncation from 'double' to 'const btScalar'
1>d:\github\mandala\ext\bullet3-2.87\src\linearmath\btidebugdraw.h(170): warning C4305: 'argument': truncation from 'double' to 'const btScalar'
1>d:\github\mandala\ext\bullet3-2.87\src\linearmath\btidebugdraw.h(171): warning C4305: 'argument': truncation from 'double' to 'const btScalar'
1>rigid_body_component.cpp
1>d:\github\mandala\ext\bullet3-2.87\src\linearmath\btidebugdraw.h(169): error C2220: warning treated as error - no 'object' file generated
1>d:\github\mandala\ext\bullet3-2.87\src\linearmath\btidebugdraw.h(169): warning C4305: 'argument': truncation from 'double' to 'const btScalar'
1>d:\github\mandala\ext\bullet3-2.87\src\linearmath\btidebugdraw.h(170): warning C4305: 'argument': truncation from 'double' to 'const btScalar'
1>d:\github\mandala\ext\bullet3-2.87\src\linearmath\btidebugdraw.h(171): warning C4305: 'argument': truncation from 'double' to 'const btScalar'
1>Generating Code...
1>Done building project "engine.vcxproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
```

madebr:
Maybe wrap the values with `btScalar` for the case Bullet is compiled with `USE_DOUBLE_PRECISION=ON`?

PR #1849, opened 2018-08-30
Fix for bug #1827

madebr:
Same as #1828, but with the feedback of @jslee02 included.

I just want to make sure that this fix is included in the next bullet release because the warnings caused by this are the only ones remaining in our project.


erwincoumans:
Thanks, looks good to me.
close/re-open, due to some issue in cmake files, sometimes CI fails (copy files related)

erwincoumans:
Thanks! LGTM

PR #1852, opened 2018-08-31
Removes visual shape data when a body gets removed.

RanTig:
When you add a body, remove it and then add a new body, the body id may be reused by Bullet.
Because the visual shape data was not being removed when a body was
removed, requesting the visual shape data for a 'recycled' body id
results in both the visual shape data of the new body as well as the old one.
Removing the visual shape data when the body gets removed fixes this.

Changes the btAlignedObjectArray for visual shapes to a hashmap, so
that removing is faster. Additionally, functions like getNumVisualShape
don't perform a linear search anymore.


erwincoumans:
Thanks for the contribution. Can you add a space between the >> so it compiles?
~~~~~~~~~~~~
btHashMap<btHashInt,btAlignedObjectArray<b3VisualShapeData>   > 
~~~~~~~~~~~~

PR #1854, opened 2018-09-03
Small fix in executePluginCommand

RanTig:
Small fix in executePluginCommand, now you can pass int arguments while keeping floats null without errors.


erwincoumans:
Thanks for the fix!

PR #1855, opened 2018-09-04
more work on grpc/proto/pybullet.proto

erwincoumans:


BluePrintRandom:
could one make a multiplayer pybullet server with permissions with this?

kinda like second life?

PR #1872, opened 2018-09-11
Change b3GetLinkState to use numLinks in status handle

RanTig:
Changes the b3GetLinkState function to use numLinks from the
SharedMemoryStatus returned from RequestActualState instead of the
cached number of link value returned by b3GetNumJoints.

The cached value can be outdated when a new body is added and
SyncBodyInfo isn't called, while using the value in the status is always
up-to-date.


erwincoumans:
LGTM, thanks!

PR #1875, opened 2018-09-12
EGL Plugin window resize fix

BlGene:
I looks like rendering sizes were not being correctly passed to the OpenGL renderer. This means that the the default resolution of 640x480 was being used, especially for small image sizes such as 84x84 this resulted in a 6x speed increase.

Thanks @dchichkov for finding this bug.


erwincoumans:
Closing and opening the window each time you request a different camera image width/height is not practical. We should look into creating a temporary framebuffer at the right size.

Also, this PR has a lot of white space changes, making it hard to spot the actual changes, can you undo them? I need to start activating the use of Clang format (https://github.com/bulletphysics/bullet3/pull/1822)

BlGene:
>Closing  and opening the window each time you request a different camera image  width/height is not practical. We should look into creating a temporary  framebuffer at the right size. 

Would using/fixing `m_data->m_window->getResizeCallback()(width, height);` suffice?

>Also, this PR has a lot of white space changes, making it hard to  spot the actual changes, can you undo them? I need to start activating  the use of Clang format (#1822)

fixed. My preference would be 4 space indentation and spaces only, no tabs.

dchichkov:
If possible, I would advocate for pre-allocating a single buffer (or opening a single window) and using view-ports on it for each camera, rather than allocating separate buffers.  View ports can be set by [glViewport](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glViewport.xhtml), and it is a fast operation.  This has the advantage of that a continuous region of GPU memory can be used in other operations (for example inference on a batch of rendered images, rather than on an individual 84x84 image).

BlGene:
I've added a `updateWindow` function to `EGLOpenGLWindow` which resizes it without closing. This seems to do the trick.

If somebody has time they could double check if they also see 10% performance drop between 
615f917 and af3f41b.

dchichkov:
Resizing a window is still a costly call which results in GPU memory management - not ideal.   It still seems to me that glViewport is a better way of doing it (and also an approach to single-pass/multiple viewport rendering and running inference in a batch mode).

BlGene:
It’s hopefully good enough for merging. I’ll have a look at glviewport if I start needing multiple viewport rendering. Feel free to submit a more ideal PR though.

dchichkov:
well, just in case - glViewport(x,y,width,height) is straightforward, once you call it, rendering calls and glCopyPixels are directed into a window region - you'll get same effect as calling window resizing, but with no memory management.

BlGene:
It seems like there are still some issues:

- [ ] Vertex colors don't work, I'm importing from `.sdf`
- [ ] shapes don't get deleted with `pybullet.resetSimulation`

erwincoumans:
We now use glViewport to select the right image size when rendering, so glReadPixels reads the right size image. This issue can be closed now.

erwincoumans:
See also

erwincoumans:
>> shapes don't get deleted with pybullet.resetSimulation

Are the not removed in
~~~~~~~~~~~
void EGLRendererVisualShapeConverter::resetAll()
~~~~~~~~~~~

PR #1882, opened 2018-09-18
Spelling fix: on-way -> one-way

lucasw:


erwincoumans:
Thanks for the attention to detail.

Issue #1889, opened 2018-09-20
Expose tangential friction forces in pybullet.getContactPoints


jviereck:
Right now `p.getContactPoints(...)` only allows access to the normal force. The todo in `struct b3ContactPointData` [1] hints it's also possible to expose the tangential friction forces.

Is it complicated to get access to the tangential contact forces and expose them through `p.getContactPoints(...)`? I tried to figure out where `struct b3ContactPointData.m_normalForce` is set, but with no luck.

[1]: https://github.com/bulletphysics/bullet3/blob/d315673d3a17973149307c7455750f1f2151458a/examples/SharedMemory/SharedMemoryPublic.h#L585-L591

jviereck:
Closed by #1896 and #1919.

PR #1896, opened 2018-09-24
Adding support for lateral friction to getContactPoints()

jviereck:
This addresses #1889. Is there a good way to write a unit test for this kind of features?


erwincoumans:
I just applied a clang-format on the source tree, which may result in conflicts.
Can you try applying the script clang-format-all.sh in the root (works on Windows, Mac and Linux).
Alternatively, I can apply the diff with no-whitespace manually.
(using https://github.com/bulletphysics/bullet3/pull/1896.diff)

jviereck:
> I just applied a clang-format on the source tree, which may result in conflicts.

Rebased on current master and ran `clang-format-all.sh`.

PR #1899, opened 2018-09-26
Add const qualifiers to the double [] inputs in PhysicsClientC_API

WenlongLu:


erwincoumans:
Thanks for the contribution, LGTM!

PR #1900, opened 2018-09-26
Update URDF Importer to process Sensor elements.

crewmatt:
Update the URDF Importer to add sensors as 0 mass 0 inertia elements
attached by a fixed joint. This way their states can be read as links.


erwincoumans:
We use a different code style at the moment (tabs instead of spaces etc),
can you run clang-format-all.sh ? or clang-format using the _clang-format file in the root of Bullet?
Also, where is the loadUrdf called with the new boolean?

crewmatt:
Updated formatting and included flag plumbing

erwincoumans:
Thanks for the contribution, let's merge it.

PR #1902, opened 2018-09-27
fix mingw build

cloudwu:
I can't build it because of lib dependency issue like 

```
[ 54%] Linking CXX executable App_RobotSimulator_NoGUI.exe
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ../../lib/libBulletRobotics.a(win32.obj):win32.c:(.text+0x14): undefined reference to `__imp_WSAStartup'
```

We should link ws2_32 **after** libBulletRobotics.a .


erwincoumans:
Thanks for the PR!

PR #1910, opened 2018-10-02
Two minor updates for PhysicsServerCommandProcessor

WenlongLu:
1) Update the status when removing collision shape is successful.
2) Pass plane constant along when creating collision shape.


erwincoumans:

PR #1941, opened 2018-10-16
Remove unneeded keyword `register`

ShFil119:
Why is it unneeded?

C++11 has deprecated it,
c++17 removed.

So, how about older (c++98)?

According to Herb Sutter, it is as meaningful
as whitespace.

http://www.drdobbs.com/keywords-that-arent-or-comments-by-anoth/184403859


erwincoumans:
Looks good to me, thanks!

PR #1946, opened 2018-10-22
Uses btCollisionBody::isActive instead of btMultiBody::isAwake to determine whether to send transform changed notifications.

RanTig:
Uses btCollisionBody::isActive instead of btMultiBody::isAwake to determine whether to send transform changed notifications.


erwincoumans:
Thanks, LGTM.

PR #1951, opened 2018-10-24
Added missing extern declaration

duckdoom5:
The extern declaration  was missing from this file.


erwincoumans:
(retrigger close/open, likely some glitch in CMake file handling make windows CI build fail sometimes)

erwincoumans:
sure, LGTM.

PR #1978, opened 2018-11-02
Added Benchmark Demo for packed convex shapes.

stolk:
I added a benchmark demo that simulates a packing of 90 convex shapes, falling to the ground.

In release, on my machine, I see up to 6ms internal sim step.

When in SAT mode, it spikes to 90ms internal sim step.


erwincoumans:
Thanks Bram, I'll check it out soon. We should be able to increase the SAT performance much, but have lots of more important (from our perspective) work, related to [PyBullet](https://docs.google.com/document/d/10sXEhzFRSnvFcl3XxNGhnD4N2SedqwdAvK3dsihxVUA/edit#heading=h.2ye70wns7io3).

erwincoumans:
Thanks for the demo! 
I noticed a camera issue (ok after reset), but that is an issue with the benchmark demo. There is another issue when calling 'initializePolyhedralFeatures': the demo shows visible penetration, since there is a difference between collision shape and graphics shape. This should be fixed in this commit: https://github.com/bulletphysics/bullet3/pull/1990

PR #1996, opened 2018-11-18
Fix Visual Studio section of BulletQuickstart.pdf

### LINKED ISSUE #902: Premake Batch File Munged, Documentation Incorrect
Just downloaded ver 2.85.1 for Win.  Following the Quickstart pdf, as recommended in main documentation.  Paragraph 2.1.1 in pdf refers to non-existent directories and files.  Guessing at the actual meaning led me to run vs2010_bullet2gpu.bat, which generated error  about "--bullet2gpu" being an invalid switch.  Removed this switch from the batch file, and Premake4 ran and generated files -- too early to say if they're the right files, etc.

The documentation and the batch file need attention.  Realize that for anyone trying Bullet for the first time (or the first time in awhile) on Windows, this issue will be their first impression.

bonizz:
Fixes #902

This should make things easier for newcomers building Bullet with Visual Studio.

Updating path to vs2010.bat and 0_Bullet3Solution.sln, and adding vs2010.bat which does not require vr components.
pdf file generated with the command `latexmk -pdf BulletQuickstart.tex` using MiKTeX.


wmcnamara:
Will this ever be merged? Was very frustrated with the quickstart guide and this PR helped a lot.

PR #1999, opened 2018-11-19
Show btHinge2 oscillation.

stolk:
This demonstrates the oscillation in Hinge2.
It gets progressively worse, if you let it run for a while.


stolk:
I've added the fix to this pull request.
This change contains:
* modified demo to show issue.
* fix for #2000 issue.

erwincoumans:
Thanks for tracking this down, Bram!

>>so it is much simpler (and more robust) just to simply use inf (with the proper sign)

There may be cases where using +/- INF is more robust according to the comment.
It would be nice to add an option to the constraint to enable/disable this, just in case someone finds a regression.

~~~~~~~~~~~~
		// after the spring force affecting the body(es) the new velocity will be
		// vel + f / m * (rotational ? -1 : 1)
		// so in theory this should be set here for m_constraintError
		// (with m_constraintError we set a desired velocity for the affected body(es))
		// however in practice any value is fine as long as it is greater then the "proper" velocity,
		// because the m_lowerLimit and the m_upperLimit will determinate the strength of the final pulling force
		// so it is much simpler (and more robust) just to simply use inf (with the proper sign)
		// you may also wonder what if the current velocity (vel) so high that the pulling force will not change its direction (in this iteration)
		// will we not request a velocity with the wrong direction ?
		// and the answare is not, because in practice during the solving the current velocity is subtracted from the m_constraintError
		// so the sign of the force that is really matters
#if 0
		info->m_constraintError[srow] = (rotational ? -1 : 1) * (f < 0 ? -SIMD_INFINITY : SIMD_INFINITY);
#else
		// Using infinite constraints makes the damping not work, and oscillate for ever.
		info->m_constraintError[srow] = vel + f / m * (rotational ? -1 : 1);
#endif
~~~~~~~~~~~~

stolk:
Added an option to use infinite constraint error by setting BT_6DOF_FLAGS_USE_INFINITE_ERROR in m_flags of 6DofSpring2 constraint.

erwincoumans:
Thanks, LGTM!

Issue #2000, opened 2018-11-19
btHinge2 oscillates.


stolk:
Even with using btHinge2Constraint::setDamping() the spring will remain oscillating.

I modified Hinge2Vehicle demo to show this issue:
https://github.com/bulletphysics/bullet3/pull/1999

A video of the issue:
https://www.youtube.com/watch?v=WHhR5poTSew

stolk:
Ha! I found the bug in Bullet's constraint solver, and fixed it.

The hack with +/- infinite constraints is what caused it. I removed it, and now both 6Dof and Hinge2 joints are working properly.

Will submit pull rq.

stolk:
Fixed in main development branch.

PR #2016, opened 2018-12-01
Fix for #2000

GaborPuhr:
Tangential velocity is now added to the linear velocity when the current velocity is calculated for the calculation of the damping force in btGeneric6DofSpring2Constraint.
Better mass estimation while calculating spring forces in btGeneric6DofSpring2Constraint.
BT_6DOF_FLAGS_USE_INFINITE_ERROR flag has been removed as it's no longer needed.


erwincoumans:
This relates to Issue #2000. If this causes a regression, we should add the option to not use INIFINITY there. See https://github.com/bulletphysics/bullet3/issues/2000

PR #2042, opened 2019-01-04
Increase mesh allocation for vertices/indices in PyBullet.createColli…

erwincoumans:
…sionShape

See createMesh.py for an example.

The data has to fit in shared memory, hence the limit on Mac is lower than Windows and Linux:

#ifdef __APPLE__
#define B3_MAX_NUM_VERTICES 8192
#define B3_MAX_NUM_INDICES 32768
#else
#define B3_MAX_NUM_VERTICES 131072
#define B3_MAX_NUM_INDICES 524288
#endif

### DIFF
--- a/examples/SharedMemory/PhysicsClientC_API.cpp
+++ b/examples/SharedMemory/PhysicsClientC_API.cpp
@@ -1343,7 +1343,7 @@ B3_SHARED_API int b3CreateCollisionShapeAddConcaveMesh(b3PhysicsClientHandle phy
 			command->m_createUserShapeArgs.m_shapes[shapeIndex].m_numVertices = numVertices;
 
 			int totalUploadSizeInBytes = numVertices * sizeof(double) *3  + numIndices * sizeof(int);
-			char* data = (char*)malloc(totalUploadSizeInBytes);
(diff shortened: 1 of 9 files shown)


examples/SharedMemory/PhysicsClientC_API.cpp
```diff
@@ -1342,21 +1342,28 @@ B3_SHARED_API int b3CreateCollisionShapeAddConcaveMesh(b3SharedMemoryCommandHand
 			command->m_createUserShapeArgs.m_shapes[shapeIndex].m_meshFileName[0]=0;
 			command->m_createUserShapeArgs.m_shapes[shapeIndex].m_numVertices = numVertices;
 
+			int totalUploadSizeInBytes = numVertices * sizeof(double) *3  + numIndices * sizeof(int);
+			char* data = (char*)malloc(totalUploadSizeInBytes);
```
ShFil119: `std::vector` or `new`?

erwincoumans: Yes, let's use new instead of malloc here. I was just editing pybullet.c which is C only.

PR #2043, opened 2019-01-04
initialize m_numUserCollisionShapes correctly for b3InitRemoveBodyCom…

crashcover:
Due to the missing initialization of the variable
`m_numUserCollisionShapes`
the "collisionShape used" check in PhysicsServerCommandProcessor.cpp:8956 was executed randomly.


erwincoumans:
Thanks for the fix, good find!

PR #2050, opened 2019-01-10
Update btDbvt.h, slightly faster Intersect for btDbvtAabbMm

TrianglesPCT:
use movemask instead of the casting and scalar ops


erwincoumans:

PR #2051, opened 2019-01-10
Added SIMD implementation of btDbvtAabbMm::Contain(const btDbvtAabbMm& a)

TrianglesPCT:
Added SIMD implementation of btDbvtAabbMm::Contain(const btDbvtAabbMm& a)


erwincoumans:

PR #2052, opened 2019-01-10
SIMD version of NotEqual for btDbvtAabbMm

TrianglesPCT:
Only uses SSE1


erwincoumans:

PR #2053, opened 2019-01-10
Enable deferred collide for btDbvtBroadphase

TrianglesPCT:
See post on forum: https://pybullet.org/Bullet/phpBB3/viewtopic.php?f=9&t=12571

It is significantly faster when many objects are active and moving.

It everything is sleeping it is probably slightly slower, but I don't think it makes sense to optimize for the fast case--


erwincoumans:
We don't change defaults anymore. At best, we can expose an API to allow the user to set it.

erwincoumans:
We don't change behavior for existing users for this feature. A little API to expose the setting would be acceptable.

PR #2055, opened 2019-01-11
Pybullet enhancements

xantares:
- Add pybullet install rule
- OSX Python extension is .so
- Print build time only in debug mode
- Install dependent libs
- Fix osx linking of Python libs


erwincoumans:
Thanks for the contribution. We use pip nowadays to install PyBullet, but it is nice to also allow cmake indeed.

>> Print build time only in debug mode

I intentionally want to see the build time, also in release mode. It makes it more obvious if an old version is installed. Can you undo that part, and I 'll merge the PR.

xantares:
hello @erwincoumans, can you merge it ?

erwincoumans:
Sure, thanks for the contribution!

PR #2058, opened 2019-01-15
For pybullet_envs/deep_mimic, adding a specialized inverse kinematics, reference render and support for a simplified Human3.6M dataset

Somedaywilldo:
Dear maintainers, 

I've developed a specialized inverse kinematics module for deep_mimic environment, should be more efficiently.

And I've added the support for a human motion capture dataset created by [VideoPose3D](https://github.com/facebookresearch/VideoPose3D), the environment can extract data, render the ground truth using 3D-coordinates, should be useful.
The detailed usage is written in the README.md,

Looking forward to your reply,

Best wishes,

Somedaywilldo


erwincoumans:
Thanks a lot for the surprising contribution. I'm still actively looking how to get the basic deep_mimic flow to work, including training a simple model and testing/enjoying it.

I want to avoid adding a torch dependency to in this directory, keep PyBullet simple to use and install.

It looks like the inverse_kinematics is used purely offline as a preprocessing step in conversion from an external motion capture dataset? It would be OK to have torch in the optional offline step, since people can just use the included humanoid/humanoid.urdf, which is a 1:1 conversion from the original 
DeepMimic project.

So I'm not sure if I can merge this PR, but I'll keep this PR in mind while getting the training/testing up and running. Thanks a lot!

Somedaywilldo:
Thanks for your reply!  I'm trying to re-implement the deepmimic environment and algorithm these days, I will get rid of the PyTorch dependency.  I think I'm quite close to get the training/testing running, when I'm done I'll make another pull request.  : ) @erwincoumans

erwincoumans:
Note that there is now a new example that allows to use either spherical joint motor (setJointMotorControlMultiDof) or using stable PD control. The CL is here, video here:
https://github.com/bulletphysics/bullet3/pull/2047
Enabling spherical stable PD control still requires an update in the python setup.py script (and cmake, premake). In a nutshell, that code relies on Eigen, which is now included in Bullet/examples/ThirdPartyLibs/Eigen.
For stable PD control for spherical joints, it uses the inverse dynamics/mass matrix computation from Jason Peng. I'm hoping to use joint motors though, it may be faster and more stable.
See video here: https://www.youtube.com/watch?v=zwEEQd9nLz8

erwincoumans:
Exciting times, this QuaterNet also uses Human3.6M:
https://venturebeat.com/2019/01/24/facebook-and-googles-ai-generates-3d-human-poses
https://arxiv.org/pdf/1901.07677.pdf

Somedaywilldo:
I've finally removed the torch dependencies, so now we're about to use Human3.6M dataset freely, since there is many changes in the deep_mimic env, I'll make another pull request. @erwincoumans it seems there has been a huge progress on transferring the DeepMimic, could you please tell me which state we're in right now, I want to know if I can help with this.

Somedaywilldo:
@erwincoumans new version is #2102

Issue #2032, opened 2018-12-21
Heightfield and raycast performance


Zylann:
I came across this issue in Godot engine, which uses Bullet for its heightmap shape: https://github.com/godotengine/godot/issues/24390

I didn't see any issue open about this on Github so I opened one. 
Details are in the ticket linked above, but in short: I found raycasting a heightfield is tremendously slow. When you use picking raycasts (which are usually very long, think about sniper kind of raycast), performance goes down to 1 FPS with 2049x2049, or 2 FPS on 1025x1025. I also tried 513x513 and causes frame drops too, and this is on a 3.4GHz Ryzen CPU.

The reason is, it seems rays are enclosed in a AABB which is then used to query every single triangle lying inside it. Tell me if I'm wrong, but this is highly suboptimal for a collider of this kind, to the point even running a supernaive method like [this one](https://github.com/Zylann/godot_heightmap_plugin/blob/master/addons/zylann.hterrain/hterrain.gd#L1056) outperforms it, not even using C++ (i.e advancing a point by small amounts until it is under the terrain).

I'm not be the first one who bumped into this, so I saw various workarounds in a few old forum posts, such as chunking the heightfield into many smaller ones (but only if you know you need raycasts...), or chunking the ray in many smaller ones (but only if you know what you raycast first...). Both these solutions are impractical, it's something we *systematically* would have to do as a special case, and kind of defeats the point in having a heightfield shape in the first place.

Another method would be to represent the heightfield with many chunked AABBs enclosing chunks, rather than a single huge one, without touching the underlying data grid, however that would have to happen internally.

Is there any plan to improve this?

erwincoumans:
Thanks for the report. It should be easy to add some acceleration structure indeed, for example using the btDbvt dynamic aabb tree. Perhaps there is already a pull request or patch somewhere.

>> Is there any plan to improve this?

This project relies purely on volunteers adding/fixing things and contributing patches. If you have time, consider adding the feature.

erwincoumans:
If you volunteer improving it, a pull request if welcome.
Here is some old discussion too:
https://github.com/bulletphysics/bullet3/issues/73
Thanks!

Zylann:
Cool, thanks for digging it up :) I'll have a look when I get some time.

Zylann:
It looks like the old PR got archived, and files I can download are not patches, just the whole files as they were years ago... quite hard to find out what actually happened

Zylann:
[...]

PR #2062, opened 2019-01-16
Accelerated terrain raycast with "Bresenham" and chunks

Zylann:
I moved my implementation from Godot to a Bullet fork, so I can create this PR.

This implements raycast acceleration for `btHeightfieldTerrainShape`, in order to fix https://github.com/bulletphysics/bullet3/issues/2032.
This approach uses a kind of grid traversal (which is similar but not really Bresenham) in order to only execute triangle intersection on heightfield cells that come in contact with the planar projection of the ray.
This allows results to come up ordered, and improves performance even without building any data structure.

Optionally, an accelerator data structure can be built by calling `buildAccelerator()`: this creates a higher-level grid of chunks encompassing the minimum and maximum heights. This structure can be traversed using the same algorithm in a hierarchical fashion, speeding it up on larger terrains. If rays are near vertical, shortcuts are taken to avoid overhead, so it mostly benefits to "line of sight" rays.

Status:
- I tested it in the Godot game engine on a random noise terrain 2049x2049 with pretty good results.
- The data structure is only one level of chunks at the moment, but recently I haven't had much time to dedicate further improvements to it. If a hierarchy is to be introduced, it would probably be done using more grids of lower LOD rather than a tree.
- Only implemented for Y axis
- The data structure could be useful for general collision too, but didn't work on that.
- The algorithm currently doesn't stop on first hit, so it rather does a linecast. This is mostly because I followed the same pattern of using a `btTriangleCallback`, but this doesn't tell if there was an intersection.
- Although I was testing in Godot, `btHeightfieldTerrainShape` needs to be ported to ExampleBrowser so we can test it from the SDK.

### DIFF
--- a/src/BulletCollision/CollisionDispatch/btCollisionWorld.cpp
+++ b/src/BulletCollision/CollisionDispatch/btCollisionWorld.cpp
@@ -22,6 +22,7 @@ subject to the following restrictions:
 #include "BulletCollision/CollisionShapes/btSphereShape.h"                 //for raycasting
 #include "BulletCollision/CollisionShapes/btBvhTriangleMeshShape.h"        //for raycasting
 #include "BulletCollision/CollisionShapes/btScaledBvhTriangleMeshShape.h"  //for raycasting
+#include "BulletCollision/CollisionShapes/btHeightfieldTerrainShape.h"     //for raycasting
 #include "BulletCollision/NarrowPhaseCollision/btRaycastCallback.h"
 #include "BulletCollision/CollisionShapes/btCompoundShape.h"
 #include "BulletCollision/NarrowPhaseCollision/btSubSimplexConvexCast.h"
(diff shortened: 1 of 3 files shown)


erwincoumans:
Thanks for the work on this. I'll give you some feedback once I have the chance to look at it.

slapin:
Need this one, will this PR eventually go in or there are problems with it?
Thanks!

erwincoumans:
Will look into this today, need to add some example in the ExampleBrowser and try it out.

erwincoumans:
The parallel version on a large btHeightfieldTerrainShape runs indeed much faster:
image

I added an example and a way to disable the accelerator, so I'll point to a new PR.

image

erwincoumans:
See https://github.com/bulletphysics/bullet3/pull/2204

erwincoumans:
#2204 is applied now.

Zylann:
Awesome! Thank you for the detailed profiling insight as well :)

glebm:
I'd love to see this TODO addressed:

https://github.com/bulletphysics/bullet3/blob/abea1a848411cf53385fb8288c89db05e5751ef7/src/BulletCollision/CollisionShapes/btHeightfieldTerrainShape.cpp#L336

`btHeightfieldTerrainShape::processAllTriangles` is 15% of the total frame time in openmw. /cc @psi29a

PR #2063, opened 2019-01-17
Update the installation of the bullet_robotics package

MaximilienNaveau:
- Create a bullet_robotics.pc for the bulletRobotics extras package.
- Install all necessary includes in include/bullet_robotics.
- The .pc adds directly the bullet_robotics folder as include directory.
  This is due to some relative include in the examples.


erwincoumans:
Thanks for the contribution! At some stage the re-usable BulletRobotics parts may move into a proper library. For now, this is an acceptable solution. We primarily use Python / PyBullet in my group, so we don't use the BulletRobotics C/C++ directly (nor cmake).

MaximilienNaveau:
Thank you very much

PR #2071, opened 2019-01-23
Applying collision filters for all physic clients in pybullet_setCollisionFilterPair and pybullet_setCollisionFilterGroupMask

### LINKED ISSUE #2070: Only physics client 0 taken into account for pybullet_setCollisionFilterPair and pybullet_setCollisionFilterGroupMask
Hi,

I have several simulation instances in DIRECT mode, with a robot in each. When I load the robot into the simulation, I apply some collision filters in order to avoid specific links to self collide. I noticed those filters are correctly applied on the physics client 0, but not on the others...

The problem seems to be coming from the method pybullet_setCollisionFilterPair (the pybullet_setCollisionFilterGroupMask method has the same problem) in examples/pybullet/pybullet.c,
```c 
sm = getPhysicsClient(physicsClientId);
```
Is called before parsing the arguments, physicsClientId is by default set to 0, so the filter is always applied to the client 0.

Thanks for that wonderful simulator !

Pandhariix:
This pull request aims to fix #2070 

The pybullet_setCollisionFilterPair and pybullet_setCollisionFilterGroupMask will take into account the physicsClientId parameter passed.

### DIFF
--- a/examples/pybullet/pybullet.c
+++ b/examples/pybullet/pybullet.c
@@ -6293,17 +6293,17 @@ static PyObject* pybullet_setCollisionFilterGroupMask(PyObject* self, PyObject* 
 
 	static char* kwlist[] = {"bodyUniqueId", "linkIndexA", "collisionFilterGroup", "collisionFilterMask", "physicsClientId", NULL};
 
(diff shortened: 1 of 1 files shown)


erwincoumans:
Well spotted, thanks for the fix!

Issue #1504, opened 2018-01-10
Unable to load color for MJCF files


ethanluoyc:
Hi,

It appears that when using the `pybullet` API, the colour for the models in MJCF files are not loaded properly. Right now, the colors are somehow picked randomly. I am not sure how this can be fixed.

To reproduce, simply run the examples. If I change the model files in mjcf folder, there will not be any effect if I change the color.

Thanks!

benelot:
It is not extracted for now. And they are indeed picked randomly. That is
all I know.

ethanluoyc:
@benelot That's good to know.

After some digging, it appears that the importing happens here https://github.com/bulletphysics/bullet3/blob/master/examples/Importers/ImportMJCFDemo/BulletMJCFImporter.cpp#L709. It appears that rgba has been parsed. I am not entirely sure why after this the color is not set up. Maybe I am missing something. If someone can point me to the right direction I can probably make a PR and get this fixed.

erwincoumans:
At the moment, the MJCF parser doesn't create visual shapes. This means, there is an autogeneration of visual shapes, using the Bullet collision shapes that don't have color.

I have some early patch that creates visual shapes, with color, but don't have texture coordinates, so it doesn't look nice.

Also, I have an early patch that adds planar reflection, but it needs to use the stencil buffer.

Until then, you can use 
~~~~~~~~~~
p.changeVisualShape(sphere,linkIndex,rgbaColor=[red,green,blue,1])
~~~~~~~~~~
for each link, to assign colors.

ethanluoyc:
@erwincoumans Thanks! I am working with RL with image observations of the environment. To me, the texture becomes less important but they have to be distinct from the environment. Would you be able to point to me where the Google colours are instantiated in the codebase? I am trying to understand the how pybullet works right now. I will like to be able to extend it in the long term :D

erwincoumans:
See this pull request, once merged, use the option 
~~~~~~~~~
 p.loadMJCF(filename, flags=p.URDF_MJCF_COLORS_FROM_FILE
~~~~~~~~~
When using the option, it uses a highly tesselated spheres/capsules without instancing. We can optimize that later.

ethanluoyc:
Thanks! @erwincoumans

PR #2093, opened 2019-02-01
Added fetch mjcf model

josiahls:
Added the loading line in the MJCF setup to load the fetch model
Added a 'camera' skip message (and reason why) instead of warning.

You can now run the example window app, and the first model that gets loaded during the MJCF setup will be the fetch (however it does not have a base, so it falls, but still loads)

Notes:
The goal was to see how colors are being loaded from MJCF files. It does not seem like they get loaded at all regardless of #1504 . What currently needs to be done to get colors to load? I understand that I need to look at the geom tag and how that is being handled, but I am not sure even if I log materials, that they will load correctly since you said there isn't support for the visual meshes yet? Also, I did this because the flags for the MJCF 'works', but all that happens is the google colors dont get applied so the meshes are all just white. I have successfully gotten the URDF version working perfectly (textures/colors :) ), but MJCF robots and envs do not load colors period.


erwincoumans:
Thanks for the feedback. Is the URDF version of fetch in your PR as well?

There isn't much development in MJCF parser. If URDF_MJCF_COLORS_FROM_FILE flag doesn't work, someone has to digg deeper.

What is the license of the Fetch robot, did you include a license file?

erwincoumans:
Without license file we cannot accept the contribution. Closing.

PR #2102, opened 2019-02-10
For pybullet_envs/deep_mimic, add support for a simplified Human3.6M dataset

Somedaywilldo:
Dear @erwincoumans ,

Now the deep_mimic env has the support for a popular humanoid dataset, Human3.6M, although this is not the official version of Human3.6M but a prepossessed one, as the former requires manual permission by the creators to download. However this version is easy to use, and free to download, hope this will help. And there is no torch dependency now : ) .

Somedaywilldo


Somedaywilldo:
@erwincoumans  You can try it by running:

python3 render_reference.py --dataset_path=\<path to data_3d_h36m.npz\> \\
	  --subject=S11 \\
	  --action=Walking \\
	  --json_path=\<path to save Walking.json\> \\
	  --fps=24 \\
	  --loop=wrap \\
	  --draw_gt

**data_3d_h36m.npz** can be downloaded at [here](https://www.dropbox.com/s/z5bwig0h6mww590/data_3d_h36m.npz?dl=0)
And it should be look like this [video](https://youtu.be/goew_FmUtOE)

erwincoumans:
That's great, thanks for working on this. I am still tracking down an issue that prevents the transfer from the original trained policy to the PyBullet version. Once that is working, we can re-implement the DeepMimic to train much faster, using a large time step (120 Hz). I'll get back when this is sorted.

erwincoumans:
Here is a video, still needs a bit more work (it still falls eventually and speed needs improvements, the video is 4x sped up: https://www.youtube.com/watch?v=PsE48F39Kkw&feature=youtu.be

Somedaywilldo:
Thanks! Let me see if I can help with that.

erwincoumans:
Latest version should perform the backflip at least once or twice.
Can you move all your files (including readme, ik etc) into a 'mocap' subdirectory, under pybullet_envs.deep_mimic?
https://github.com/bulletphysics/bullet3/tree/master/examples/pybullet/gym/pybullet_envs/deep_mimic

Somedaywilldo:
Yes, it's compatible, many joints are removed, the number of joints changed from 32 to 17, and after that, shoulders joint's link parents are reseted to chest joint. I've tried many motions in the dataset, should be OK.

I see many updates in the humanoid module, I'll try to make it compatible with the PD control humanoid.

erwincoumans:
It is odd, this PR has 40 changed files, some git merge issue? Could you create a new PR with only the added/modified files under Bullet/examples/pybullet/gym/pybullet_envs/deep_mimic?

Somedaywilldo:
OK, will do it right away.

Somedaywilldo:
See #2105 @erwincoumans

erwincoumans:
We merge https://github.com/bulletphysics/bullet3/pull/2105, so this can be closed.

erwincoumans:
I finally tried it out, and it works well with Python 3.
The debug draw --draw_gt could be made much faster likely, by re-using the same debug lines (it is an option), but it is not important.
Thanks!

PR #2103, opened 2019-02-11
PyBullet deep_mimic backflip re-using original DeepMimic policy from Jason Peng

erwincoumans:
(still falls over after second backflip, due to some differences in implementation)
~~~~~~~~~~~~~~~~~
pip install pybullet
python -m pybullet_envs.deep_mimic.testrl --arg_file run_humanoid3d_backflip_args.txt
~~~~~~~~~~~~~~~~~
Or in the local repository (make sure pybullet_data, pybullet_utils have symbolic links to the folders in your repo location)
~~~~~~~~~~~~~~~~~
python testrl.py --arg_file run_humanoid3d_backflip_args.txt
~~~~~~~~~~~~~~~~~


Somedaywilldo:
I'm also trying to run your code for backflip, I'm wondering why the agent couldn't even finish the first backflip, not as your video shows, but seem to flip more than one cycle, is that due to the configuration? 
https://youtu.be/uZp8kgb_cF0

PR #2104, opened 2019-02-12
Adds extra information about the UserData in the notifications.

RanTig:
When using the USER_DATA_ADDED and USER_DATA_REMOVED notifications, some
more information is necessary than just the user data id, especially
when a user data entry has been removed, it is valuable to know the key and body identifier of the removed entry.


erwincoumans:

PR #2111, opened 2019-02-15
Stub test for deterministic save and restore

mihan69rus:
Write test


erwincoumans:
There is a PyBullet test for deterministic save and restore here:
https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/unittests/saveRestoreStateTest.py
It can simulation, save the state, continue the simulation and compare with loading the state and continue simulating, make sure the results are identical.

PR #2120, opened 2019-02-20
Refactor pybullet_envs.__init__

kngwyu:
Thanks for great works here.
Since this file is the entry point for gym users, I want it well-formatted.
BTW, is it OK to send a PR that forces the same indent style to python codes?
Some files(e.g. https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/scene_abstract.py) contains both tabs and spaces for indent, and I don't like that.

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/__init__.py
+++ b/examples/pybullet/gym/pybullet_envs/__init__.py
@@ -1,25 +1,27 @@
 import gym
-from gym.envs.registration import registry, make, spec
-def register(id,*args,**kvargs):
-	if id in registry.env_specs:
-		return
-	else:
-		return gym.envs.registration.register(id,*args,**kvargs)
+
+
+def register(id, *args, **kvargs):
+    if id in gym.envs.registry.env_specs:
+        raise RuntimeError('{} is already registerd'.format(id))
+    else:
+        return gym.envs.registration.register(id, *args, **kvargs)
 
 # ------------bullet-------------
 
+
 register(
-        id='HumanoidDeepMimicBulletEnv-v1',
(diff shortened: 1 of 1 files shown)


erwincoumans:
Formatting changes are best done automatically. At the moment, the C++ code uses tabs, and it uses this [clang_format file](https://github.com/bulletphysics/bullet3/blob/master/_clang-format) 
For Python we can use spaces, like Google style.
Can we add a Python section in there, with Google Style? Then I'll run that clang format on my end.

kngwyu:
Thanks, I'll try it.

kngwyu:
As an experiment I formatted whole python files by yapf by [google style](https://github.com/kngwyu/bullet3/pull/1) and [pep8 style](https://github.com/kngwyu/bullet3/pull/2).
It looks the only diff is [google style inserts a empty line](https://github.com/kngwyu/bullet3/blob/af50a4b3a05960be7086376ffa6f7559e2faae2a/data/xacro_standalone.py#L102) before __init__.

erwincoumans:
So can I use the yapf tool instead of applying this PR? If so, what settings do I use?

kngwyu:
I like [this setting](https://github.com/kngwyu/bullet3/blob/5a1a910d92611d7f0fc4bc5af7e1ae24c629ea43/.style.yapf) but have no strong opinion.

erwincoumans:
If we apply a style to Python files, it will be Google style. I think it doesn't use 4 spaces, does it?

kngwyu:
I tried 
```
[style]
based_on_style = google
```
and looks it uses 4 spaces by default.

kngwyu:
Closed in favor of #2192

PR #2122, opened 2019-02-20
Limit compilation parallelism

waveform80:
On smaller platforms like the Raspberry Pi, which have a deficit of physical memory, the default algorithm for determining compilation parallelism (2*cores) is far too optimistic and usually results in crashing the platform.

This modification limits the number of parallel jobs to the size of RAM in Gb. On a mid-size PC (e.g. quad-core with 8+Gb of RAM), this will make no difference. On a Pi (quad-core with 1Gb of RAM), this will limit it to 1 job at a time (which is ideal).

Some background to this: I'm one of the maintainers of [piwheels](https://piwheels.org/) which attempts to build pre-compiled packages from PyPI for the Raspberry Pi. Currently, the pybullet package has a tendency to crash our python 3.5 builders (the 3.4 builders *just* manage it because the gcc version Raspbian Jessie is a bit less memory-hungry). I've marked the current version to be skipped, but it'd be nice to build future versions once the setup.py is a little less ... enthusiastic!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,18 @@ def parallelCCompile(self, sources, output_dir=None, macros=None, include_dirs=N
     cc_args = self._get_cc_args(pp_opts, debug, extra_preargs)
     # parallel code
     N = 2*multiprocessing.cpu_count()# number of parallel compilations
+    try:
+        # On Unix-like platforms attempt to obtain the total memory in the
+        # machine and limit the number of parallel jobs to the number of Gbs
+        # of RAM (to avoid killing smaller platforms like the Pi)
+        mem = os.sysconf('SC_PHYS_PAGES') * os.sysconf('SC_PAGE_SIZE') # bytes
+    except (AttributeError, ValueError):
+        # Couldn't query RAM; don't limit parallelism (it's probably a well
+        # equipped Windows / Mac OS X box)
+        pass
+    else:
+        mem = max(1, int(round(mem / 1024 ** 3))) # convert to Gb
+        N = min(mem, N)
     def _single_compile(obj):
         try: src, ext = build[obj]
         except KeyError: return


erwincoumans:
Thanks for the report/PR.
I will check it on my desktop Linux versions (make sure it doesn't degrade to 0 there).
Can you only perform the check for Linux platforms by adding:
~~~~~~~
if _platform == "linux" or _platform == "linux2":
~~~~~~~

waveform80:
> Can you only perform the check for Linux platforms by adding:
>
> if _platform == "linux" or _platform == "linux2":

I could add that test, but frankly that's what the `try..except` block is for. Ideally, I'd apply the algorithm to all platforms (because a Windows/Mac/BSD box with a single gig of RAM is equally as limited) but I don't know a simple cross-platform mechanism of determining physical memory without introducing additional dependencies (which would be overkill for something which only applies to `setup.py`).

Hence, I figure it's better to *attempt* the limit on all platforms, catching the exceptions that occur if `os.sysconf` is unavailable (`AttributeError`, which will be the case on non-Unix platforms), or if the required system configuration variables are unavailable (`ValueError`, which *may* be the case on the BSD platforms or for that matter ancient Linux platforms). In other words: better to enforce the limit wherever possible (without arbitrarily limiting it to a particular platform).

Incidentally, the `max(1, ...)` call should ensure things never degrade to 0 unless `multiprocessing.cpu_count()` returns 0 (in which case the current code would use 0 anyway).

erwincoumans:
I did some quick tests on my Linux, Mac OSX and Windows machines.
On Windows, the parallel compile doesn't seem to be used at all. On Linux and Mac desktop, N doesn't seem affected indeed. Let's merge it. Thanks for your contribution.

PR #2129, opened 2019-02-27
Fix crash in btGImpactMeshShape vs btBvhTriangleMeshShape

dictoon:
This PR fixes a crash encountered when rigid bodies with `btGImpactMeshShape` shapes collide with rigid bodies with `btBvhTriangleMeshShape` shapes.

### DIFF
--- a/src/BulletCollision/Gimpact/btGImpactCollisionAlgorithm.cpp
+++ b/src/BulletCollision/Gimpact/btGImpactCollisionAlgorithm.cpp
@@ -18,7 +18,7 @@ subject to the following restrictions:
 3. This notice may not be removed or altered from any source distribution.
 */
 /*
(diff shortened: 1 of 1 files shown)


erwincoumans:
Haven't seen btGimpactShape for a while, we always use convex decompositions (V-HACD) btConvexHull+btCompoundShape instead.
Probably good to recover an old demo and move it into the example browser.
Thanks!

dictoon:
I should give V-HACD a try, thanks for the tip.

That said, I'm positively surprised by the stability and performance of GIMPACT, especially in the `btGImpactMeshShape` vs `btGImpactMeshShape` case, which is _much_ faster than `btGImpactMeshShape` vs `btBvhTriangleMeshShape`.

Issue #2131, opened 2019-02-28
[pybullet] rayTestBatch hit with parentObjectUniqueId and debugLine but don't hit without debugline


mcaniot:
I get an issue with rayTestBatch. I attached my ray to my robot id and use addUserDebugLine for debugging. The ray found a hit whith this configuration. But when I don't use the function addUserDebugLine, the ray don't hit in the same configuration.

I found a workaround : I create a user debugline attached to my robot and I remove it just after. The ray hit in the same configuration as before.

erwincoumans:
Thanks a lot for the report. The PR with a fix has been applied.

PR #2132, opened 2019-02-28
Retrieve rayTestBatch data without having to display debug lines

Pandhariix:
Fixing the rayTestBatch issue #2131

With the proposed fix, no need to use our workaround (creating a user debugline attached to the robot and removing it just after) to get correct ray test batch data.

The PhysicsServerCommandProcessor will use the correct structure __m_requestRaycastIntersections__ in the __processRequestRaycastIntersectionsCommand__ method


erwincoumans:
Good find, that is indeed a bug. The code would be less fragile by passing in the relevant structure for each command (clientCmd.m_requestRaycastIntersections in this case), instead of the full clientCmd. At some stage, I'll look into fixing that.
Thanks a lot!

PR #2142, opened 2019-03-06
Fixed joint

fuchuyuan:
I created an example joining a row of boxes by joints. p2p joints looks ok. But when I use fixed joint or prismatic joint, the boxes don't appear in the sim. I am only adding in the cubes and joining them, there is no gravity for now.


erwincoumans:
Looks good. Just replace b3JointInfo jointInfo;
with	
~~~~~~~~~~~~
b3RobotJointInfo jointInfo;
~~~~~~~~~~~~
and it should work: b3JointInfo is a c structure (unintialized fields) and b3RobotJointInfo  has a constructor.

PR #2145, opened 2019-03-08
Add example

fuchuyuan:
I added three examples in ExampleBrowser to show cases where the default SI solver does not converge well.

1. A stack of blocks of large mass ratio would fall down.
2. Three blocks connects by joints. Joint limits would be exceeded due to numerical error.
3. A row of boxes connected by fixed joints would bend due to numerical error.


erwincoumans:

PR #2161, opened 2019-03-14
Adds a timestamp as a physics parameter.

crewmatt:


erwincoumans:
The patch includes internal path changes from Google and also whitespace changes, we use tabs not spaces. In addition, there is some nullptr check
~~~~~~~~~~~
status->m_sendActualStateArgs.m_stateDetails != nullptr
~~~~~~~~~~~
Was that causing a crash? If so, how to reproduce this.
We don't use C++11 in Bullet, can you replace nullptr by NULL or 0?
Thanks!

crewmatt:
Updated the formatting and pulled files direct from bullet not Google. 

The null check causes a hard crash if you reset a simulation and ask for a link without loading any objects. This is one of the tests in our system. Not super important if you prefer we don't have it, just figured it was a better failure semantic then a seg fault.

erwincoumans:
Great, that makes sense. Thanks for the contribution!

PR #2169, opened 2019-03-22
Add __version__ information for pybullet

iory:
I would like to know the version information from ```__version__```.
This PR adds version information for pybullet and we can access version like the following code.

```
In [1]: import pybullet
pybullet build time: Mar 22 2019 18:28:58

In [2]: pybullet.__version__
Out[2]: '2.4.8'
```

### DIFF
--- a/examples/pybullet/pybullet.c
+++ b/examples/pybullet/pybullet.c
@@ -10825,6 +10825,8 @@ initpybullet(void)
 	if (m == NULL) return;
 #endif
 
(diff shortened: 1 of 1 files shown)


erwincoumans:
This will be out-of-sync immediately. Try using this, assuming you used pip/pip3:

~~~~~~~~~~
>>> import pkg_resources
>>> pkg_resources.get_distribution("pybullet").version
~~~~~~~~~~

PR #2171, opened 2019-03-23
[pybullet/gym/pybullet_envs] Modified register's timestep_limit to max_episode_steps for gym

iory:
From `openai/gym version 0.7.4-alpha`, `timestep_limit` is deprecated.
This PR modified it to `max_episode_steps`.

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/__init__.py
+++ b/examples/pybullet/gym/pybullet_envs/__init__.py
@@ -25,14 +25,14 @@ register(
 register(
     id='MinitaurBulletEnv-v0',
     entry_point='pybullet_envs.bullet:MinitaurBulletEnv',
-    timestep_limit=1000,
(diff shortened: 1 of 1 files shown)


erwincoumans:
Thanks for the PR. Does this mean our environments are broken with older Gym versions?

iory:
> Does this mean our environments are broken with older Gym versions?

Not broken because `openai/gym` warns only for backward compatibles.
https://github.com/openai/gym/blob/master/gym/envs/registration.py#L54-L57

It is bothersome that warning is given and some `register` use `max_episode_steps` in pybullet_env.
https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/__init__.py#L11-L16

So we should use `max_episode_steps` to suppress warnings.

erwincoumans:
OK, thanks for the contribution.

PR #2174, opened 2019-03-25
Use two SI solvers as blocks in block solver.

fuchuyuan:
In the btBlockSolver we are experimenting with, we have SI for both multibody and rigid body. I'm currently replacing rigid body SI solver with two smaller SI solvers. The two examples provided by RigidBodyBoxes.h should have the same behavior.


erwincoumans:
Thanks for the pull request. It is best to remove the white space changes in the PR.

fuchuyuan:
Hi Erwin, I used the clang format tool in bullet to format my changes. Is that ok?

erwincoumans:
I try to keep the commits with actual changes fully separate from commits that contain white-space changes (manual or clang-format). I'll look into it later.

Issue #1976, opened 2018-11-02
EGL + changeVisualShape segfault


BlGene:
If I delete an object A, create a replacement B and try to call `changeVisualShape` for B it causes a segfault.

see: https://gist.github.com/BlGene/afa726e7364db6701e1793c3a85ebcd2 <-- search `#TODO():`

erwincoumans:
Well, the eglPlugin is your contribution. Can you help fixing the issue?

BlGene:
Yes, I can have a look. I'm a bit surprised it works for the debug viewer though.

BlGene:
closing with #2176

PR #2176, opened 2019-03-26
[WIP] EGL fixes: change color & device selection 

BlGene:
Addresses #1976

I'm not sure if we still need/want m_swRenderInstances/TinyRenderObjectData. I could remove them if not.

### DIFF
--- a/examples/OpenGLWindow/EGLOpenGLWindow.cpp
+++ b/examples/OpenGLWindow/EGLOpenGLWindow.cpp
@@ -141,7 +141,12 @@ void EGLOpenGLWindow::createWindow(const b3gWindowConstructionInfo& ci)
 				if (eglGetError() == EGL_SUCCESS && initialized == EGL_TRUE)
 				{
 					m_data->egl_display = display;
+					break;
 				}
+			}
+			else
+			{
+				fprintf(stderr, "GetDisplay %d failed with error: %x\n", i, eglGetError());
 			}
 		}
 	}
@@ -165,6 +170,10 @@ void EGLOpenGLWindow::createWindow(const b3gWindowConstructionInfo& ci)
 			{
 				m_data->egl_display = display;
 			}
+		}
+		else
+		{
+			fprintf(stderr, "GetDisplay %d failed with error: %x\n", m_data->m_renderDevice, eglGetError());
 		}
 	}
 


BlGene:
On a related note the option strings(e.g.  `cid = p.connect(p.DIRECT, options=optionstring)`) are no longer being parsed by EGL.

BlGene:
One can now define `EGL_VISIBLE_DEVICES` as environment variable to select rendering device.

erwincoumans:
It is best to submit the smallest possible patch, without refactoring etc, if you want to get a pull request accepted. Probably restart from the latest revision, to avoid merge conflicts and add one thing at a time?

BlGene:
Yes, of course. The PR is not done yet and I'll do a rebase. I just wanted to post this in the meantime to save others the hassle of figuring it out themselves.

I was having some issues with the egl render perspective being different in a current branch (see images, I'm not yet sure why and if its potentially only the default perspective), so I wanted to get to the bottom of that first.

my branch:
original

current branch:
current

alexpashevich:
I have also encountered both of those issues. I modified examples/OpenGLWindow/EGLOpenGLWindow.cpp to look at $CUDA_VISIBLE_DEVICES but an egl-specific variable might be a better solution.

BlGene:
>Pybullet 2.4.0 seems to be more stable with EGL.

Did you notice any other problems than the ones described above? If so could you please be more specific, or ideally provide a test script showing the problem(s)?

alexpashevich:
@BlGene no, the only problem I have noticed was the one you described (the perspective). However, I can not be absolutely confident because I downgraded to 2.4.0 after noticing it.

muupan:
Is the perspective issue related to https://github.com/bulletphysics/bullet3/issues/2390?

buoyancy99:
Is the EGL device selection finished?

BlGene:
@buoyancy99 

More or less yes.

There are some rough edges if a specific GPU needs to be selected on a multi-GPU machine. I'm not sure if specifying `m_renderDevice` from the plugin option string works as intended and I had a different solution parsing the environment variable "EGL_RENDER_DEVICE" at some point. But people don't seem to have any issues with this at the moment.

BlGene:
@buoyancy99 

Here is a slightly improved  version : https://github.com/BlGene/bullet3/tree/egl

This version has the following two changes:
1. you can select the EGL device you want to use by setting the environment variable EGL_RENDER_DEVICE=X. Just start at 0 and iterate up until it tells you "Invalid render_device choice"
2. It requests an EGL context with OpenGL version  > 3.3, which is needed for the EGL plugin. On the machines I was using so far EGL was nice and gave me up to date OpenGL versions, but this was needed for my Intel HD / Mesa laptop.

If you have some time you could test out if it works for you, in which case I would turn it into a PR.

P.S.  If you want to play around with EGL I find this quite useful for development: https://github.com/BlGene/egl_check

buoyancy99:
Great! This is very helpful

I am also trying to implementing egl render of deformable cloth for pybullet but has made no progress so far. Started an issue [here](https://github.com/bulletphysics/bullet3/issues/2951) yet it seems egl cannot reuse the opengl renderer fix in [this](https://github.com/bulletphysics/bullet3/pull/1555) PR. Do you have any suggestions on how to access the cloth data in egl plugin and draw?

BlGene:
@buoyancy99

I have only looked at at this briefly, so take these suggestions with a grain of salt, but I would look at the following approaches:
A) Try if you can get OpenGLGuiHelper.cpp to send the updateShape to the EGL GLInstancingRenderer and making sure that GLInstancingRenderer handles this information.
B) Replicate the functionality that is in OpenGLGuiHelper.cpp in PhysicsServerCommandProcessor.cpp and send the updateShape to the EGL GLInstancingRenderer and making sure that GLInstancingRenderer handles this information.

PR #2177, opened 2019-03-26
Initializes simulation timestamp to 0 explicitly.

crewmatt:


erwincoumans:
You can't add 'm_simulationTimestamp(0);' in the initializer list.

Issue #2189, opened 2019-04-07
pybullet crashes when used in Blender scripts


JayFoxRox:
I'm unable to use pybullet in Blender scripts.

I'm running:

- Arch Linux (x64)
- pybullet 2.4.8 (latest on PIP)
- Python 3.7.3 (latest on Arch Repos)
- Blender 17:2.79.b.git5.a29446da-2 (latest on Arch Repos)
- Blender 2.80.190326.adfdae3fc2f4-1 (latest on Arch User Repos)

To reproduce, just open a scripting area in Blender, and type:

```python
>>> import pybullet
>>> i = pybullet.connect(pybullet.DIRECT)
>>> pybullet.resetSimulation(i)
```

The last command will reliably crash Blender 2.79 (and the latest Blender 2.80 builds).
The same set of commands work fine in stand-alone Python on the same machine.

Some other pybullet functions (other than `resetSimulation`) also result in crashes. Other connection methods don't solve the problem either.

---

The crash I'm getting has the following backtrace:

```
$ gdb --args blender --python-expr "import pybullet; i = pybullet.connect(pybullet.DIRECT); pybullet.resetSimulation(i)"
GNU gdb (GDB) 8.2.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from blender...(no debugging symbols found)...done.
(gdb) r
Starting program: /usr/bin/blender --python-expr import\ pybullet\;\ i\ =\ pybullet.connect\(pybullet.DIRECT\)\;\ pybullet.resetSimulation\(i\)
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[New Thread 0x7fffe39ff700 (LWP 6185)]
[New Thread 0x7fffe24fe700 (LWP 6186)]
[...]

PR #2190, opened 2019-04-08
Reduce visibility of Bullet symbols in PyBullet

JayFoxRox:
Avoids version conflicts in programs using PyBullet, which also link against Bullet libraries.

This addresses #2189 for Linux - it will no longer crash, and pybullet works when imported in Blender 2.79 or 2.8; the command line shown below also works with this fix.

---

A simple way to reproduce the original issue on Linux is (which will crash):

```
LD_PRELOAD=/usr/lib/libBulletSoftBody.so python3.7m -c "import pybullet; i = pybullet.connect(pybullet.DIRECT); pybullet.resetSimulation(i)"
```

*(Where libBulletSoftBody.so and PyBullet are different versions / compilations of Bullet; also crashes with preloading libBulletDynamics.so, and possibly others)*

---

**Implementation notes**

* I assume this solution could be extended to MacOS and BSD (or other GNU-C compatible compilers).
* I did not fix / review the CMake and Premake installations of PyBullet.
* I'm not sure if this should also be extended for eglRenderer.
* The GCC 4.0 version check came from https://gcc.gnu.org/wiki/Visibility .
* `-fvisibility-inlines-hidden` might be redundant / implied by `-fvisibility=hidden` (I did not check).

I'm a bit surprised that this visibility isn't handled by pypi / setuptools / CPython / ... internally (using the `PyMODINIT_FUNC` macro and built-in flags). I might investigate further wether this solution could move upstream into those tools (whichever might be the responsible project).

An alternative solution would be to add versioning information to Bullet libraries (which should be done regardless). However, I feel for PyBullet this is the cleaner solution to address the issue at hand: it shouldn't expose the Bullet symbols.

One thing to keep in mind, is that this reduced visibility probably makes it impossible to link against PyBullet as a generic Bullet shared-library. However, I feel that's not a valid use-case, as people probably build Bullet the "standard" (CMake / Premake) way, and not through setup.py.

### DIFF
--- a/examples/pybullet/pybullet.c
+++ b/examples/pybullet/pybullet.c
@@ -10842,6 +10842,9 @@ static struct PyModuleDef moduledef = {
 };
 #endif
 
+#if __GNUC__ >= 4
+__attribute__((visibility ("default")))
+#endif
 PyMODINIT_FUNC
 #if PY_MAJOR_VERSION >= 3
 PyInit_pybullet(void)
--- a/setup.py
+++ b/setup.py
@@ -370,6 +370,8 @@ if _platform == "linux" or _platform == "linux2":
     CXX_FLAGS += '-DDYNAMIC_LOAD_X11_FUNCTIONS '
     CXX_FLAGS += '-DHAS_SOCKLEN_T '
     CXX_FLAGS += '-fno-inline-functions-called-once '
+    CXX_FLAGS += '-fvisibility=hidden '
+    CXX_FLAGS += '-fvisibility-inlines-hidden '
     EGL_CXX_FLAGS += '-DBT_USE_EGL '
     EGL_CXX_FLAGS += '-fPIC ' # for plugins
 


erwincoumans:
Thanks for tracking down your issue with Blender+PyBullet.

We do dynamically load PyBullet.so as part of our experimental Unity plugin. For that plugin, we only need to expose a singer header file, the C-bindings here:
https://github.com/bulletphysics/bullet3/blob/master/examples/SharedMemory/PhysicsClientC_API.h
Since we compile that plugin with cmake or premake, it probably isn't affected.

This __attribute__((visibility ("default"))) only affects that single pybullet.c file, right?

JayFoxRox:
The change in setup.py hides *all* symbols / across all compilation units.

The attribute in the C file turns it on per symbol; in this case, I've only marked the entry-point used by CPython (`PyInit_pybullet`), but no other functions / variables / .. - if there's any other functions which Python / Unity must see, they have to be marked, too.

Any function that's already using this macro (which appears to be all of them in the linked file?) should still be visible:

https://github.com/bulletphysics/bullet3/blob/c243f9e7c70c6fd3ab5991abdd1fc652b3bd1156/examples/SharedMemory/PhysicsClientC_API.h#L20

I don't have Unity myself, so I can't test it - but it should still work fine (even if built through setup.py).

When I wrote my patch I wasn't aware of existing code using visibility attributes in GCC; otherwise I'd also just have used `#if defined (__GNUC__)` for consistency (even if it will break for GCC below version 4.0). Shall I change my code?

erwincoumans:
No, we can change it in the other place. Are people still using older gcc versions (<=3)?

PR #2192, opened 2019-04-08
Format python codes

kngwyu:
Concretely, I added `.style.yapf` and run `yapf -r -i ./`.
I used [yapf](https://github.com/google/yapf) 0.27.0.
The setting is 
```
[style]
based_on_style = google
column_limit = 99
indent_width = 2 
```
.

I believe this would be useful for contributors.


erwincoumans:
I tried running the yapf tool myself using your .style.yapf, but it runs REALLY slow. Probably better to find a better solution. Thanks a lot for looking into this though. Once we have a reasonably fast solution, I like to run it myself, so I can be 100% sure the PR only contains formatting changes.

kngwyu:
I see, thanks.

erwincoumans:
Let me check what the best option is, I still want to have the Python formatting consistent.

erwincoumans:
Re-created this PR with some changes to make yapf run faster, see 
https://github.com/bulletphysics/bullet3/commit/ef9570c315ee0db1dffbb950dbd46aa726e06f83

erwincoumans:
This CL is used instead. Thanks for the contribution!
https://github.com/bulletphysics/bullet3/pull/2218

PR #2195, opened 2019-04-09
don't need to call the computeAccelerationsArticulatedBodyAlgorithmMu…

Steven89Liu:
…ltiDof() the second pass if the m_internalNeedsJointFeedback is false.


erwincoumans:
Thanks for the improvement, looks good to me.

PR #2200, opened 2019-04-13
Fix render issue in server render mode (e.g., docker)

YungKC:
px is an 1D array, and cause a runtime error when trying to strip the color channels into RGB. I fixed this by first reshape the px array into am image compatible format first.


erwincoumans:
Thanks for the contribution.

PR #2204, opened 2019-04-16
add raycast accelerator for btHeightfieldTerrainShape, thanks to Marc…

erwincoumans:
… Zylann, see https://github.com/bulletphysics/bullet3/pull/2062

it can be disabled by setting the flag cb.m_flags |= btTriangleRaycastCallback::kF_DisableHeightfieldAccelerator;
acceleration is disabled for z axis up.
fixed memory allocator todo, use btAlignedObjectArray so allocation goes through proper allocator
add btHeightfieldTerrainShape example to example browser


slapin:
Thanks a lot!

PR #2209, opened 2019-04-16
Should sort best directions that gives the highest rewards

YungKC:
Found bug where the sort order would incorporate the lowest N rewards, instead of the highest. With this fix, one can now use fewer best directions instead.


erwincoumans:
Thanks for the report, LGTM.

Issue #2219, opened 2019-04-28
build script should build a shared library of the C API


goretkin:
The build script produces a pybullet shared library. Could it also build a shared library of just the C API (`bullet3/examples/SharedMemory/PhysicsClientC_API.h`)?

erwincoumans:
That would be easy. What would you do with it?

goretkin:
I made an attempt at a PR that would solve this for my case.
I would like to use Bullet from Julia. The [build script for the Julia package builds the C++ API.](https://github.com/SimonDanisch/Bullet3Builder/blob/master/build_tarballs.jl#L46) I would like it to build the C API instead (or in addition) since it is easier to wrap the C API right now.

goretkin:
Closed by https://github.com/bulletphysics/bullet3/pull/2221

PR #2221, opened 2019-04-29
C api refactor build

goretkin:
An attempt at solving issue #2219


goretkin:
There is a lot of "pybullet" flags that are checked in the C API build. All those configurations apply equally to the C API build as far as I can tell. So I'm not sure if they should be renamed, or just left as is.

goretkin:
It looks like the test failure is spurious (some SSL timeout). I am not sure how to re-trigger it.

Is there anything in the way of getting this merged?

erwincoumans:
Thanks for the contribution. I think the 'c_api' is similar to Bullet/Example/BulletRobotics?
If it doesn't break anything, let's merge it.

goretkin:
Do you mean: https://github.com/bulletphysics/bullet3/tree/master/examples/BulletRobotics ?

Thanks for considering this merge. It will make it easier to wrap the project this way! Are you satisfied with the naming?

erwincoumans:
No, it is this cmake target: https://github.com/bulletphysics/bullet3/tree/master/Extras/BulletRobotics

The subtle thing is that BulletRobotics doesn't depend on OpenGL/graphics, to keep the dependencies low. There is also a rudimentary C++ api on top of that BulletRobotics c api, to make it more confusing.
https://github.com/bulletphysics/bullet3/tree/master/examples/RobotSimulator

See https://github.com/bulletphysics/bullet3/blob/master/examples/SharedMemory/b3RobotSimulatorClientAPI_NoDirect.h
and
https://github.com/bulletphysics/bullet3/blob/master/examples/SharedMemory/b3RobotSimulatorClientAPI_NoGUI.h

erwincoumans:
It seems your 'c_api' is a duplicate of the cmake BulletRobotics target in Extras/BulletRobotics.
If so, shall we remove it again, and make pybullet depend on BulletRobotics?

goretkin:
If it's a duplicate, then that seems like the way to go! Could we rename BulletRobotics to something else, in analogy to pybullet?

And ideally, there would be a way to compile it with and without OpenGL support.

erwincoumans:
best to revert it soon. It also breaks the build (we don't require OpenGL since we dynamically load it, not link it)

~~~~~~~~~~~~~~~~
OPENGL_gl_LIBRARY (ADVANCED)
    linked by target "bullet_c_api" in directory /home/travis/build/bulletphysics/bullet3/examples/c_api
OPENGL_glu_LIBRARY (ADVANCED)
    linked by target "bullet_c_api" in directory /home/travis/build/bulletphysics/bullet3/examples/c_api
~~~~~~~~~~~~~~~~

PR #2226, opened 2019-05-01
Soft body exp

fuchuyuan:
This PR does the following changes:
1. Turn on SoftDemo in ExampleBrowser
2. Adds setting startPosition and startOrientation options to b3RobotSimulatorClientAPI_NoDirect
3. Adds setting startPosition and startOrientation options to pybullet

PR #2231, opened 2019-05-03
set softbody start position and orientation in pybullet

fuchuyuan:
sorry missed two lines from https://github.com/bulletphysics/bullet3/pull/2226


erwincoumans:
Thanks, LGTM!

PR #2232, opened 2019-05-04
fixed build on latest version of Clang

EmilNorden:
Build fails on Clang 8.0.0 with:
```
argument value 10880 is outside the valid range [0, 255]
      [-Wargument-outside-range]
```


PokeMMO (commented):
~~Looks good to me~~. Wrong ticket... Please forgive me for spam. lol

PR #2233, opened 2019-05-06
Multibody memory invalid writing

AndreaCatania:
If you create a multibody with 0 links, and then later you resize it, the matrixBuffer is not resized and line of code like https://github.com/bulletphysics/bullet3/blob/fd161fa061539257036dd8f651bda3d3bf079556/src/BulletDynamics/Featherstone/btMultiBody.cpp#L1919
will write on a wrong segment of memory.

This change make sure to resize everything.

Found this error while implementing multibody inside Godot

### DIFF
--- a/src/BulletDynamics/Featherstone/btMultiBody.h
+++ b/src/BulletDynamics/Featherstone/btMultiBody.h
@@ -492,6 +492,7 @@ void addJointTorque(int i, btScalar Q);
 	void setNumLinks(int numLinks)//careful: when changing the number of m_links, make sure to re-initialize or update existing m_links
 	{
 		m_links.resize(numLinks);
+		m_matrixBuf.resize(n_links + 1);
 	}
 
 	btScalar getLinearDamping() const


AndreaCatania:
Probably manage this inside this function: /home/andrea/Workspace/git/godot/thirdparty/bullet/BulletDynamics/Featherstone/btMultiBody.cpp:340 is even better

erwincoumans:
You are never supposed to 'resize' the number of links, at construction time you pass in the number of links. How do you resize it?

AndreaCatania:
I'm using setNumLinks to resize it and then I call the finalize function
and it seems to work fine.

is useful to resize it so I can avoid to recreate it each time a link is
added

Let me know if is better to recreate it each time please, and Thanks for
the answer

erwincoumans:
If that works for you it is fine.
Can you resolve the errors?
>>> c:\projects\bullet3\src\bulletdynamics\featherstone\btMultiBody.h(505): error C2065: 'n_links': undeclared identifier [C:\projects\bullet3\cm\src\BulletDynamics\BulletDynamics.vcxproj]

AndreaCatania:
Sorry if I did a new PR to fix this but I'm not on my machine, right now, and I've managed this from GitHub Page.

So check this #2236 and thanks!

PR #2235, opened 2019-05-06
Fixes bug in getNumUserData in the physics direct client.

RanTig:
Fixes bug in getNumUserData in the physics direct client. This was inconsistent with the shared memory client.


erwincoumans:
Thanks, LGTM.

PR #2261, opened 2019-05-20
Fix -Wreorder and -Wunused-parameter in headers

EmilyMansfield:
This fixes warnings that occur in any user code including these files (in particular any file transitively including `btBulletDynamicsCommon.h`) when the `-Wreorder` and `-Wunused-parameter` flags are enabled in clang.


erwincoumans:
fixes were applied manually. THanks!

PR #2264, opened 2019-05-22
Fix code files encoding to make them all UTF-8

akien-mga:
I was checking encoding issues in Godot's codebase and found a couple coming from vendored Bullet files, so I went through the Bullet codebase to fix them.

I skipped everything in the `data/` folder to avoid modifying text files where the encoding might be required for proper parsing.
There's still `examples/OpenGLWindow/TwFonts.h` with a weird encoding due to it trying to embed the whole extended ASCII table in a comment (including control characters), but as it's thirdparty I skipped it (I also don't know how this table could be correctly embedded in a UTF-8 file).

I normalized Francisco León Nájera to Francisco Leon Najera as was already done in most other source files they authored, but if preferred I could also reencode it as Francisco León Nájera in UTF-8.

PR #2265, opened 2019-05-22
Convert DOS (CRLF) source files to Unix (LF) line endings

akien-mga:
Excluded `examples/pybullet/gym/pybullet_data/` which has many (3000+)
CRLF data files (obj, mtl, urdf), and `docs/pybullet_quickstart_guide`
which has generated .js and .htm files with CRLF line endings too.

----

Before this change:
```
$ grep --exclude-dir={.git,data,docs,examples} -IUlr $'\r' | wc -l
18
$ grep --exclude-dir={.git,data,docs,examples} -IUlr $'\n' | wc -l
1085
```

After:
```
$ grep --exclude-dir={.git,data,docs,examples} -IUlr $'\r' | wc -l
0
$ grep --exclude-dir={.git,data,docs,examples} -IUlr $'\n' | wc -l
1085
```


akien-mga:
Note: will likely conflict with #2264 which fixes encoding in a file with CRLF line endings, but I can rebase either as needed.

erwincoumans:
We use clang-format for formatting. It is much faster if I run that tool myself, I don't have time to sift through tons of files, making sure no other change slipped though.

I'll run clang-format again soon, to make sure formatting is ok. At the moment it is only used on .h and .cpp files, no .mk or txt.

akien-mga:
AFAIK clang-format doesn't impact line ending characters.

akien-mga:
You can use `?w=1` to see changes without whitespace changes: https://github.com/bulletphysics/bullet3/pull/2265/files

Or run `dos2unix` on those files yourself if you prefer.

erwincoumans:
OK, ?w=1 resolves the issue, I'll accept it. Thanks for the help!

PR #2270, opened 2019-05-27
Destroy semaphore after join

jamoflaw:
I get a consistent hang on the pthread worker threads when using btCreateDefaultTaskScheduler() (both my worker threads get stuck on sem_wait() - Line 181) when the semaphore is destroyed before the thread is joined, and the application never exits.

This change resolves this issue in my testing, thought it's difficult to initially see why, however given that both threads end up waiting on a semaphore which has been destroyed the change enforces that that situation cannot happen.

### DIFF
--- a/src/LinearMath/TaskScheduler/btThreadSupportPosix.cpp
+++ b/src/LinearMath/TaskScheduler/btThreadSupportPosix.cpp
@@ -304,8 +304,8 @@ void btThreadSupportPosix::stopThreads()
 		checkPThreadFunction(sem_post(threadStatus.startSemaphore));
 		checkPThreadFunction(sem_wait(m_mainSemaphore));
 
-		destroySem(threadStatus.startSemaphore);
 		checkPThreadFunction(pthread_join(threadStatus.thread, 0));
+		destroySem(threadStatus.startSemaphore);
 	}
 	destroySem(m_mainSemaphore);
 	m_activeThreadStatus.clear();


erwincoumans:
Thanks for the report. How can we reproduce this hang exactly?

jamoflaw:
https://github.com/jamoflaw/bullet3-pr2270

So it's not the easiest to reproduce as I think it is a race, which likely means that this fix will only mask a deeper issue - though I still think that destroying the resource after the thread is joined is a more sensible approach longer term.

The way I reproduced this more reliably is to run it with the commandline:


`
valgrind ./bullet_mt_test
`

Running it with valgrind will show the stacktrace of the access to the _deleted_ semaphore (showing in run 2 in the repo). I'm not familiar with the flow of the threads right now so haven't really had a look into where the races may be occurring, but the stacktrace of run1 shows another instance where it triggered an assertion in the bullet code which may be a clue.

Also I hope I'm actually using it right! This is my first outing with the MT version so I apologise now if I'm just using it completely wrong!!!!!

jamoflaw:
If you run it alone enough you can trigger the assertion pretty easily too:

`
bullet_mt_test: /home/james/evodeps/linux_x64/bullet3/src/LinearMath/btAlignedObjectArray.h:177: T& btAlignedObjectArray<T>::operator[](int) [with T = btThreadSupportPosix::btThreadStatus]: Assertion `n >= 0' failed.
`

erwincoumans:
Seems reasonable, thanks for the contribution!

PR #2278, opened 2019-06-03
Fix GCC alignment macros on Windows

biller23:


devshgraphicsprogramming:
Maybe use `alignas` if C++11 is available?

+1 for this PR, the alignment issue makes our Bullet integration with our 3D engine a bit unsafe (unaligned loads) for x86 GCC builds (Linux).

@khom-khun anything to add?

erwincoumans:
I haven't used MinGW/cygwin to test, but it looks OK. Let's give it a try and revert it if causes issues.

erwincoumans:
Thanks!

PR #2279, opened 2019-06-03
Fix initialization order

AndreaCatania:
This commit is necessary in order to fix the error that we get in godot with the flag: "-Werror=reorder"


erwincoumans:
LGTM, thanks for the help!

PR #2293, opened 2019-06-18
get mesh data api

fuchuyuan:


erwincoumans:
Here is a follow up PR that makes it work for soft body:
https://github.com/bulletphysics/bullet3/pull/2294

Issue #2139, opened 2019-03-04
[Question] Python API for Light source position adjustment


junhyeokahn:
Hello,
Is there a python API for changing light source in GUI?
I saw some related work for camera images but not for GUI light source.

Pandhariix:
Hello, I don't think so, but that could indeed be a very nice addition

maedmaex:
@junhyeokahn could you share the related work you mentioned :) thanks

Pandhariix:
By the way,
> Is there a python API for changing light source in GUI?

This has been resolved in PR #2298

junhyeokahn:
@maedmaex Hi, please see [this](https://github.com/junhyeokahn/PnC/blob/7ef3cc3163d4aa1f32d8b5271223c3dba5f5d1da/ReinforcementLearning/MyGym/WalkerBaseEnv.py#L120).
Note that this is modified based on [pybullet-gym](https://github.com/benelot/pybullet-gym).

PR #2295, opened 2019-06-18
Python API for GUI light source position adjustment, solving #2139

Pandhariix:
This PR proposes a __setLightPosition__ method, also available through pybullet, allowing the user to modify the position of the __light source__ in __GUI mode__ (it won't work in __DIRECT__ mode). It solves issue #2139.

The method can be used as pictured in the following snippet:
```python
pybullet.setLightPosition(15.0, 0.0, 15.0)
```

The method can be called several times at runtime:
light

### DIFF
--- a/examples/BlockSolver/RigidBodyBoxes.cpp
+++ b/examples/BlockSolver/RigidBodyBoxes.cpp
@@ -8,7 +8,7 @@ class RigidBodyBoxes : public CommonRigidBodyBase
 {
 	int m_option;
 	int m_numIterations;
-	int m_numBoxes = 4;
+	int m_numBoxes;
 	btAlignedObjectArray<btRigidBody*> boxes;
 	static btScalar numSolverIterations;
 
@@ -38,7 +38,8 @@ btScalar RigidBodyBoxes::numSolverIterations = 50;
 RigidBodyBoxes::RigidBodyBoxes(GUIHelperInterface* helper, int option)
 	: CommonRigidBodyBase(helper),
 	  m_option(option),
-	  m_numIterations(numSolverIterations)
+	  m_numIterations(numSolverIterations),
+	  m_numBoxes(4)
 {
 	m_guiHelper->setUpAxis(2);
 }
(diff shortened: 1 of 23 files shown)


erwincoumans:
Thanks for the contribution, that robot looks great!
I was considering re-using the pybullet.configureDebugVisualizer to set light position, configure shadow resolution etc. But since you already put in all the work, I'll consider merging it as-is after having a closer look.

Pandhariix:
Thanks, it's our virtual Pepper model :) 
Wonderful! I'm not really familiar with the bullet code base, so I might have missed some things here and there, but globally I tried to match the way the other pybullet functionalities were implemented.

erwincoumans:
There are some conflicts easy to resolve, could you do it in your PR? Thanks!

Pandhariix:
Sure, the last commits should resolve them!

erwincoumans:
Thanks for the update. I had a closer look and decided that it is better to add this feature in the 'configureDebugVisualizer', to keep it separate from the 'getCameraImage' which also has a lightPosition argument.
See https://github.com/bulletphysics/bullet3/pull/2298
This also exposes the shadowMapWorldSize and shadowMapResolution (texture resolution)
Thanks for the PR anyway!

Pandhariix:
Very nice, I will use that feature a lot!

>  I had a closer look and decided that it is better to add this feature in the 'configureDebugVisualizer', to keep it separate from the 'getCameraImage' which also has a lightPosition argument.

It might be clearer that way, you're right. I also wanted to implement a way to control the intensity / color of the light source (as an extra parameter in _configureDebugVisualizer_ perhaps), do you think that it could be a relevant addition to the project ?

erwincoumans:
Yes, if you want to create a PR to set light color/intensity to configureDebugVisualizer that is welcome.

Pandhariix:
Alright, I'll surely submit a PR then.

Kartoffelchen:
> Thanks for the update. I had a closer look and decided that it is better to add this feature in the 'configureDebugVisualizer', to keep it separate from the 'getCameraImage' which also has a lightPosition argument.
> See #2298
> This also exposes the shadowMapWorldSize and shadowMapResolution (texture resolution)
> There is an example in pybullet/examples/configureDebugVisualizer.py, like this
> 
> animate_light
> 
> Thanks for the PR anyway!

Is there any method to change the intensity of the light source? Thanks in advance!

PR #2296, opened 2019-06-19
fix the issue of converting the double to single.

Steven89Liu:


erwincoumans:
Thanks for tracking this down!

Issue #2246, opened 2019-05-08
Numerical difference between pybullet and RBDL/KDL


lillmaria:
Hi! 

I am using pybullet, RBDL and KDL to evaluate the inverse dynamics and inertia matrix of a kuka-lwr and a ur5. When comparing 1000 evaluations of the inverse dynamics/inertia matrix, the numerical difference between RBDL and KDL is 10^(-12), while the difference between KDL/RBDL and pybullet is 10^(-3). 

I use a standard kuka-lwr urdf and ur5 urdf to load the robot model for all the libraries. I have also enabled the URDF_USE_INERTIA_FROM_FILE flag. 

Also, I have noticed that if all the joint origins and link origins are changed to (0, 0, 0), or close to this, the numerical difference decreases to 10^(-12).  

Du you know why the numerical difference is so "big" for the original urdfs?

erwincoumans:
Without access to your test it is impossible to tell where the differences are. Is your test available somewhere?

lillmaria:
After some testing, it seems to me that the problem is related to integers vs. decimals. As long as the urdf's origins consist of integers the numerical difference is of factor 10^(-12), while when changing just one element from integer to a decimal, the numerical difference is of factor 10^(-3), sometimes 10^(-2). 

To demonstrate a minimal example, I use a pantilt-urdf and a test script to compare gravitational impact on the joints using the inverse dynamics functionality: 

```

import rbdl
import numpy as np
import PyKDL as kdl
import kdl_parser_py.urdf as kdlurdf
import pybullet as pb

root = 'base_link'
tip = 'tilt_link'
path_to_urdf = 'pantilt.urdf'
ok, tree = kdlurdf.treeFromFile(path_to_urdf)

kdlmodel = tree.getChain(root,tip)
rbdlmodel = rbdl.loadModel(path_to_urdf)

sim = pb.connect(pb.DIRECT)
pbmodel = pb.loadURDF(path_to_urdf, useFixedBase=True, flags = pb.URDF_USE_INERTIA_FROM_FILE)
pb.setGravity(0, 0, -9.81)
q_max = [3.14, 4.64]
q_min = [-3.13, -4.64]
n_joints = 2

q_kdl = kdl.JntArray(n_joints)
tau_kdl = kdl.JntArray(n_joints)
gravity_kdl = kdl.Vector()
gravity_kdl[2] = -9.81

q_pb = [None]*n_joints
zeros_pb = [None]*n_joints

[...]

PR #2301, opened 2019-06-20
port the rbdl into the bullet as well as write a simple test to compa…

Steven89Liu:
…re the ABA between the bullet and the rbdl, including:

1. download the rbdl from the https://bitbucket.org/rbdl/rbdl/downloads/?tab=downloads, (rbdl-rbdl-0879ee8c548a.zip)
2. change three CMakeList.txt and remain the original file there as CMakeList.txt-orig,
3. add a m_flag in the class MyBtMultiBodyFromURDF to indicate that if using the inertia value from the urdf file, or other usage in the future,
4. add a test test/InverseDynamics/test_dynABA_bulletVSrbdl.cpp to compare the ABA between the rbdl and the bullet.

### DIFF
--- a/Extras/InverseDynamics/btMultiBodyFromURDF.hpp
+++ b/Extras/InverseDynamics/btMultiBodyFromURDF.hpp
@@ -30,6 +30,7 @@ public:
 		m_collisionConfiguration = 0x0;
 		m_dynamicsWorld = 0x0;
 		m_multibody = 0x0;
+                m_flag = 0x0;
 	}
 	/// dtor
 	~MyBtMultiBodyFromURDF()
@@ -43,6 +44,7 @@ public:
 	}
 	/// @param name path to urdf file
 	void setFileName(const std::string name) { m_filename = name; }
+        void setFlag(int flag) { m_flag = flag; }
 	/// load urdf file and build btMultiBody model
 	void init()
 	{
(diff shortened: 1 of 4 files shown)


erwincoumans:
I love this comparison and am going to check it out. I find adding the whole RBDL library as dependency in Bullet too large. Wonder if the forward dynamics can be compared similar to the inverse dynamics, purely in Python, like the RBDL/KDL/Pybullet comparison of inverse dynamics?
https://github.com/bulletphysics/bullet3/issues/2246

Steven89Liu:
i think it is more better to add the whole RBDL libray for the first version, we can delete some unused part in next version.  
It should can be implemented in Python, i will look into this later.

erwincoumans:
It is not a good idea to add full other dynamics libraries into Bullet, it clutters the git history and confuses the users into thinking Bullet depends on those external libraries (which is not the case).

Steven89Liu:
OK, i will delete these unused part.

Steven89Liu:
”The job exceeded the maximum time limit for jobs, and has been terminated.“
don't know what this error mean, i will check it again tomorrow.

Steven89Liu:
Could you help to merge this PR?

erwincoumans:
Sorry I wasn't clear, but I don't want to add any RBDL files into Bullet.  One option is to add a continuous integration test (travis, appveyor) that pulls the rbdl files (specific revision) into the right location. So we only need to add an option to enable this RBDL comparison test (it is disabled by default!) and the test itself.

erwincoumans:
I'll try to find time soon, to only add the test file itself, without RBDL, and merge it into Bullet.

erwincoumans:
Thanks a lot, this can be merged.

PR #2302, opened 2019-06-20
Add ability to throw objects towards deepmimic humanoid to check stability

BartMoyaers:
This pull request adds functionality to the deepmimic humanoid example file "testrl.py".
By pressing the 'x' button, balls are thrown towards the humanoid model, like in the original deepmimic repo.

This way you can interactively test how stable the policy is.

For future reference: run the testrl.py file like this:
`python testrl.py --arg_file run_humanoid3d_backflip_args.txt`

P.S. I noticed that the `removeBody` pybullet method can be quite slow removing all the thrown balls after the simulation finishes. (I might create an issue for this.)

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/deep_mimic/env/humanoid_stable_pd.py
+++ b/examples/pybullet/gym/pybullet_envs/deep_mimic/env/humanoid_stable_pd.py
@@ -144,6 +144,7 @@ class HumanoidStablePD(object):
     self.setSimTime(0)
 
     self.resetPose()
+    self.thrown_body_ids = []
 
   def resetPose(self):
     #print("resetPose with self._frame=", self._frame, " and self._frameFraction=",self._frameFraction)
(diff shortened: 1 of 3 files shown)


erwincoumans:
Thanks, I'll check it out.
>> removing all the thrown balls after the simulation finishes

Can you try disabling rendering briefly before removing all objects one-by-one, then re-enable?

`pybullet.configureDebugVisualizer(pybullet.COV_ENABLE_RENDERING=0)`
remove objects
`pybullet.configureDebugVisualizer(pybullet.COV_ENABLE_RENDERING=1)`

BartMoyaers:
> Can you try disabling rendering briefly before removing all objects one-by-one, then re-enable?

This seems to work, thanks for the tip!

Just for reference, the correct way to disable the rendering is:
`pybullet.configureDebugVisualizer(pybullet.COV_ENABLE_RENDERING,0)`

(This is correctly mentioned in the pybullet quickstart guide.)

PR #2305, opened 2019-06-24
parse fall contact bodies

BartMoyaers:
This pull request adds parsing of the `--fall_contact_bodies` argument for the DeepMimic example.

This is necessary for training and running different policies like "CartWheel", where the hands need to be able to touch the ground.

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/deep_mimic/env/humanoid_stable_pd.py
+++ b/examples/pybullet/gym/pybullet_envs/deep_mimic/env/humanoid_stable_pd.py
@@ -19,9 +19,11 @@ jointFrictionForce = 0
 
 class HumanoidStablePD(object):
 
-  def __init__(self, pybullet_client, mocap_data, timeStep, useFixedBase=True):
+  def __init__( self, pybullet_client, mocap_data, timeStep, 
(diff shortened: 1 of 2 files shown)


examples/pybullet/gym/pybullet_envs/deep_mimic/env/humanoid_stable_pd.py
```diff
@@ -135,7 +137,10 @@ def __init__(self, pybullet_client, mocap_data, timeStep, useFixedBase=True):
     self._jointDofCounts = [4, 4, 4, 1, 4, 4, 1, 4, 1, 4, 4, 1]
 
     #only those body parts/links are allowed to touch the ground, otherwise the episode terminates
-    self._allowed_body_parts = [5, 11]
+    fall_contact_bodies = []
```
erwincoumans: I'd like to be backward compatible, so that if you don't pass in the argument it will use the [5,11] default. With the new change, things break unless you pass this flag.

erwincoumans:
Thanks for the contribution, let's merge it anyway to learn the other motions.

Also note that the learning speed increased a lot, since recent optimizations.

PR #2309, opened 2019-06-26
Update PhysicsServerCommandProcessor and plugins to support render

fuchuyuan:


examples/SharedMemory/PhysicsServerCommandProcessor.cpp
```diff
@@ -12021,7 +12060,18 @@ b3Notification createTransformChangedNotification(int bodyUniqueId, int linkInde
 	return notification;
 }
 
-void PhysicsServerCommandProcessor::addTransformChangedNotifications()
+#ifndef SKIP_SOFT_BODY_MULTI_BODY_DYNAMICS_WORLD
+b3Notification createSoftBodyChangedNotification(int bodyUniqueId, int linkIndex, const btSoftBody* psb)
```
erwincoumans: btSoftBody* psb pointer is not used, can it be removed?

erwincoumans:
I'll merge it for now, we can remove the psb in a separate PR.
Thanks for the contribution!

PR #2315, opened 2019-07-07
Call drawRect on Mac to fix OpenGL issue

nbelakovski:
This fixes an issue on Mac where an OpenGL window doesn't show anything until it is resized.

I'm not sure if this is an issue for others, or if this fix has side effects, but this was an issue I ran into and I figured I would at least share my solution in case someone else stumbles onto it.


erwincoumans:
Thanks. Since our Mac's don't have this issue, I can't reproduce it, but it doesn't hurt so let's add it.

PR #2316, opened 2019-07-09
Fix MSAN error 

macandy13:
Fix memory sanitizer error on potential reading of uninitialized memory. This can happen for the RGB values that are only conditionally set in TinyRendererVisualShapeConverter::addVisualShape.


erwincoumans:
Thanks, LGTM.

PR #2324, opened 2019-07-11
Update readme and gitignore

sanchitgarg:
Updated readme to reflect correct sln file name
Added build3/vs2010 to .gitignore


erwincoumans:
Good point, it needed updating. Thanks!

PR #2326, opened 2019-07-13
Fix GCC alignment macros for btScalar.h

biller23:
As of the merged #2278, but for btScalar.h.
I forgot to include this file in the original pull request.


erwincoumans:
Seems to be safe to apply (I don't use gnu on Windows so I didn't test), thanks for the contribution!

PR #2354, opened 2019-08-05
add support to load rigidbody

fuchuyuan:
This PR updates  PhysicsServerCommandProcessor.cpp and btRigidBody for us to load rigidbody in sim.


erwincoumans:
We can't just add 'mass' to rigid body, can you remove it? There is already inv_mass.

fuchuyuan:
I have changed it so that btRigidbody doesn't store mass

erwincoumans:
Thanks for the fix, LGTM!

PR #2357, opened 2019-08-06
API and pybullet bindings for setting warmstarting factor

fuchuyuan:
This CL adds functions in PhysicsClientC_API and pybullet.c for setting the  physics engine parameter warm starting factor.


erwincoumans:
Thanks for the contribution, LTGM!

PR #2361, opened 2019-08-08
 Clears body and user data caches on syncBodyInfo and syncUserData

RanTig:
When working with two physics clients on a single physics server, the following problem can occur:

Client 1 adds bodies
Client 2 calls syncBodyInfo() - receives state of all bodies correctly.
Client 1 calls resetSimulation()
Client 2 calls syncBodyInfo() - still has all bodies in the cache (e.g. getNumBodies() does not return 0).

Similar problems happen with user data when resetSimulation() is called in multi-client setups.


erwincoumans:
Thanks for the report and fix, LGMT.

PR #2365, opened 2019-08-13
update loading softbody

fuchuyuan:
This PR updates PhysicsServerCommandProcessor::processLoadSoftBodyCommand.

1. Change default scaling factor and mass to 1.
2. Apply transformation in the order of: scale, rotate, transform, which better matches the user's expectation.


erwincoumans:
LGTM, thanks!

PR #2378, opened 2019-08-22
Add missing semicolons for debug build

sponji:


erwincoumans:
manually added them. please let me know if I missed some.

PR #2387, opened 2019-08-29
Add base name

fuchuyuan:
In case urdf is loaded as rigidbody, the baselink name is used both for body name and link name.


examples/SharedMemory/PhysicsClientSharedMemory.cpp
```diff
@@ -621,6 +621,7 @@ const SharedMemoryStatus* PhysicsClientSharedMemory::processServerStatus()
 					BodyJointInfoCache* bodyJoints = new BodyJointInfoCache;
 					m_data->m_bodyJointMap.insert(bodyUniqueId, bodyJoints);
 					bodyJoints->m_bodyName = serverCmd.m_dataStreamArguments.m_bodyName;
+					bodyJoints->m_baseName = serverCmd.m_dataStreamArguments.m_bodyName;
```
madebr: This is a copy of the line just above it.

PR #2393, opened 2019-09-03
Fix typo: s/duraction/duration/g

muupan:


erwincoumans:
Thanks for the typo fix.

PR #2396, opened 2019-09-03
Add body type to dynamics info

fuchuyuan:
Add a field in b3DynamicsInfo for users to query for body type (btRigidBody, btMultiBody, btSoftBody).


erwincoumans:
Thanks, LGTM.

PR #2397, opened 2019-09-04
Fix memory leak in RestoreState

mbreyer:
The `btMultiBodyWorldImporter` within `processRestoreStateCommand` is not deallocated, leading to memory issues if this method is called many times. This pull request simply adds a `delete` statement at the end of the method.


erwincoumans:
Good find, thanks for the report/PR!

PR #2417, opened 2019-09-22
Fix warnings with -Wcomment, -Wundef and -Wdeprecated-copy

dbartolini:


dbartolini:
It would be nice to get #2261 merged as well.

dbartolini:
Splitting in three distinct PRs.

PR #2449, opened 2019-10-24
Fix deformable performance

xhan0619:
Changed sdf grid size for better performance.
Fixed an uninitialized variable m_useSelfCollision.


erwincoumans:
LGTM, thanks!

PR #2454, opened 2019-10-29
Fix call to convert_to_leg_model

krishpop:
Fix call to convert_to_leg_model which is defined as a static method without an underscore in MinitaurExtendedEnv


erwincoumans:
Well spotted, thanks for the fix. Since over a year we have been using and developing envs_v2 for various research quadruped locomotion projects. I'm planning to open source this later this year.

PR #2458, opened 2019-10-30
cull out most of the SVD calculation when stress is low

xhan0619:
We want to perform SVD to clamp the max stress if max(abs(sigma)) > max_p. So we perform a rough estimate on  tr(P^T P) = sigma_0^2 + sigma_1^2 + sigma_2.
If tr(P^T P) < max_p * max_p, then max(abs(sigma)) < max_p, and we don't need to calculate the SVD.


erwincoumans:
Thanks!

PR #2460, opened 2019-10-31
Update bullet.pc.cmake

sjdrc:
Change pkgconfig file to match example in pkg-config man files.

This PR updates the pkgconfig file to a more standard format instead of directly using hardcoded paths. This now allows for more flexibile usage, such as installations that can be moved between locations and systems. Using `pkg-config --define-prefix` allows pkg-config to override the prefix variable based on where it thinks the library is installed to.

With the lack of proper CMake targets support, this is the easiest way to use bullet with a project using modern CMake.

### DIFF
--- a/bullet.pc.cmake
+++ b/bullet.pc.cmake
@@ -1,6 +1,11 @@
+prefix=@CMAKE_INSTALL_PREFIX@
+exec_prefix=${prefix}
+libdir=${exec_prefix}/@LIB_DESTINATION@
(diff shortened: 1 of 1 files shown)


sjdrc:
CI Job failed due to maximum time limit.

erwincoumans:
@scpeters Could you please give some feedback on those changes?

scpeters:
It looks good to me.

erwincoumans:
Thanks, then let's merge it.

PR #2466, opened 2019-11-05
Correct xScale in b3ComputeProjectionMatrixFOV

wkentaro:
fov is inside the tan(.) so we can't simply apply . / aspect.

### DIFF
--- a/examples/SharedMemory/PhysicsClientC_API.cpp
+++ b/examples/SharedMemory/PhysicsClientC_API.cpp
@@ -4270,8 +4270,10 @@ B3_SHARED_API void b3ComputeProjectionMatrix(float left, float right, float bott
 
 B3_SHARED_API void b3ComputeProjectionMatrixFOV(float fov, float aspect, float nearVal, float farVal, float projectionMatrix[16])
 {
-	float yScale = 1.0 / tan((B3_PI / 180.0) * fov / 2);
-	float xScale = yScale / aspect;
+	float yFOV = (B3_PI / 180.0) * fov;
+	float xFOV = yFOV * aspect;  // aspect = width / height
+	float yScale = 1.0 / tan(yFOV / 2);
+	float xScale = 1.0 / tan(xFOV / 2);
 
 	projectionMatrix[0 * 4 + 0] = xScale;
 	projectionMatrix[0 * 4 + 1] = float(0);


erwincoumans:
retrigger CI

erwincoumans:
We are using the OpenGL formulation described here:
https://gamedev.stackexchange.com/questions/120338/what-does-a-perspective-projection-matrix-look-like-in-opengl

OpenCV uses something similar to PyBullet:
https://github.com/opencv/opencv/blob/422b967e9ae802cc0b4aefcabe78a9568a71754b/modules/viz/src/types.cpp#L216

We don't change this existing API.
If you can provide convincing documentation/paper/explanation we can consider adding a second API b3ComputeProjectionMatrixFOV2.

wkentaro:
I will add some visualizations once I get time to do that.
But basically, this issue is caused by missing basic math principle: `tan` is not linear function, so `tan(xy) != tan(x) * y` (every high school student knows 🤣)

Without my modification, `yScale` is correct, but `xScale` is wrong,
because `aspect` should be applied inside of the `tan(.)` so we can't do `xScale = yScale / aspect`.
This wrong rendering makes the point cloud always wrong in x-axis using the intrinsic matrix that is supposed to be correct (because the actual intrinsic matrix (specifically fx) is different from the expected one).

> We don't change this existing API.

That makes sense. Probably many people are already using the wrong pairs of rendered rgb, mask, depth and intrinsic matrix on their dataset anyways, so changing the API can mess up everything (hopefully they are using different API e.g. b3ComputeProjectionMatrix, which may be correct).

erwincoumans:
You repeat that aspect has to be inside the 'tan' function but don't give any explanation or link/doc why it has to be inside. Please check the derivation in previous links and also OpenCV, they clearly explain why aspect is outside of tan.

wkentaro:
Ah, I see. I will.

wkentaro:
I checked those links, and I found the mistake on my side.
I was thinking `fovx = aspect * fovy` but it's not correct. The correct equation is `tan(fovx) = aspect * tan(fovy)`, which is something obvious.

Really sorry disturbing you with this totally wrong PR, and thank you for correcting me!

erwincoumans:
fovx = fovy * aspect is wrong indeed.
the correct version is: fieldOfViewX = 2 * atan(tan(fieldOfViewY * 0.5) * aspect)
See details here: https://stackoverflow.com/questions/5504635/computing-fovx-opengl

wkentaro:
Yeah, Thanks for the useful links!

PR #2467, opened 2019-11-06
Split impulse

xhan0619:
Support position correction between deformable and rigid bodies using split impulse.

### DIFF
--- a/src/BulletSoftBody/btDeformableContactConstraint.h
+++ b/src/BulletSoftBody/btDeformableContactConstraint.h
@@ -50,6 +50,9 @@ public:
     // the constraint is solved by calculating the impulse between object A and B in the contact and apply the impulse to both objects involved in the contact
     virtual btScalar solveConstraint() = 0;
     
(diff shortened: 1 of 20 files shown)


src/BulletSoftBody/btDeformableContactConstraint.h
```diff
@@ -154,6 +181,13 @@ class btDeformableRigidContactConstraint : public btDeformableContactConstraint
     virtual btVector3 getVa() const;
     
     virtual btScalar solveConstraint();
+    
+    virtual btScalar solveSplitImpulse(const btContactSolverInfo& infoGlobal);
+    
+    virtual void setPenetrationScale(btScalar scale)
```
erwincoumans: Is the 'scale' the amount of penetration to be resolved? In other parts of the code, we call this the 'error reduction parameter'.

xhan0619: Yes, this is the same as erp. I will modify.

Issue #2473, opened 2019-11-07
BulletExampleBrowser: collisions don't work in soft body's collide/collide 2/collide 3 samples


consultit:
Hi, I've pulled the [latest commit](https://github.com/bulletphysics/bullet3/commit/e66982d6586a660439bbb8a91c1c5a86bb338dd8), and built on debian 10 (using premake), but the soft bodies in collide samples don't collide anymore.

PR #2475, opened 2019-11-07
Fix issue #2473

xhan0619:


erwincoumans:
Thanks for the fixes. There are two PR that seem to contain the erp changes, do we need both?
(see also https://github.com/bulletphysics/bullet3/pull/2474)

xhan0619:
No, they are the same commit and only one is need it. I'll close the other PR.

PR #2478, opened 2019-11-09
fix the bug that prevents the pd control forces/torques being added

xhan0619:
Fixes the following bug:
In deformable simulation, we need to move objects to temporary position x_{n+1}* = x_n + dt * v_{n+1}*, where v_{n+1}* for rigid/multibody is the velocity after gravity was applied. So we applied the gravity and cleared the force. But we accidentally clear the PD control forces/torques along with it. This commit fixes that issue by only clearing the gravity force. 

It also removes an unnecessary macro guard around constraint solver switch.


erwincoumans:
Thanks for the fixes! LGTM.

PR #2484, opened 2019-11-13
Add macro guard for compiling

xhan0619:


erwincoumans:
Thanks! It would be good to consider the conditional creation of a btDeformable*World instead of using the #ifdef SKIP_DEFORMABLE_WORLD. Let's discuss it soon.
Also, the ExampleBrowser demos don't show boundary faces for the tetrahedral meshes, can we add those too?

xhan0619:
Yeah I'll add the faces for the demos in ExampleBrowser. We can discuss conditional creation tomorrow.

erwincoumans:
Since the world creation is now determined at run-time, this PR can be closed?

xhan0619:
> Since the world creation is now determined at run-time, this PR can be closed?

Yes, we should close it.

PR #2494, opened 2019-11-18
add command to set Sparse SDF grid voxel size

xhan0619:
Add option to configure sparse sdf voxel size in pybullet. 
Tune parameters to get the rolling behavior of the deformable ball.


examples/SharedMemory/SharedMemoryPublic.h
```diff
@@ -53,6 +53,7 @@ enum EnumSharedMemoryClientCommand
 	CMD_REQUEST_BODY_INFO,
 	CMD_REQUEST_INTERNAL_DATA,
 	CMD_STEP_FORWARD_SIMULATION,
+    CMD_SET_SPARSE_SDF,
```
erwincoumans: best to add those at the end, to avoid version incompatibilities. If not at the end, of changing existing structures, the version at the top of SharedMemoryPublic.h  needs to be updated.

erwincoumans:
Thanks for the PR. I re-implemented it here, together with various other fixes:
https://github.com/bulletphysics/bullet3/pull/2495

PR #2500, opened 2019-11-21
get correct AABB for softbody instead of using default AABB

fuchuyuan:


erwincoumans:

PR #2505, opened 2019-11-25
Add common CMake and Visual Studio project files to gitignore

stephenjust:


erwincoumans:
Thanks for the contribution, that seems reasonable.

PR #2506, opened 2019-11-26
Deformable vs. Rigid face contact improvement and Deformable vs. Deformable contact improvement

xhan0619:


erwincoumans:
Thanks, looks good to me.

PR #2526, opened 2019-12-06
Make standalone example Apps run in realtime

AndrewMeadows:
I was trying to test an idea by hacking the **BasicDemo** example and discovered the simulation in the standalone App binary would run too fast.  On closer examination I discovered: small timesteps were being clamped to be at least 0.1 second which I assume was a typo: they should be clamped to not be larger than that.

This pull request fixes the typo in the first commit, however I then introduced an explicit multi-step simulation strategy for the standalone examples.  If the extra work is not useful feel free to discard this PR and port just the first commit where the inequality operator is flipped.

### DIFF
--- a/examples/StandaloneMain/main_opengl_single_example.cpp
+++ b/examples/StandaloneMain/main_opengl_single_example.cpp
@@ -94,7 +94,7 @@ int main(int argc, char* argv[])
 		app->m_instancingRenderer->updateCamera(app->getUpAxis());
 
 		btScalar dtSec = btScalar(clock.getTimeInSeconds());
-		if (dtSec < 0.1)
+		if (dtSec > 0.1)
 			dtSec = 0.1;
 
 		example->stepSimulation(dtSec);


examples/CommonInterfaces/CommonRigidBodyBase.h
```diff
@@ -75,7 +75,12 @@ struct CommonRigidBodyBase : public CommonExampleInterface
 	{
 		if (m_dynamicsWorld)
 		{
-			m_dynamicsWorld->stepSimulation(deltaTime);
+			const btScalar lowestExpectedFrameRate = 10.0;
+			const btScalar subStepFrequency = 60.0;
+			const int maxSubSteps = int(subStepFrequency / lowestExpectedFrameRate);
+			const btScalar subStepDuration = 1.0 / subStepFrequency;
+
+			m_dynamicsWorld->stepSimulation(deltaTime, maxSubSteps, subStepDuration);
```
AndrewMeadows: This is how I reason through what should be the `subStepDuration` and `maxSubSteps`, so I'm adding it here to help clarify for future devs.

examples/StandaloneMain/main_opengl_single_example.cpp
```diff
@@ -93,9 +95,8 @@ int main(int argc, char* argv[])
 		app->m_instancingRenderer->init();
 		app->m_instancingRenderer->updateCamera(app->getUpAxis());
 
-		btScalar dtSec = btScalar(clock.getTimeInSeconds());
-		if (dtSec < 0.1)
-			dtSec = 0.1;
```
AndrewMeadows: The real bug was at old line 97: the `<` comparison operator should go the other direction.  Should a github issue be added for bug tracking?  Or is that unnecessary?

erwincoumans: Thanks for the contribution. Can you just swap the > sign and undo the other changes? I'd rather keep the change minimal, and not start clamping by default in the CommonRigidBodyBase.h

erwincoumans:
Thanks for the contribution. Can you just swap the > sign and undo the other changes? I'd rather keep the change minimal, and not start clamping by default in the CommonRigidBodyBase.h

AndrewMeadows:
Sure no problem.  I reverted the second and third commits, done.

Ah, github tells me you ported it manually.  That is good.  I'll close this PR.

PR #2528, opened 2019-12-08
Actually use self collisions when specified

jackson-waschura:
This fork fixes a small bug where the useSelfCollisions argument to pybullet's loadSoftBody method was never used or passed onto the client. I found that there existed a b3 API function for this purpose that was never called. I suspect this was simply overlooked previously.


erwincoumans:
Thanks for hooking up the API.

PR #2530, opened 2019-12-09
Unify usage of BulletClient in examples

davidrusu:
We have 3 variations of the `BulletClient` class, all at seemingly different levels of maturity. This PR modifies some the gym examples to all use the `pybullet_utils.bullet_client.BulletClient` version and removes the others.


erwincoumans:
Good point, the bullet_client in pybullet_utils is the best version, good to get rid of the others.

PR #2540, opened 2019-12-14
Remove unnecessary copy constructor in btBroadphasePair

dbartolini:
The implicitly-defined copy constructor would be equivalent.

Moreover, the generation of the implicitly-defined copy assignment operator is deprecated since C++11 if the type has a user-defined copy constructor. This generates an annoying warning on recent compilers.


erwincoumans:
thanks for the small fix (I don't have bandwidth to sift through enormous patches), looks good to me.

PR #2541, opened 2019-12-14
Fix multi-line comment warning with -Wcomment

dbartolini:
Fixes an annoying warning with -Wcomment under GCC.
This is probably some leftover from clang-format.


erwincoumans:
Thanks for the improvement, LGTM.

PR #2543, opened 2019-12-15
Fix warning with -Wunused-parameter

dbartolini:
This fixes an annoying warning with -Wunused-parameter in user-included header.


erwincoumans:
Thanks, looks good to me.

PR #2544, opened 2019-12-15
Remove unnecessary copy constructor in btDbvt

dbartolini:
The implicitly-defined copy constructor would be equivalent.

Moreover, the generation of the implicitly-defined copy assignment operator is deprecated since C++11 if the type has a user-defined copy constructor. This generates an annoying warning on recent compilers.


erwincoumans:
Thanks! LGTM

PR #2546, opened 2019-12-17
add link index to setdamping C APIs. If link index is not set, the co…

fuchuyuan:
…mmand is igored for rigidbody


erwincoumans:
Thanks, LGTM.

PR #2550, opened 2019-12-19
Group deformable constraint solves by islands

xhan0619:
Modify constraint solves involving deformable objects so that they are only solved along with the islands they are in to avoid repetition.


examples/SharedMemory/b3RobotSimulatorClientAPI_NoDirect.h
```diff
@@ -89,6 +89,55 @@ struct b3RobotSimulatorLoadSoftBodyArgs
 	}
 };
 
+
+struct b3RobotSimulatorLoadDeformableBodyArgs
```
erwincoumans: A lot of of parameters are not initialized in the constructor. Can you initialize them either with some reasonable value or -1 (or some flag) to mark that they should not be used?

erwincoumans: See for example this: https://github.com/bulletphysics/bullet3/blob/master/examples/SharedMemory/b3RobotSimulatorClientAPI_NoDirect.h#L100

xhan0619: Done. I also tuned the parameter of the cloth a little bit.

erwincoumans:
Thanks, LGTM

erwincoumans:
Fixed a memory leak in this PR (delete m_solverDeformableBodyIslandCallback!)

PR #2554, opened 2019-12-24
fix sorting bug in btAxisSweep3Internal.h

xu-xionglong:
2 situations may lead to wrong sorting result in btAxisSweep3Internal::updateHandle

situation1:
The resorting of pHandle->m_minEdges[axis] change the value of pHandle->m_maxEdges[axis], so the value of emax is no longer equals to pHandle->m_maxEdges[axis].

situation2:
it' s hard to describe in text so I have made a example
edge values before updating: [0, 1, 2, 3]
after change edge[1] to 5 and edge[2] to 6: [0, 5, 6, 3]
try to do sortMinUp(emin = 1) and you will find that the value of pHandleNext->m_minEdges[axis] does not increase while it ought to.

to fix this bug, the modification of edge value and resorting should be done step by step.

### DIFF
--- a/src/BulletCollision/BroadphaseCollision/btAxisSweep3Internal.h
+++ b/src/BulletCollision/BroadphaseCollision/btAxisSweep3Internal.h
@@ -721,27 +721,23 @@ void btAxisSweep3Internal<BP_FP_INT_TYPE>::updateHandle(BP_FP_INT_TYPE handle, c
 	for (int axis = 0; axis < 3; axis++)
 	{
 		BP_FP_INT_TYPE emin = pHandle->m_minEdges[axis];
+        int dmin = (int)min[axis] - (int)m_pEdges[axis][emin].m_pos;
+        m_pEdges[axis][emin].m_pos = min[axis];
+        
+        if (dmin < 0)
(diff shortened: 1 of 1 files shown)


erwincoumans:
Can you please create a test (a small C++ example code) that currently fails the broadphase?
Note that min handles and max handles are automatically sorted, since one is always even and the other always odd.

xu-xionglong:
I found that what really mess up the sorting is the abnormal value in aabbMin and aabbMax. Is it possible that values in aabbMin are bigger than aabbMax? I will try to figure out what cause this if aabbMin should always smaller than aabbMax, or else I will make an example code to show how smaller aabbMax mess up the sorting.

erwincoumans:
aabbMin should always be smaller than aabbMax. If you have a repro case (or other bug) related to the broadphase, please share it. Thanks!

xu-xionglong:
I am sorry that it seems the bug comes from the game engine I used. It create some collision shape with negative extend, which cause the final error.

erwincoumans:
Thanks for trying to help anyhow. You may want to consider btDbvtBroadphase instead, it is often faster and accelerates ray casts better.

Issue #2002, opened 2018-11-22
[Multithreading] btCollisionDispatcherMt doesn't work with Compound vs Compound


mburetorp:
When trying to get multithreading up and running I found that compound vs compound don't collide at all if you use btCollisionDispatcherMt.

The problem appears to be that `algo->getAllContactManifolds(m_manifoldsPtr);` in `btCollisionDispatcherMt::dispatchAllCollisionPairs` never returns any manifolds. Also when I shutdown bullet during this penetration phase I get a crash in `btCollisionDispatcherMt::releaseManifold` since there are no manifolds in the array at that point.

My test scene is a static compound heightfield shape vs a dynamic compound sphere shape. If I remove the wrapper compound from the dynamic sphere it collides just fine.

erwincoumans:
Thanks for the report. Probably best to enable the AABB tree, since there are no volunteers working on this. Or fix it and consider a pull request. Thanks!

PR #2558, opened 2019-12-30
[crash fix for btCollisionDispatcherMt] a better approach to merge new manifolds when using the jobs CPU dispatcher

Lopuska:
Hi, after several ways I found to resolve the issue, I think this is the most elegant as it only requires few minimal changes in the multi threaded dispatcher (which is the only place the bug happens) implementation without touching anything else.

Notes:
- The crash only happens using btHashedOverlappingPairCache probably because btSortedOverlappingPairCache uses a deferred pair removal policy by default.
- This PR might resolve a similar issue already reported in the past which I believe is related to #2002 

The scenario is the following:
```
mOverlappingPairCache = new btHashedOverlappingPairCache();
mOverlappingPairCache->setInternalGhostPairCallback( mGhostPairCallback );
mDynamicsWorld = new btDiscreteDynamicsWorldMt( mDispatcher, mBroadphase, mSolverPool, mSolver, mCollisionConfiguration );
```
Then, push 2 objects in the world, in my case I have the Capsule and a Box and make them to collide to trigger the collision pair cache creation and starts to collect manifolds information.
mDynamicsWorld->stepSimulation( TickSec * mEditorTimeScale, smPhysicsMaxSubSteps, smPhysicsStepTime );

then what I do is to perform a manual call to my ghost object:
`collWorld->getDispatcher()->dispatchAllCollisionPairs(   mGhostObject->getOverlappingPairCache(), collWorld->getDispatchInfo(), collWorld->getDispatcher() );`

In the next frame the `mDynamicsWorld->stepSimulation` should trigger an assert then a crash in:
`void btCollisionDispatcherMt::releaseManifold -> btAssert(findIndex < m_manifoldsPtr.size());`

What happens is that in the multi-threaded implementation of dispatchAllCollisionPairs we are clearing the m_manifoldsPtr array object which collects all the manifolds from any pair cache and then query again all the contact manifold using getAllContactManifolds and not assuming that the object can collect any manifolds from any pair cache which can be created and dispatched from user code (in my case, the ghost object). This causes some manifolds to get cleared from m_manifoldsPtr array object before the pair cache invoke the deletion, then when the deletion is triggered you reach that assert because the element has not been found in the array.
For a reason I didn't really investigate however,  as I said before, the problem doesn't happen if a btSortedOverlappingPairCache is used probably because of the deferred removal policy, but it might causes missing collision as described in #2002 

The fix consists on avoiding to defer the manifold collecting operation to the main thread, instead a local array per thread is used to collect manifolds and then merged together to the main thread when all jobs are finished, keeping everything lockless, fast as before and without the crash.

Hope it helps

### DIFF
--- a/src/BulletCollision/CollisionDispatch/btCollisionDispatcherMt.cpp
+++ b/src/BulletCollision/CollisionDispatch/btCollisionDispatcherMt.cpp
@@ -28,6 +28,7 @@ subject to the following restrictions:
 btCollisionDispatcherMt::btCollisionDispatcherMt(btCollisionConfiguration* config, int grainSize)
 	: btCollisionDispatcher(config)
 {
+	m_batchManifoldsPtr.resize(btGetTaskScheduler()->getNumThreads());
 	m_batchUpdating = false;
 	m_grainSize = grainSize;  // iterations per task
 }
@@ -64,6 +65,10 @@ btPersistentManifold* btCollisionDispatcherMt::getNewManifold(const btCollisionO
 		//btAssert( !btThreadsAreRunning() );
 		manifold->m_index1a = m_manifoldsPtr.size();
 		m_manifoldsPtr.push_back(manifold);
+	}
+	else
+	{
+		m_batchManifoldsPtr[btGetCurrentThreadIndex()].push_back(manifold);
 	}
 
 	return manifold;
@@ -121,7 +126,7 @@ struct CollisionDispatcherUpdater : public btIParallelForBody
 
 void btCollisionDispatcherMt::dispatchAllCollisionPairs(btOverlappingPairCache* pairCache, const btDispatcherInfo& info, btDispatcher* dispatcher)
 {
-	int pairCount = pairCache->getNumOverlappingPairs();
+	const int pairCount = pairCache->getNumOverlappingPairs();
 	if (pairCount == 0)
 	{
 		return;
@@ -136,16 +141,17 @@ void btCollisionDispatcherMt::dispatchAllCollisionPairs(btOverlappingPairCache* 
 	btParallelFor(0, pairCount, m_grainSize, updater);
 	m_batchUpdating = false;
 
-	// reconstruct the manifolds array to ensure determinism
-	m_manifoldsPtr.resizeNoInitialize(0);
-
-	btBroadphasePair* pairs = pairCache->getOverlappingPairArrayPtr();
-	for (int i = 0; i < pairCount; ++i)
+	// merge new manifolds, if any
+	for (int i = 0; i < m_batchManifoldsPtr.size(); ++i)
 	{
-		if (btCollisionAlgorithm* algo = pairs[i].m_algorithm)
+		btAlignedObjectArray<btPersistentManifold*>& batchManifoldsPtr = m_batchManifoldsPtr[i];
+
+		for (int j = 0; j < batchManifoldsPtr.size(); ++j)
 		{
-			algo->getAllContactManifolds(m_manifoldsPtr);
+			m_manifoldsPtr.push_back(batchManifoldsPtr[j]);
 		}
+
+		batchManifoldsPtr.resizeNoInitialize(0);
 	}
 
 	// update the indices (used when releasing manifolds)
--- a/src/BulletCollision/CollisionDispatch/btCollisionDispatcherMt.h
+++ b/src/BulletCollision/CollisionDispatch/btCollisionDispatcherMt.h
@@ -30,7 +30,7 @@ public:
 	virtual void dispatchAllCollisionPairs(btOverlappingPairCache* pairCache, const btDispatcherInfo& info, btDispatcher* dispatcher) BT_OVERRIDE;
 
 protected:
-	btAlignedObjectArray<btAlignedObjectArray<btPersistentManifold*>> m_batchManifoldsPtr;
+	btAlignedObjectArray<btAlignedObjectArray<btPersistentManifold*> > m_batchManifoldsPtr;
 	bool m_batchUpdating;
 	int m_grainSize;
 };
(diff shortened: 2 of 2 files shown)


Lopuska:
Sadly we don't have a move operator defined in the btAlignedObjectArray, moving containers could significantly increase merging Progress, but requires C++11

Lopuska:
Hi @erwincoumans , does travis build is failing because of my changes? appveyor says successful ending

erwincoumans:
Thanks for the report. Can you briefly describe steps how to reproduce the issue  (in a few sentences sharing a (modified) demo in a zip file, not a long story)?

Lopuska:
I've been able to reproduce it by using the demo BasicExample from master branch.
Just replacing those 2 files should allows you to reproduce it easily.
[demo.zip](https://github.com/bulletphysics/bullet3/files/4017743/demo.zip)

dyanikoglu:
I think I had exactly same issue described in this PR with my custom engine implementation. I applied a different fix to get rid of compound-compound collision problem, and random crashes.

I will give your pr a try today, and let you know if it works out!

Lopuska:
Thanks a lot @dyanikoglu for taking time and test it.
I'm pretty sure that this is the proper fix, at least the one which involves only minimal changes on interested part.
The multithreaded dispatcher seems not widely used, that's why there is a very little report on this issue.
Waiting for your feedback!
Cheers

erwincoumans:
thanks for the repro!
Yes, travis ci fails because of this pr, it tests linux and mac. appveyor tests windows.

erwincoumans:
Here is a snippet of the error, just add a space between the >>

~~~~~~~~~~
/home/travis/build/bulletphysics/bullet3/src/BulletCollision/CollisionDispatch/btCollisionDispatcherMt.cpp:16:0:
/home/travis/build/bulletphysics/bullet3/src/BulletCollision/CollisionDispatch/btCollisionDispatcherMt.h:33:65: error: ‘>>’ should be ‘> >’ within a nested template argument list
  btAlignedObjectArray<btAlignedObjectArray<btPersistentManifold*>> m_batchManifoldsPtr;
~~~~~~~~~~

Lopuska:
Thanks a lot Erwin, I forgot about this C++03 thing.. eheh

erwincoumans:
Where are the manifolds removed from the m_batchManifoldsPtr at releaseManifold? The original code used m_index1a for this.

Lopuska:
that's still the same.
it's here:

https://github.com/bulletphysics/bullet3/blob/de9ac113c49a0c9724b9eb027ec157f1155c5526/src/BulletCollision/CollisionDispatch/btCollisionDispatcherMt.cpp#L87

there is a swap function done probably on purpose to remove the element with the pop_back. it's the only place where a manifold is removed.
The problem comes when you try to remove a manifold which has already been erased from the array caused by him:

https://github.com/bulletphysics/bullet3/blob/88826da6b532b8c06f4a55a07bcb8058d7a22789/src/BulletCollision/CollisionDispatch/btCollisionDispatcherMt.cpp#L140

The caller of releaseManifold should be the overlapping pair cache object which is proposed in 2 implementation. The key critical issue is when client code calls collWorld->getDispatcher()->dispatchAllCollision(...)
because this call invokes the m_manifoldsPtr.resizeNoInitialize(0);
but the pair cache still hold a references to elements of this object (through m_index1a).
I think this is the reason you miss collisions or when manifols is released you get the assert: because the pair cache hold the array index (m_index1a) of a manifold which has already been removed in un "unexpected" way from client code.

erwincoumans:
where is the manifold removed in case of m_batchUpdating?

Lopuska:
it's still in the same function (releaseManifold)
because this flag m_batchUpdating is true only during the parallelFor (which has a barrier included)
When you get the crash in the mt implementation the flag is false, probably because this manifold is released at a different time and not during the parallel for execution.

erwincoumans:
I'm not getting crashes, and I've never used the multitreaded code path (and no plans to do so).
I just wanted to confirm that the code makes sense.

I don't see any line in 'releaseManifold' that remove a manifold from m_batchManifoldsPtr.
It seems the manifold is removed from m_batchManifoldsPtr inside btCollisionDispatcherMt::dispatchAllCollisionPairs, can you confirm?

So if the manifold is added to m_batchManifoldsPtr, then you remove the object (which would release the manifold) and then call step simulation, the manifold is still in m_batchManifoldsPtr, right?

Lopuska:
Sorry, I missed the object name you were referring to, causing confusion between m_batchManifoldsPtr and m_manifoldsPtr.

> So if the manifold is added to m_batchManifoldsPtr, then you remove the object (which would release the manifold) and then call step simulation, the manifold is still in m_batchManifoldsPtr, right?

The manifold is added only temporary here (m_batchManifoldsPtr) during the stepSimulation in the Batch update, then transfered to the m_manifoldsPtr after when the batch is finished to keep the state consistent in order to make releaseManifold work as the caller expects.

> I don't see any line in 'releaseManifold' that remove a manifold from m_batchManifoldsPtr.
It seems the manifold is removed from m_batchManifoldsPtr inside btCollisionDispatcherMt::dispatchAllCollisionPairs, can you confirm?

Correct! every btCollisionDispatcherMt::dispatchAllCollisionPairs guarantees that m_batchManifoldsPtr gets cleared (after its content is merged to m_manifoldsPtr). It's basically a per Batch copy. The idea was to replicate the same behavior we have in the non parallel version (btCollisionDispatcher::dispatchAllCollisionPairs)

Lopuska:
If you want to verify the case personally by reproducing it, I can show you a quick video in order to reproduce the issue, by starting to the .zip file I attached last week.

erwincoumans:
Thanks for the PR

PR #2561, opened 2019-12-31
Remove unused custom vector math cmake option

Qix-:
This option is not used anywhere in the codebase. Seems to have hung around for over a decade at this point - came across it whilst configuring and was curious what it is.

Judging from some old articles on game engine designers weighing bullet and ODS, I'd wager it had to do with the new SIMD vector library that Bullet used a while back. But that's just a guess.

Anyway, doesn't appear to be needed anymore. Figured I'd help remove _some_ cruft :)


erwincoumans:
Thanks, it is not used indeed. A recent new project, FEMFX by AMD uses this vectormath library though:
https://github.com/GPUOpen-Effects/FEMFX

PR #2563, opened 2019-12-31
add btCollisionDispatcherMt.cpp on unity build

Lopuska:
to avoid linking Errors when using Unity builds.
Other *Mt source files such as btDiscreteDynamicsWorldMt.cpp, are already included.


Lopuska:
@erwincoumans , any hint on why travis failed on this?

Lopuska:
yay! thanks

PR #2565, opened 2020-01-02
Add Stable-Baselines example with SAC and TD3

araffin:
This PR adds a [Stable-Baselines](stable-baselines.readthedocs.io/) example with pybullet envs (using SAC and TD3 algorithms).
It is a simplified version of the [rl zoo](https://github.com/araffin/rl-baselines-zoo) with tuned hyperparameters (it should solve `HalfCheetahBulletEnv-v0` in less than 1M steps, i.e. reach a reward > 2500)

Colab notebook: https://colab.research.google.com/drive/1nZkHO4QTYfAksm9ZTaZ5vXyC7szZxC3F

For now, there is only the training script, i can add the enjoy script too if needed ;)

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/stable_baselines/train.py
+++ b/examples/pybullet/gym/pybullet_envs/stable_baselines/train.py
@@ -1,6 +1,6 @@
 # Code adapted from https://github.com/araffin/rl-baselines-zoo
 # it requires stable-baselines to be installed
-# Colab Notebook:
+# Colab Notebook: https://colab.research.google.com/drive/1nZkHO4QTYfAksm9ZTaZ5vXyC7szZxC3F
 # Author: Antonin RAFFIN
 # MIT License
 
--- a/examples/pybullet/gym/pybullet_envs/stable_baselines/train.py
+++ b/examples/pybullet/gym/pybullet_envs/stable_baselines/train.py
@@ -3,18 +3,9 @@
 # Colab Notebook: https://colab.research.google.com/drive/1nZkHO4QTYfAksm9ZTaZ5vXyC7szZxC3F
 # Author: Antonin RAFFIN
 # MIT License
-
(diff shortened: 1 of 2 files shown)


erwincoumans:
Thanks, I'll check it out. I added a simpler CartPoleContinuousBulletEnv-v0 that trains in a minute and renders in Colab:
https://colab.research.google.com/drive/15JSROMJbeiqxcUwifPR2NYeeFBKmyIlX#scrollTo=E2eWDjPZsQc5
will send you a PR in return to add those hyper params for zoo

araffin:
> I added a simpler CartPoleContinuousBulletEnv-v0 that trains in a minute and renders in Colab

Do you prefer me to submit a simpler version with those environments? I know that the current version is complex but I wanted to show the best practices (non breaking markov assumption and evaluating with a separate test env).

erwincoumans:
Thanks for the example, let's add it.

erwincoumans:
Thanks. Do you have some commands to train and enjoy, assuming this CL would be part of pypi (pip install pybullet)?

Note that when you install pybullet using pip, the train and enjoy scripts are hidden, so you would call them as part of pybullet_envs like this:

python3 -m pybullet_envs.stable_baselines.train.py and
python3 -m pybullet_envs.stable_baselines.enjoy.py

Then the data is usually put in the pybullet_data folder, and you can get the data from pybullet_data.getDataPath()

For example, for the deep_mimic example (we still need to turn this into a proper OpenAI gym environment)
pip3 install pybullet
python -m pybullet_envs.deep_mimic.testrl --arg_file run_humanoid3d_walk_args.txt

araffin:
>Do you have some commands to train and enjoy,

Good question, I would assume 
```
python3 -m pybullet_envs.stable_baselines.train
```
should work if pybullet is installed but I did not tested it yet.
I will do a PR for the enjoy script ;)

PR #2576, opened 2020-01-09
[BUGFIX] Update existing deformable gravity forces on gravity change.

dmcconachie:
Without this change, it matters what order you call `setGravity` versus call `loadSoftBody`. Tested on `deformable_anchor.py`.


erwincoumans:
Thanks for the fix, it makes sense.
(close/re-open to trigger CI, one Mac OSX target is flacky)

PR #2577, opened 2020-01-10
Add position error into deformable vs. rigid contact

xhan0619:
Add position error into deformable vs. rigid contact where the error is added directly to the constraint (i.e. not splitImpulse). 

Change the default damping model for mass pring to the angular momentum one for more accuracy and visual correctness.

Fix a bug in the interpolation transform of the rigid body collision wrapper.

### DIFF
--- a/src/BulletSoftBody/btDeformableContactConstraint.cpp
+++ b/src/BulletSoftBody/btDeformableContactConstraint.cpp
@@ -79,7 +79,7 @@ btVector3 btDeformableNodeAnchorConstraint::getVa() const
     return va;
 }
 
-btScalar btDeformableNodeAnchorConstraint::solveConstraint()
+btScalar btDeformableNodeAnchorConstraint::solveConstraint(const btContactSolverInfo& infoGlobal)
 {
     const btSoftBody::sCti& cti = m_anchor->m_cti;
     btVector3 va = getVa();
@@ -141,7 +141,7 @@ btDeformableRigidContactConstraint::btDeformableRigidContactConstraint(const btS
     m_total_normal_dv.setZero();
     m_total_tangent_dv.setZero();
     // The magnitude of penetration is the depth of penetration.
-    m_penetration = c.m_cti.m_offset;
+    m_penetration = btMin(btScalar(0),c.m_cti.m_offset);
 }
 
 btDeformableRigidContactConstraint::btDeformableRigidContactConstraint(const btDeformableRigidContactConstraint& other)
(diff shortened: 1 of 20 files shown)


erwincoumans:
Thanks a lot for all the improvements. Let me review the changes and get back to you.

src/BulletSoftBody/btDeformableContactConstraint.cpp
```diff
@@ -256,6 +256,8 @@ btScalar btDeformableRigidContactConstraint::solveConstraint()
     impulse = impulse_normal + impulse_tangent;
     // apply impulse to deformable nodes involved and change their velocities
     applyImpulse(impulse);
+	if (residualSquare < 1e-7)
```
erwincoumans: Could you describe the idea behind this early out?

src/BulletSoftBody/btSoftBody.cpp
```diff
@@ -3681,8 +3698,8 @@ void btSoftBody::defaultCollisionHandler(const btCollisionObjectWrapper* pcoWrap
                     docollideFace.psb = this;
                     docollideFace.m_colObj1Wrap = pcoWrap;
                     docollideFace.m_rigidBody = prb1;
-                    docollideFace.dynmargin = basemargin + timemargin;
-                    docollideFace.stamargin = basemargin;
+					docollideFace.dynmargin = 0.05*(basemargin + timemargin);
```
erwincoumans: Since it is still in research stage, I'll LGTM, , Do we want to expose those constants?

erwincoumans:
Thanks for the contribution!
I'm fine with merging the code. Some of the features are conditionally compiled, we may want to enable/disable them at run-time? 
(#ifdef USE_STRAIN_RATE_LIMITING and #define USE_QUADRATURE 1, //#define CACHE_PREV_COLLISION etc)

src/BulletSoftBody/btDeformableContactConstraint.cpp
xhan0619: This is actually a bug, not a serious one though. I'll fix this in the next PR.

src/BulletSoftBody/btSoftBody.cpp
xhan0619: Yes, eventually we do. But I'm still trying to understand why a smaller margin for face than for nodes helps with reducing stickiness.

xhan0619:
> CACHE_PREV_COLLISION

Yes, we can change them to run-time flags.

PR #2589, opened 2020-01-17
Adds support for loading user data from URDF files.

RanTig:
User data should be specified inside a <bullet> tag, with a <user-data> tag.
The user-data tag should also specify a 'key' attribute.

### DIFF
--- a/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
+++ b/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
@@ -31,6 +31,7 @@
 #include "BulletCollision/NarrowPhaseCollision/btPersistentManifold.h"
 #include "Bullet3Common/b3HashMap.h"
 #include "../Utils/ChromeTraceUtil.h"
+#include "SharedMemoryPublic.h"
 #include "stb_image/stb_image.h"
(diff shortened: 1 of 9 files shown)


examples/SharedMemory/PhysicsServerCommandProcessor.cpp
```diff
@@ -3013,6 +3014,46 @@ void PhysicsServerCommandProcessor::createJointMotors(btMultiBody* mb)
...
+		body->m_userDataHandles.push_back(userDataHandle);
+	}
+	userData->replaceValue(valueBytes, valueLength, valueType);
+	return userDataHandle;
+}
+
+void PhysicsServerCommandProcessor::addUserData(const btHashMap<btHashString, std::string>& user_data_entries, int bodyUniqueId, int linkIndex, int visualShapeIndex) {
+	for (int i = 0; i < user_data_entries.size(); ++i) {
+		const std::string key = user_data_entries.getKeyAtIndex(i).m_string1;
+		const std::string* value = user_data_entries.getAtIndex(i);
```
erwincoumans: best to add a check if value is valid, to avoid crash when using the pointer.

erwincoumans:
Thanks for the PR, it looks good to me.

PR #2594, opened 2020-01-23
Adds an option to syncUserData to specify the bodies for which to sync.

RanTig:
When the user data of a few bodies are modified (e.g. one new object is added to a large scene), it is useful to only sync the user data of a subset of bodies.

If no bodies are specified, the old behavior of synchronizing the complete state remains.


erwincoumans:
I went through the changes and LGTM, thanks for the contribution!

PR #2602, opened 2020-01-31
Fix indexing in obj2sdf

fuchuyuan:


erwincoumans:
Great, looks like the issue is resolved by this patch. Thanks!

fuchuyuan:
I think we can merge this now if it solves the issue.

PR #2605, opened 2020-02-03
Only clear the user data cache when a full user data sync is requested.

RanTig:
When syncing a single user data entry, we shouldn't clear the client-side user data cache.


erwincoumans:
LGTM, thanks!

PR #2620, opened 2020-02-13
apply custom pair-cache filter to predictive contacts

stephengold:
If the pair cache has a custom overlap filter (set using `setOverlapFilterCallback()`), apply that same filter to the predictive contacts generated for continuous collision detection (CCD).


stephengold:
What's needed in order for this to pull request be integrated into the repository?

erwincoumans:
Can you remove the  comments with your name from the source please.
I don't have time yet to make sure things don't break in various proprietary code bases using Bullet.

erwincoumans:
triggering CI (close/re-open issue)

erwincoumans:
Thanks, it looks good to me. Thanks for the contribution.

PR #2623, opened 2020-02-14
Put TinyRender functions and classes inside a TinyRender namespace.

marcel303:
The TinyRender library defines some classes with very generic names, such as vec, Matrix, Model. In my case, when compiling the example browser using a different library for graphics, this was giving me issues with multiply defined symbols.


erwincoumans:
thanks, looks good to me.

PR #2627, opened 2020-02-15
Add enjoy script for Stable Baselines

araffin:
This is a follow up to #2565 :
- it fixes call from the module
- it adds the enjoy script

You can use:
```
python3 -m pybullet_envs.stable_baselines.train --algo sac --env HalfCheetahBulletEnv-v0
# If the saved model "sac_HalfCheetahBulletEnv-v0.zip" is in the current folder
python3 -m pybullet_envs.stable_baselines.enjoy --algo sac --env HalfCheetahBulletEnv-v0 --n-episodes 5
```

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/stable_baselines/enjoy.py
+++ b/examples/pybullet/gym/pybullet_envs/stable_baselines/enjoy.py
@@ -1,0 +1,86 @@
+# Code adapted from https://github.com/araffin/rl-baselines-zoo
+# it requires stable-baselines to be installed
+# Colab Notebook: https://colab.research.google.com/drive/1nZkHO4QTYfAksm9ZTaZ5vXyC7szZxC3F
+# You can run it using: python -m pybullet_envs.stable_baselines.enjoy --algo td3 --env HalfCheetahBulletEnv-v0
+# Author: Antonin RAFFIN
+# MIT License
+import argparse
+import multiprocessing
+
+import gym
+import numpy as np
+import pybullet_envs
+
+from stable_baselines import SAC, TD3
+from stable_baselines.common.vec_env import SubprocVecEnv
+from stable_baselines.common.evaluation import evaluate_policy
+
+from pybullet_envs.stable_baselines.utils import TimeFeatureWrapper
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser("Enjoy an RL agent trained using Stable Baselines")
+    parser.add_argument('--algo', help='RL Algorithm (Soft Actor-Critic by default)', default='sac',
+                        type=str, required=False, choices=['sac', 'td3'])
+    parser.add_argument('--env', type=str, default='HalfCheetahBulletEnv-v0', help='environment ID')
+    parser.add_argument('-n', '--n-episodes', help='Number of episodes', default=5,
+                        type=int)
+    parser.add_argument('--no-render', action='store_true', default=False,
+                    help='Do not render the environment')
+    args = parser.parse_args()
+
+    env_id = args.env
+    # Create an env similar to the training env
+    env = TimeFeatureWrapper(gym.make(env_id))
+
+    # Use SubprocVecEnv for rendering
+    if not args.no_render:
+        # Note: fork is not thread-safe but usually is faster
+        fork_available = 'fork' in multiprocessing.get_all_start_methods()
+        start_method = 'fork' if fork_available else 'spawn'
+        env = SubprocVecEnv([lambda: env], start_method=start_method)
+
+
+    algo = {
+        'sac': SAC,
+        'td3': TD3
+    }[args.algo]
+
+    # We assume that the saved model is in the same folder
+    save_path = '{}_{}.zip'.format(args.algo, env_id)
+    # Load the saved model
+    model = algo.load(save_path, env=env)
+
+    try:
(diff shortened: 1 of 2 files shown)


erwincoumans:
Thanks for the contribution! LGTM

erwincoumans:
The zip file is called sac_HalfCheetahBulletEnv-v0_best.zip and the enjoy script doesn't work unless you copy that file to sac_HalfCheetahBulletEnv-v0.zip 
Why doesn't it just work with sac_HalfCheetahBulletEnv-v0_best.zip?

araffin:
>The zip file is called sac_HalfCheetahBulletEnv-v0_best.zip and the enjoy script doesn't work unless you rename/copy that file to sac_HalfCheetahBulletEnv-v0.zip

That's because you did not wait until the end, no?
There will be two files: one creating at the end of training, one updated each time a best model is found.

erwincoumans:
Thanks! Indeed, no idea there was an automatic 'end'. I assumed terminating the training would be the end. Perhaps enjoy should try to use the _best if it cannot find the other file (with a warning that the results are from an intermediate policy that was not fully trained).

araffin:
Good point.
If you agree, i would do a PR that does like the zoo:
- it saves the model when you terminate the script
- it adds an option '--load-best' to load the best model

erwincoumans:
I'd like to see a fully automatic fallback to load the best, that doesn't require command line options. So it just loads the _best file, if it cannot find its file.

Also, the enjoy script would be nice to use PyBullet's own visualizer.
Simply call the 'render(mode='human') before the first reset, then you get a visualizer GUI window.
It looks better than the OpenCV one, and it actually works.

For me, the current enjoy script is more frustration:

~~~~~~~~~~~~~~
  File "C:\python-3.5.3.amd64_numpydebug\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\python-3.5.3.amd64_numpydebug\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\python-3.5.3.amd64_numpydebug\lib\pybullet_envs\stable_baselines\enjoy.py", line 69, in <module>
    env.render(mode='human')
  File "C:\python-3.5.3.amd64_numpydebug\lib\site-packages\stable_baselines\common\vec_env\subproc_vec_env.py", line 138, in render
    cv2.imshow('vecenv', bigimg[:, :, ::-1])
cv2.error: OpenCV(4.2.0) C:/projects/opencv-python/opencv/modules/highgui/src/precomp.hpp:137: error: (-215:Assertion failed) src_depth != CV_16F && src_depth != CV_32S in function 'convertToShow'
~~~~~~~~~~~~~~

erwincoumans:
Please check this enjoy script, from the CL (if it hasn't been applied yet)
https://github.com/bulletphysics/bullet3/pull/2649

erwincoumans:
Also, how does baselines determine that the training is ended?

araffin:
>I'd like to see a fully automatic fallback to load the best, that doesn't require command line options. So it just loads the _best file, if it cannot find its file.

ok, I'll do that

>Also, the enjoy script would be nice to use PyBullet's own visualizer.
Simply call the 'render(mode='human') before the first reset, then you get a visualizer GUI window

Oh, I did not know that... In fact, I've been struggling (cf zoo) with it. Using OpenCV was a hack.

>Also, how does baselines determine that the training is ended?

Good question.
It depends. For now, you give a budget (maximum number of interaction with the environment) but we would also use a callback that ends the training once a reward threshold is attained (cf the new [callback collection](https://stable-baselines.readthedocs.io/en/master/guide/callbacks.html#stoptrainingonrewardthreshold) that is now on master but not yet released).

erwincoumans:
It would help a lot to save weights during training at regular intervals (and perhaps also whenever the policy exceeds the current best). Is there a way to achieve this with the regular rl_zoo_baselines? Preferably this interval can be user specified (in terms of number of steps).
~~~~~~~~~~~~~
python3 train.py --algo ppo2 --env HumanoidDeepMimicWalkBulletEnv-v1 --tensorboard-log deepmimic
~~~~~~~~~~~~~

I'm testing a new environment and like to test various intermediate snapshots.

Also, the reward in Tensorboard doesn't seem to line up with avg episode reward in the console.

Finally, do the actions need to be normalized in range -1,1 for a custom env ? How about observations/rewards?

araffin:
>It would help a lot to save weights during training at regular intervals 

I've got good news for you, we just release a [callback collection](https://stable-baselines.readthedocs.io/en/master/guide/callbacks.html) and checkpoints are included. I will add that feature to the zoo soon.

>Also, the reward in Tensorboard doesn't seem to line up with avg episode reward in the console.

That's normal if you are normalizing the reward. The console should output the original mean reward.

>Finally, do the actions need to be normalized in range -1,1 for a custom env ? How about observations/rewards?

Answers for those questions are in the [tips and tricks](https://stable-baselines.readthedocs.io/en/master/guide/rl_tips.html#general-advice-when-using-reinforcement-learning) section of the documentation ;)
in short yes, and especially true (obs/reward normalization using `VecNormalize`) for A2C/PPO2.

araffin:
This will be added (checkpoints) in https://github.com/araffin/rl-baselines-zoo/pull/69

PR #2630, opened 2020-02-17
Include dlerror() in handling of failed dlopen()

vabr-g:
X11OpenGLWindow::X11OpenGLWindow() ends up calling dlopen() to get libX11.
When this fails, it only reports that it failed, and what filename it was looking for.
This commit adds dlerror() to the error message, which makes investigating failures easier.


erwincoumans:
LGTM, thanks!

PR #2634, opened 2020-02-18
use safeNormalize() to avoid assert failure in btVector3::normalize()

stephengold:
I had a `btKinematicCharacterController` that was reaching velocity=0 at the apex of its jump. With `#define _DEBUG`, that caused an assertion failure in `btVector3::normalize()`. Replacing `normalize()` with `safeNormalize()` avoids the crash.


erwincoumans:
thanks, LGTM.

PR #2637, opened 2020-02-19
Replace tensorflow with tf.compat.v1

drigz:
This means they won't break when using TensorFlow 2.0.


erwincoumans:
Thanks for the pull request, LGTM.

PR #2643, opened 2020-02-25
CMake reformat

madebr:
This pr consist of the cmake reformatting of #2389:

- Bump CMake version: bullet3 was already using functions from more recent cmake versions
- Use targets of CMake by using `TARGET_LINK_LIBRARIES`
- Make the `install()` calls a bit easier to read by wrapping it.


madebr:
@erwincoumans Please review.

glebm:
@erwincoumans Looks like @madebr did good work here and has a solid understanding of CMake. Any chance you can have a look at this and the other CMake PRs?

@madebr Are you still using bullet CMake actively? Will you be able take care of following-up on bug reports if this is merged?

madebr:
I'm still using it + packaging it for conan-center-index.
If problems/issues arrive, I'd be happy to look at them.
But I think the main issue is that @erwincoumans is overloaded with other work to perform a thorough review.

eliasdaler:
This PR looks solid and makes Bullet usable with CMake - right now the targets don't define include dirs so linking to Bulet targets via `target_link_libraries` is not enough to use Bullet properly.

PR #2652, opened 2020-03-02
Support separate render and simulation obj & fix sleeping issue for deformable.

xhan0619:
1) Support loading two obj files for deformable simulation, one for rendering and the other for simulation. At initialization, each node n of the rendering mesh searches for the closest triangle f in the simulation mesh in the barycentric sense. Then we store the distance d from n to the plane containing f, and the projection of n, p. At render time, we calculate the position of n as p + normal_of_f * d.
2) Fix the deformable sleeping issue by preventing geometric collision for sleeping deformable objects
3) Minor bug fixes and formatting.


erwincoumans:
Thanks, LGTM.

PR #2661, opened 2020-03-09
Support Deformable Picking and Add Demo for Physics Parameters.

xhan0619:


examples/CommonInterfaces/CommonRigidBodyBase.h
```diff
@@ -7,7 +7,7 @@
 #include "CommonGUIHelperInterface.h"
 #include "CommonRenderInterface.h"
 #include "CommonCameraInterface.h"
-
+#include "BulletSoftBody/btSoftBody.h"
```
erwincoumans: Would it be possible to revert this change in CommonRigidBodyBase and move the btSoftBody.h into CommonDeformableBase.h?

src/BulletCollision/CollisionDispatch/btCollisionWorld.h
```diff
@@ -263,6 +263,36 @@ class btCollisionWorld
...
 		}
 	};
+    
+    struct ClosestRayResultCallbackWithInfo : public ClosestRayResultCallback
+    {
+        ClosestRayResultCallbackWithInfo(const btVector3& rayFromWorld, const btVector3& rayToWorld)
+        : ClosestRayResultCallback(rayFromWorld, rayToWorld)
+        {
+        }
+        LocalShapeInfo* m_localShapeInfo;
```
erwincoumans: this m_localShapeInfo has a very short lifetime, so best not to keep it persistent like this, it is intended only during the callback and not any longer, I think.

erwincoumans: Also, is is ok to keep this class local to the demo?

erwincoumans:
Some of the demos crashes sometimes when picking (for example cloth friction demo), need to track down why.

examples/CommonInterfaces/CommonRigidBodyBase.h
erwincoumans: The SoftBody demo crashes now (due to changes in CommonRigidBodyBase.h), could you remove those changes (there should be no btSoftBody.h in that class)

erwincoumans:
The crashes are indeed due to storing the pointer to m_localShapeInfo, please store an object instead.
Also, you need to update the pick position to most recent, when setting the constraint.
Remove the changes in CommonRigidBodyBase.

src/BulletCollision/CollisionDispatch/btCollisionWorld.h
xhan0619: Done.

erwincoumans: the change in btCollisionWorld.h seems to be still there, can you revert that file?

examples/CommonInterfaces/CommonRigidBodyBase.h
erwincoumans: please remove line

src/BulletCollision/CollisionDispatch/btCollisionWorld.h
xhan0619: I already did. I'm fixing another issue revealed by mouse picking. Will push soon.

examples/CommonInterfaces/CommonRigidBodyBase.h
xhan0619: Done.

xhan0619: Done.

erwincoumans:
Thanks a lot for the CL, LGTM.

PR #2667, opened 2020-03-12
Teleport deformables

xhan0619:
Move deformable objects through InitPose Command.

### DIFF
--- a/examples/CommonInterfaces/CommonDeformableBodyBase.h
+++ b/examples/CommonInterfaces/CommonDeformableBodyBase.h
@@ -1,0 +1,219 @@
+
+#ifndef COMMON_DEFORMABLE_BODY_SETUP_H
+#define COMMON_DEFORMABLE_BODY_SETUP_H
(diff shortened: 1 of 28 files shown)


erwincoumans:
Thanks for the PR!
Looks like some error:
src/LinearMath/btImplicitQRSVD.h:474:54: error: ‘numeric_limits’ is not a member of ‘std’
                            const btScalar tol = 64 * std::numeric_limits<btScal

also this warning, perhaps initialized optimal_dist?

src/BulletSoftBody/btSoftBodyHelpers.cpp:1611:35: warning: 'optimal_dist' may be used uninitialized in this function [-Wmaybe-uninitialized]
         psb->m_z[i] = optimal_dist;

erwincoumans:
Ah, there are some conflicts with the other CL (that is merged now). Could you check it out?

erwincoumans:
Too many changes to review, but since they are all related to deformables/soft body, let's merge now and make fixes if needed.

PR #2686, opened 2020-03-21
Make file formatting comply with POSIX and Unix standards

aaronfranke:
UTF-8, LF, and no BOM are all recommended for file formatting. These settings are the most compatible with various tools and also take up the least amount of space (CRLF -> LF in a file means saving one byte per line, and no BOM saves two bytes per file). This is especially important for `html` and `js` files because these are used for websites and a small file size is important.

* However, I did not use LF in `.bat` files because CRLF is required for `.bat` files.

The POSIX standard for text files says that each text file should end with a newline, otherwise it is by definition not a text file. GitHub shows a warning when a file is committed without a terminating newline (see the "before" diffs of this PR), and GCC shows warnings when compiling files that don't end with newlines. See [this SO question](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline/25322168) and [this Unix SE question](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file) for more information. 

* However, I didn't change this for many of the file types related to 3D models because I wasn't sure about them, including `obj` and `mtl` and `urdf`.

There is one other thing, I changed the spacing in the example C# file for Unity since the brace style was inconsistent and there was also a mixing of tabs and spaces.


erwincoumans:
We have a clang_format file.
If you have suggestions how to modify this file, we can discuss those changes (instead of manually modifying various source files)

aaronfranke:
`clang-format` does not have a way to enforce newlines at the end of files.

From what I can tell `clang-format` already enforces LF line endings and the lack of a BOM, but this isn't done for non-C++ files such as the HTML files.

Looking at the list of languages supported by `clang-format`, I think the only places we could enforce this using `clang-format` are the C# files and the JavaScript files.

aaronfranke:
@erwincoumans Actually, when I run `clang-format-all.sh`, it produces 535 changed files (without this PR). What's going on? Have y'all not run `clang-format-all.sh` in a long time? Do you not run this script as part of your Travis CI builds? Why doesn't the script exclude the `ThirdPartyLibs` folder?

erwincoumansepic:
>> What's going on? 

Too busy with other stuff and didn't enforce the formatting.

>> Why doesn't the script exclude the

No-one implemented this yet.

One issue is that running clang-format can cause a lot of merge conflict hassle, so until clang-format is part of CI, it is basically a once in a long while thing to run.

aaronfranke:
Closing because it does not appear that there is interest from maintainers in improving the file formatting of this repository. If such interest arises, it's probably easier for the maintainers to just run a formatter themselves. Or, if desired, you can ping me to open a new PR with these changes.

PR #2688, opened 2020-03-22
Add CheckpointCallback and load best automatically

araffin:
Address comments from https://github.com/bulletphysics/bullet3/pull/2627#issuecomment-594637933:

- Save model on ctrl+c
- Load best model automatically if no model is found
- Use the callback collection from SB v2.10.0
- Allow to save checkpoints


erwincoumans:
Thanks for the PR! Does this also save a model every eval_freq steps? That's pretty important in RL.

araffin:
> Does this also save a model every eval_freq steps?

It does if you specify `--save-freq` (deactivated by default), but there is no automatic loading of a checkpoint for now, you would need to rename it.

PR #2704, opened 2020-04-01
Compiler warnings.

stolk:
There is an instance of an actual return of a dangling pointer (stack data.) There is also a warning about theoretical buffer overruns.


erwincoumans:
bedankt, lgtm!

PR #2708, opened 2020-04-02
fix uninitialized variable in btSoftBody

xhan0619:
m_sst is not initialized at the time initializeFaceTree is called.


erwincoumans:
thanks for the fix!

PR #2711, opened 2020-04-02
Load urdf

fuchuyuan:
1. Add UrdfDeformable struct in UrdfParser.h and corresponding functions to parse this struct from URDF.

2. Move majority of the logic in processLoadSoftBodyCommand into processDeformable, so that loadURDF and processLoadSoftBodyCommand both use this function as the main loading logic.

3. Add a torus loaded from URDF in examples/pybullet/examples/deformable_torus.py to have two torus side by side, one loaded using the original loadSoftBody, the other using loadURDF.

### DIFF
--- a/data/torus_deform.urdf
+++ b/data/torus_deform.urdf
@@ -1,0 +1,13 @@
+<?xml version="1.0"?>
+<robot name="torus">
+  <deformable name="torus">
+    <inertial>
+      <mass value="1" />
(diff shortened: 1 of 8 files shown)


erwincoumans:
Thanks for the PR. I'm looking at it and need to test a few more things before LGTM.
(there are couple of uninitialized variables in the deformable code, unrelated to this PR)

erwincoumans:
thanks, LGTM.

PR #2714, opened 2020-04-03
Add checks. Remove unused member var.

stolk:


erwincoumans:
Looks reasonable, thanks for the PR!

PR #2722, opened 2020-04-07
improve EGL context creation

BlGene:
This picks the first working EGL context, which is usually the correct one.


erwincoumans:
Thanks! It seems without it, a software EGL implementation is chosen.

PR #2732, opened 2020-04-09
deactivate objects AFTER synchronizeMotionStates

AndrewMeadows:
This PR tries to fix a bug where sometimes the final `MotionState::setWorldTransform()` is not called after an active object has moved a little but was deactivated.  In particular it is trying to solve a problem described in [this forum thread](https://pybullet.org/Bullet/phpBB3/viewtopic.php?f=9&t=12921).

Dunno how safe this change is since it does subtly change the behavior of `btDiscreteDynamicsWorld::internalSingleStepSimulation()`, and accidentally swaps the order of `updateActivationState()` with the internal callback, but I figured I would submit it for review.  Feel free to discard, or suggest an alternative fix.


erwincoumans:
Thanks for the CL. I'll try to look at it.

PR #2745, opened 2020-04-15
fix rayTest crash for btSoftBody

xhan0619:
Initialize face DBVT tree at the time when deformable object is created to avoid data race.
Set the correct body id for deformable objects in raytest.


erwincoumans:
Thanks, LGTM.

PR #2748, opened 2020-04-15
fix #ifdef for old damping method in btRigidBody

pouleyKetchoupp:
As discussed in https://github.com/godotengine/godot/issues/19182 I've made some fixes to be able to use the old damping method.

When `BT_USE_OLD_DAMPING_METHOD` is defined:
- `setDamping` allows any non-negative value instead of clamping between 0 and 1
- Velocity changes use the old damping equation

### DIFF
--- a/src/BulletDynamics/Dynamics/btRigidBody.cpp
+++ b/src/BulletDynamics/Dynamics/btRigidBody.cpp
@@ -136,8 +136,13 @@ void btRigidBody::setGravity(const btVector3& acceleration)
 
 void btRigidBody::setDamping(btScalar lin_damping, btScalar ang_damping)
 {
-	m_linearDamping = btClamped(lin_damping, (btScalar)btScalar(0.0), (btScalar)btScalar(1.0));
-	m_angularDamping = btClamped(ang_damping, (btScalar)btScalar(0.0), (btScalar)btScalar(1.0));
+#ifdef BT_USE_OLD_DAMPING_METHOD
+	m_linearDamping = btMax(lin_damping, btScalar(0.0));
+	m_angularDamping = btMax(ang_damping, btScalar(0.0));
+#else
+	m_linearDamping = btClamped(lin_damping, btScalar(0.0), btScalar(1.0));
+	m_angularDamping = btClamped(ang_damping, btScalar(0.0), btScalar(1.0));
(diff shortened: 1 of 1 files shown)


pouleyKetchoupp:
CC @AndreaCatania

src/BulletDynamics/Dynamics/btRigidBody.cpp
```diff
@@ -146,10 +151,9 @@ void btRigidBody::applyDamping(btScalar timeStep)
 	//On new damping: see discussion/issue report here: http://code.google.com/p/bullet/issues/detail?id=74
 	//todo: do some performance comparisons (but other parts of the engine are probably bottleneck anyway
 
-//#define USE_OLD_DAMPING_METHOD 1
-#ifdef USE_OLD_DAMPING_METHOD
-	m_linearVelocity *= GEN_clamped((btScalar(1.) - timeStep * m_linearDamping), (btScalar)btScalar(0.0), (btScalar)btScalar(1.0));
-	m_angularVelocity *= GEN_clamped((btScalar(1.) - timeStep * m_angularDamping), (btScalar)btScalar(0.0), (btScalar)btScalar(1.0));
+#ifdef BT_USE_OLD_DAMPING_METHOD
+	m_linearVelocity *= btClamped((btScalar(1.0) - timeStep * m_linearDamping), btScalar(0.0), btScalar(1.0));
+	m_angularVelocity *= btClamped((btScalar(1.0) - timeStep * m_angularDamping), btScalar(0.0), btScalar(1.0));
```
AndreaCatania, line 156: You can just use max here, because it can't be more than 1, so we save one if condition.

```cpp
m_linearVelocity *= btMax(btScalar(1.0) - timeStep * m_linearDamping, btScalar(0.0));
m_angularVelocity *= btMax(btScalar(1.0) - timeStep * m_angularDamping, btScalar(0.0));
```

pouleyKetchoupp: Thanks, I've just pushed this change.

AndreaCatania: Super!

erwincoumans:
thanks, looks good to me.

PR #2756, opened 2020-04-21
Add support for KKT solver.

xhan0619:


erwincoumans:
If you added files, please add them to setup.py and all the CMakeLists.txt and premake4.lua project files.

erwincoumans:
Ah, in setup.py don't add the new file to egl plugin sources, but to the file bullet3\src\btLinearMathAll.cpp

erwincoumans:
thanks, LGTM.

PR #2762, opened 2020-04-23
Error running script in Windows

NiranthS:
Hello sir,
In Windows, without "p.setAdditionalSearchPath(pybullet_data.getDataPath())" line, we are not able to import "plane.urdf" and other pre existing urdf files.

### DIFF
--- a/examples/pybullet/examples/addPlanarReflection.py
+++ b/examples/pybullet/examples/addPlanarReflection.py
@@ -1,6 +1,8 @@
 import pybullet as p
 import time
 import math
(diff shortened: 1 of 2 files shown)


erwincoumans:
Thanks. Yes, the examples assume you trun them from the full repository, and in that case it should find the resources in the Bullet/data folder (instead of Bullet/examples/pybullet/gym/pybullet_data).

Using setAdditionalSearchPath is better, and we could include the examples with pybullet by default.

PR #2771, opened 2020-04-25
Bugfix for camera_adjust and move_and_look_at in locomotion environments

sgillen:
Fixes an issue I was experiencing trying to call camera_adjust on the pybullet_envs locomotion environments. Previously when running this snippet:

```python
import gym
import pybullet_envs

env = gym.make("Walker2DBulletEnv-v0", render=True)
env.reset()

env.camera_adjust()
```

I would get the following error:
```
  File "bullet_test.py", line 7, in <module>
    env.camera_adjust()
  File "/home/sgillen/miniconda3/envs/bullet_fork/lib/python3.6/site-packages/pybullet_envs/gym_locomotion_envs.py", line 128, in camera_adjust
    x, y, z = self.body_xyz
AttributeError: 'Walker2DBulletEnv' object has no attribute 'body_xyz'
```

The change I'm submitting is working for my use case, but I did not see any tests to run to make sure I did not break anything somewhere else (although it looks like camera_adjust is not currently used anywhere in pybullet). If you'd like though I can expand the above snippet.

---
This change is [Reviewable](https://reviewable.io/reviews/bulletphysics/bullet3/2771)

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/env_bases.py
+++ b/examples/pybullet/gym/pybullet_envs/env_bases.py
@@ -26,7 +26,7 @@ class MJCFBaseBulletEnv(gym.Env):
     self.scene = None
     self.physicsClientId = -1
     self.ownsPhysicsClient = 0
-    self.camera = Camera()
+    self.camera = Camera(self)
     self.isRender = render
     self.robot = robot
     self.seed()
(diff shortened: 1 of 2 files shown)


sgillen:
Hmm, my initial commit caused the build to fail under gcc on macOS. However reverting and then unreverting this commit resulted in a passing build, this seems to me to point to a problem somewhere somewhere in the pipeline unrelated to this pull request? 

Also happy to squash this down before merging (if you decide to merge this) .

erwincoumans:
Thanks, LGTM.

PR #2774, opened 2020-04-28
updateVisualShape for softbody

fuchuyuan:


erwincoumans:
thanks for the contribution, looks good to me.

PR #2775, opened 2020-04-28
Split impulse

xhan0619:
Add split impulse for deformable objects vs. rigid objects.
Allow deformable face in collision with rigid objects to enter constraint solve even if they have separating velocity.
Change contact constraint solve to inequality constraints.
Improve parameters for a few demos.
Expose API for setting repulsion stiffness.


erwincoumans:
Thanks for the contribution! Could you make a small change, moving the physicsClientId as last arg, and modify the type string to have i at the end).

PR #2783, opened 2020-05-06
Remove unused thread-unsafe debug variable 'maxIteration' …

sippeyxp:
Remove unused thread-unsafe debug variable 'maxIteration' that is no longer used in btQuantizedBvh.cpp


erwincoumans:
Thanks, LGTM (waiting for CI to finish, before merge)

PR #2787, opened 2020-05-07
Update externalTorqueControlledSphere.py

NiranthS:


erwincoumans:
Thanks, no need to have those double calls

PR #2788, opened 2020-05-08
Write face mesh to obj and fix rigid transform for softbody

fuchuyuan:


erwincoumans:
Thanks for the update. I trust you are familiar with the SVD signs, so let's merge it.

PR #2797, opened 2020-05-11
Step simulation and conversion between force and impulse use the same dt

fuchuyuan:
This commit fixes the inconsistency for getting contact force, and setting max joint torque when using substeps.

### DIFF
--- a/examples/Importers/ImportURDFDemo/UrdfParser.cpp
+++ b/examples/Importers/ImportURDFDemo/UrdfParser.cpp
@@ -1242,9 +1242,9 @@ bool UrdfParser::parseDeformable(UrdfModel& model, tinyxml2::XMLElement* config,
 			logger->reportError("expected a filename for collision geoemtry");
 			return false;
 		}
-		fn = vis_xml->Attribute("filename");
+		fn = col_xml->Attribute("filename");
 		success = UrdfFindMeshFile(m_fileIO,
-								   model.m_sourceFile, fn, sourceFileLocation(vis_xml),
+								   model.m_sourceFile, fn, sourceFileLocation(col_xml),
 								   &deformable.m_simFileName, &out_type);
 
 		if (!success)
--- a/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
+++ b/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
@@ -1605,6 +1605,7 @@ struct PhysicsServerCommandProcessorInternalData
 	CommandLogPlayback* m_logPlayback;
 	int m_logPlaybackUid;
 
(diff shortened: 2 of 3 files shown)


erwincoumans:
Thanks for the PR. Why do we need to store yet another variable, we could just add a method 'getFixedTimeStep' and compute it from m_physicsDeltaTime  and m_numSubSteps?

Also, we need to apply a similar fix for the contact points:

~~~~~~~~~~~~~~~~~~~
MyContactResultCallback cb(m_data->m_cachedContactPoints);

cb.m_bodyUniqueIdA = bodyUniqueIdA;
cb.m_bodyUniqueIdB = bodyUniqueIdB;
cb.m_deltaTime = m_data->m_physicsDeltaTime;

pt.m_normalForce = srcPt.getAppliedImpulse() / m_deltaTime;
pt.m_linearFrictionForce1 = srcPt.m_appliedImpulseLateral1 / m_deltaTime;
pt.m_linearFrictionForce2 = srcPt.m_appliedImpulseLateral2 / m_deltaTime;
~~~~~~~~~~~~~~~~~~~

erwincoumans:
Applied a fixed here, please check:
https://github.com/bulletphysics/bullet3/pull/2809

fuchuyuan:
> Thanks for the PR. Why do we need to store yet another variable, we could just add a method 'getFixedTimeSubStep' and compute it from m_physicsDeltaTime and m_numSubSteps?
> 
> Also, we need to apply a similar fix for the contact points:
> 
> ```
> MyContactResultCallback cb(m_data->m_cachedContactPoints);
> 
> cb.m_bodyUniqueIdA = bodyUniqueIdA;
> cb.m_bodyUniqueIdB = bodyUniqueIdB;
> cb.m_deltaTime = m_data->m_physicsDeltaTime;
> 
> pt.m_normalForce = srcPt.getAppliedImpulse() / m_deltaTime;
> pt.m_linearFrictionForce1 = srcPt.m_appliedImpulseLateral1 / m_deltaTime;
> pt.m_linearFrictionForce2 = srcPt.m_appliedImpulseLateral2 / m_deltaTime;
> ```

m_deltaTime is a member field of MyContactResultCallback so m_deltaTime should be used. But since MyContactResultCallback is only used here once either should be fine.

fuchuyuan:
I don't have a strong preference wether we should add a method 'getFixedTimeSubStep' to avoid saving an extra field, or avoid recomputing. Either is fine with me.

erwincoumans:
Sounds good, it is already merged in a separate PR.

erwincoumans:
the PR missed various changes, here is a follow up PR:
https://github.com/bulletphysics/bullet3/pull/2815

PR #2800, opened 2020-05-12
Stability improvements for deformable.

xhan0619:
1) Added Baumgarte stabilization when distance is positive
2) Added split impulse for deformable vs. multibody.
3) Tune parameters for deformable pybullet examples.


src/BulletDynamics/Featherstone/btMultiBody.h
```diff
@@ -397,6 +402,26 @@ btMultiBody
...
+    void applyDeltaSplitVeeMultiDof(const btScalar *delta_vee, btScalar multiplier)
+    {
+        for (int dof = 0; dof < 6 + getNumDofs(); ++dof)
+        {
+            m_splitV[dof] += delta_vee[dof] * multiplier;
+        }
+    }
+    void addSplitV()
+    {
+        applyDeltaVeeMultiDof(&m_splitV[0], 1);
```
erwincoumans, line 414: did you want to call applyDeltaSplitVeeMultiDof?
If calling  applyDeltaVeeMultiDof, it may change default behavior of the simulation.

src/BulletDynamics/Featherstone/btMultiBodyDynamicsWorld.cpp
```diff
@@ -579,6 +579,7 @@ void btMultiBodyDynamicsWorld::integrateMultiBodyTransforms(btScalar timeStep)
 		for (int b = 0; b < m_multiBodies.size(); b++)
 		{
 			btMultiBody* bod = m_multiBodies[b];
+            bod->addSplitV();
```
erwincoumans, line 582: can this be moved after the sleeping check?

src/BulletDynamics/Featherstone/btMultiBodyDynamicsWorld.cpp
```diff
@@ -615,6 +616,7 @@ void btMultiBodyDynamicsWorld::integrateMultiBodyTransforms(btScalar timeStep)
 			{
 				bod->clearVelocities();
 			}
+            bod->substractSplitV();
```
erwincoumans, line 619: same, move within the sleep check, if possible.

src/BulletDynamics/Featherstone/btMultiBodyDynamicsWorld.cpp
xhan0619: Done.

src/BulletDynamics/Featherstone/btMultiBodyDynamicsWorld.cpp
xhan0619: Done.

src/BulletSoftBody/btSoftBody.cpp
```diff
@@ -4089,7 +4089,7 @@ void btSoftBody::defaultCollisionHandler(const btCollisionObjectWrapper* pcoWrap
             {
                 const btTransform wtr = pcoWrap->getWorldTransform();
                 const btScalar timemargin = 0;
-                const btScalar basemargin = getCollisionShape()->getMargin();
+                const btScalar basemargin = 3.0*getCollisionShape()->getMargin();
                 btVector3 mins;
```
erwincoumans, line 4093: there are more magic constants, like 2, 3 and 4. Why are those needed? If they are important, it should be exposed to the user?

xhan0619: Those are more experimental. I have removed them from this PR.

PR #2801, opened 2020-05-13
btCollisionObject.h: add getters for m_objectsWithoutCollisionCheck

stephengold:
`btCollisionObject` has an API to add/remove collision objects from its `m_objectsWithoutCollisionCheck` list. However, without extending the class there's no way to determine how many objects are in the list or to enumerate them.

This pull request fills this gap by adding 2 methods to `btCollisionObject`.


erwincoumans:
thanks, looks good to me.

stephengold:
Thank you. I look forward to v2.90 !

PR #2812, opened 2020-05-18
btAlignedAllocator: filename args should be const char*

stephengold:
Some compilers (such as Clang) object to passing a string literal such as `__FILE__` to a function that takes a `char *` argument.


erwincoumans:
Thanks, LGTM.

PR #2816, opened 2020-05-20
btOverlappingPairCache: add a getInternalGhostPairCallback() method

stephengold:
This is useful because `m_ghostPairCallback` points to caller-allocated storage. To avoid a memory leak, the caller needs to deallocate it before freeing the cache. Having a getter saves the caller from needing to keep track of the pointer.

### DIFF
--- a/src/BulletCollision/BroadphaseCollision/btOverlappingPairCache.h
+++ b/src/BulletCollision/BroadphaseCollision/btOverlappingPairCache.h
@@ -77,7 +77,7 @@ public:
 
 	virtual bool hasDeferredRemoval() = 0;
 
-        virtual btOverlappingPairCallback* getInternalGhostPairCallback() = 0;
+	virtual btOverlappingPairCallback* getInternalGhostPairCallback() = 0;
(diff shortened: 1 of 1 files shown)


src/BulletCollision/BroadphaseCollision/btOverlappingPairCache.h
```diff
@@ -77,6 +77,7 @@ class btOverlappingPairCache : public btOverlappingPairCallback
 
 	virtual bool hasDeferredRemoval() = 0;
 
+        virtual btOverlappingPairCallback* getInternalGhostPairCallback() = 0;
```
erwincoumans, line 80: can you please use tabs not spaces?

stephengold: Sure. Coming right up.

stephengold:
What needs to happen to get this PR integrated?

stephengold:
Withdrawn by the author after 6 months of inactivity.

PR #2826, opened 2020-05-26
Deformable gravity factor

xhan0619:
There are occasions where a user wants to set scale the "gravity" that deformable object experience to be different from the actual gravity. These changes enables that.


erwincoumans:

PR #2832, opened 2020-05-29
Set numNonContactInnerIterations from Pybullet

xhan0619:


erwincoumans:

PR #2840, opened 2020-06-02
KukaCamGymEnv

mahyaret:
Hi,
I have noticed that the `render()` function in KukaCamGymEnv is trying to read a nonexistent object `self._racecar` unique id. I have changed this function to call `getExtendedObservation()`

Thanks,
Mahyar


mahyaret:
?

erwincoumans:
It is an old, not much used gym env, still good to fix it. Thanks!

PR #2847, opened 2020-06-09
Decrease the dof count when decreasing the number of links.

RauchMatthias:
I don`t know what m_posVarCnt is being used for, but it looks like it should be decreased as well, so I  did.

Also removed a redeclaration of m_dofCount in btMultiBody.cpp.


RauchMatthias:
Sorry, I just made the Box2d debug draw changes yesterday and didn`t know github would automatically combine them with an old pull request. I will make another pull request for them.
Thanks for making the engine!

PR #2855, opened 2020-06-13
Fix parameter check in bFile::safeSwapPtr

asmaloney:
If either the source or destination pointers are nullptr, we shouldn't continue otherwise we dereference nullptr or memcpy with nullptr which is undefined.

(Also move the check to return as soon as possible.)


erwincoumans:
Thanks for the PR. This this actually ever happen? If so, we need to fix the source of a null pointer there.

asmaloney:
I have not seen it happen in practice. I noticed it (and the issues in my other PRs) because I'd run some static analysis on other code and included the bullet code by accident. :-)

erwincoumans:
Thanks, LGTM

PR #2858, opened 2020-06-15
Fix potential undefined softbody member in PhysicsServerCommandProcessor

CGJiaxiLiu:
Fix two compile error due to undefined softbody member variable.


erwincoumans:
Thanks, will check it. (close/re-open to trigger CI)

erwincoumans:
Thanks, LGTM!

PR #2872, opened 2020-06-20
Corotated Linear Elasticity

xhan0619:
Add the cororated linear elasticity model, and apply the corotated idea to the damping model. Improved contact stability using an implicit formulation.
See demos: Volumetric Deformable Objects, Extreme Deformation and Colliding Test.


erwincoumans:
Thanks a lot for the PR.
I notice if you pick up the cloth very quickly in the Deformable Self Collision example, it has some strange artifacts.
Please check attached video of old and new behavior.
(I didn't double-check if this PR is the reason yet)
[compare.zip](https://github.com/bulletphysics/bullet3/files/4828304/compare.zip)

erwincoumans:
Thanks a lot for all the contributions! It looks like the cloth margin/gap is working fine now, in the Deformable Self Collision demo.

PR #2873, opened 2020-06-21
Add compiler definitions for double precision to CMake config

### LINKED ISSUE #2582: BT_USE_DOUBLE_PRECISION is not added to BulletConfig.cmake.in
It looks like this is exported to bullet.pc.in but is not done for BulletConfig.cmake.in. Should BULLET_DEFINITIONS in BulletConfig.cmake.in be changed to the argument used in bullet.pc.in?

rickertm:
This patch adds the `BT_USE_DOUBLE_PRECISION` compiler definition to the `BULLET_DEFINITIONS` variable used in `BulletConfig.cmake` when using the CMake option `USE_DOUBLE_PRECISION`.

Fixes #2582


erwincoumans:

PR #2874, opened 2020-06-21
Enable relocation of Bullet installation and use absolute paths in CMake

rickertm:
This patch updates the `BULLET_ROOT_DIR` variable in `BulletConfig.cmake` to use [`CMAKE_CURRENT_LIST_DIR`](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_LIST_DIR.html) instead of the fixed value of [`CMAKE_INSTALL_PREFIX`](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html) used during installation. This enables an easy relocation of the Bullet installation to a different directory, as the other value otherwise has to be updated manually.

It also adds `${BULLET_ROOT_DIR}` as prefix to the currently relative path variables, as absolute paths are required when using these values in a CMake build script. Otherwise, these paths are seen as relative to the user's source directory.


felixf4xu (approved):
this should be merged.

Issue #2868, opened 2020-06-18
Further enhancements to DeepMimic Gym env


ManifoldFR:
Hi,

Continuing from #2821 and my PR #2822, I'd like to make more updates to the DeepMimic Gym environment. I'm thinking of:
- [x] reproducing the observation and action offset/rescaling trick from the initial DeepMimic code (using the `build_state_offset()/build_state_scale()` methods from the [PyBulletDeepMimic class](https://github.com/bulletphysics/bullet3/blob/5233b72160df8e8915320b03793d85dd0c6792ba/examples/pybullet/gym/pybullet_envs/deep_mimic/env/pybullet_deep_mimic_env.py#L132)), which should help with training
- [x] more fixes to the update/policy query frequency
- [x] having the camera of the `env.render()` method track the humanoid character (the current camera angle is a bit awkward)
- [x] for *evaluation*, set a flag to the Gym env constructor that forces the movement to start at the beginning after every reset
- [ ] implement the missing [center-of-mass reward](https://github.com/bulletphysics/bullet3/blob/63e1d7d84055532e5f29484fa34c1af604f595f1/examples/pybullet/gym/pybullet_envs/deep_mimic/env/humanoid_stable_pd.py#L913), which is the squared difference of COM velocity

I'm working on this within my own codebase and will make upstream contributions.

ManifoldFR:
I started addressing the missing COM reward, mentioned in issue #2871 in ManifoldFR/bullet3@04350bc.

Issue #2871, opened 2020-06-19
Differences in DeepMimic between bullet and pybullet versions


tfederico:
Hi,

which are the current differences between the DeepMimic implementation in pybullet and the original one?

erwincoumans:
They should be very similar, I tried using the same training code from Jason Peng, some differences include:
1) PyBullet uses a larger time step
2) PyBullet uses a URDF definition of the character, but it should be identical
3) PyBullet is easier to install (pip3 install pybullet)

There could be small other unintended differences. We could use the original policies from DeepMimic directly in PyBullet, but a while back some of them started to perform poorly (didn't have time to track down why) so I retrained some of the policies.

Closing for now, just reply if you have other questions.
Also, note that Jason Peng did an internship with us and we released another imitation environment based on PyBullet for our quadruped robots. See
https://github.com/google-research/motion_imitation

tfederico:
Looks amazing, thanks! I am currently interested in human-like characters, but I will definitively take a look at the paper :)

tfederico:
out of curiosity, is there any reason why you chose VideoPose3D instead of using HMR or HMMR like in the Skills from Video paper?

erwincoumans:
@tfederico There is no video involved in PyBullet deep_mimic.

tfederico:
sorry, I must have confused with some other repo I was looking at

PR #2883, opened 2020-06-25
DeepMimic Gym enhancements

ManifoldFR:
This PR is a sequel to issue #2821 and PR #2822 where first fixes were made to the DeepMimic Gym environment (mostly the policy querying frequency, which should be 30Hz as in the paper). These fixes are finalized and further enhancements were made as said in issue #2868:
* the render camera tracks the character
* the actions and states are rescaled as Jason Peng's original code does (this can be disabled)
* I added a test mode flag that makes the `reset()` method reset the reference clip to the beginning instead of random, which is better for evaluation/visualization

I was able to train the backflip policy using my own PyTorch codebase (which does not rely on MPI): see [this video](https://youtu.be/K-fYcoDdP7M).
The sample efficiency was not as good as the initial TensorFlow code I think but it could just be me choosing bad hyperparameters.
As mentioned in issue #2871 the character's root position tends to drift from the reference motion.

Resolves issue #2868.

### DIFF
--- a/examples/pybullet/gym/pybullet_envs/deep_mimic/env/pybullet_deep_mimic_env.py
+++ b/examples/pybullet/gym/pybullet_envs/deep_mimic/env/pybullet_deep_mimic_env.py
@@ -10,16 +10,26 @@ import pybullet_data
 import pybullet as p1
 import random
 
+from enum import Enum
+
+class InitializationStrategy(Enum):
+  """Set how the environment is initialized."""
+  START = 0
+  RANDOM = 1  # random state initialization (RSI)
+
 
 class PyBulletDeepMimicEnv(Env):
 
-  def __init__(self, arg_parser=None, enable_draw=False, pybullet_client=None):
+  def __init__(self, arg_parser=None, enable_draw=False, pybullet_client=None,
+               init_strategy=InitializationStrategy.RANDOM):
     super().__init__(arg_parser, enable_draw)
     self._num_agents = 1
     self._pybullet_client = pybullet_client
     self._isInitialized = False
     self._useStablePD = True
     self._arg_parser = arg_parser
+    self._init_strategy = init_strategy
+    print("Initialization strategy: {:s}".format(init_strategy))
     self.reset()
 
   def reset(self):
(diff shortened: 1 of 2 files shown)


erwincoumans:
Great, thanks for finishing the DeepMimic Gym environment, glad to see you can train it with another implementation.

erwincoumans:
Do you have a link to your PyTorch training codebase?

Did you try training the env using stable_baselines?

Recently, Jason Peng implemented the motion_imitation env also trained using stable_baselines.
The curriculum improves sample efficiency, we don't enable this in our DeepMimic environment, I think:
https://github.com/google-research/motion_imitation/blob/246b2e786ccb2033cf0f4eff6fe6c62dbfb51eee/motion_imitation/envs/env_wrappers/imitation_wrapper_env.py#L26

ManifoldFR:
My PyTorch codebase is currently private because it's part of an ongoing IL project with my institution, and I haven't tried training using another RL codebase yet.

I don't remember the concept of curriculum in the original animal imitation paper, though, but I'll take a look thanks !
Right now I'm training the cartwheel and crawling motions, so if I make further changes to the Gym env and add more options I will make another contribution.

erwincoumans:
Thanks. Do you also make sure the original code still trains and 'enjoys' ? How does the training speed compare with your code?

python3 mpi_run.py --arg_file train_humanoid3d_walk_args.txt --num_workers 34
python3 -m pybullet_envs.deep_mimic.testrl --arg_file run_humanoid3d_walk_args.txt

ManifoldFR:
I just checked and running the policies still works well. I didn't make any breaking changes to the sim environment file pybullet_deep_mimic_env.py and Jason Peng's RL code doesn't use the other file I changed.
I did not check if the training behaviour changed though.
If I remember another difference with the initial framework was the COM reward in `humanoid_stable_pd.HumanoidStablePD` had a weight of 0, right?

tfederico:
I would also be interested in a PyTorch version. @ManifoldFR could you share an example of how to use it instead of MPI?

ManifoldFR:
> I would also be interested in a PyTorch version. @ManifoldFR could you share an example of how to use it instead of MPI?

I think the closest public thing to what I did would be [stable-baselines3](/DLR-RM/stable-baselines3/). Since the env uses the Gym interface I can just instantiate a vectorized env that runs in subprocesses and I don't need MPI anymore. stable-baselines has its own implementation of vectorized envs, but the current version of OpenAI Gym also has these.

tfederico:
thank you, I will take a look at the link you suggested!

PR #2886, opened 2020-06-26
delete '- ' prefix on getList()

bionicles:
if we delete this prefix then we can randomly choose environments without having to slice strings


erwincoumans:
It may break existing behavior, rather add a new method (getList2()).

PR #2888, opened 2020-06-27
Ensure naming consistency in gym registry

AiRuiChen:
Hi,

I am working on adding bullet support for [garage](https://github.com/rlworkgroup/garage). I found that [`pybullet_envs.getList()`](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/__init__.py#L230) does not return a complete list of bullet gym environments because not all environments in this package contain the substring `Bullet`. 

I added duplicate environment entries in the gym registry to ensure all bullet environments can be found via `pybullet_envs.getList()`. This ensures backwards compatibility. 

Please kindly let me know if you have any questions;)


erwincoumans:
Thanks I wasn't aware of garage, looks interesting (we use stable-baselines mostly).

I prefer to maintain the names, not renaming things everywhere. 
So probably best to add the names explicitly inside the pybullet_envs.getList()

AiRuiChen:
> Thanks I wasn't aware of garage, looks interesting (we use stable-baselines mostly).
> 
> I prefer to maintain the names, not renaming things everywhere (don't break other's code whenever possible).
> 
> So probably best to add the names explicitly inside the pybullet_envs.getList(), perhaps as duplicates.

Sure! Please see the updated `pybullet_envs.getList()`

erwincoumans:
Looks good to me, thanks for the help!

erwincoumans:
Will be in upcoming 2.8.4 release (just missed the 2.8.3 one today).

PR #2894, opened 2020-06-30
Add --dynamic-runtime option to compile with /MD[d] in MSVC for compat

sinbad:
I wanted to try Bullet in an environment where I can't control the MSVC CRT in use, so I needed to use /MD[d] instead of /MT[d] as the CRT option. This PR adds --dynamic-runtime option to the premake build to that effect.

I don't know if you want this as an option since by default the example browser won't build with it (not a problem for me), or at the least you might not want the .bat by default. But it works from a library POV and is useful for people in my position (I saw a thread or 2 in the forum about it) so I thought I'd send a PR.

Cheers Erwin! You probably don't remember but we met at Siggraph many MANY years ago when I was running Ogre3d and we shared a BOF room once. :)

### DIFF
--- a/build3/premake4.lua
+++ b/build3/premake4.lua
@@ -370,12 +370,24 @@ end
 		defines {"BT_CLAMP_VELOCITY_TO=9999"}
 	end
 
+	newoption
+	{
+		trigger = "dynamic-runtime",
+		description = "Enable dynamic DLL CRT runtime"
+	}
 	configurations {"Release", "Debug"}
 	configuration "Release"
-		flags { "Optimize", "EnableSSE2","StaticRuntime", "NoMinimalRebuild", "FloatFast"}
+		flags { "Optimize", "EnableSSE2", "NoMinimalRebuild", "FloatFast"}
+		if not _OPTIONS["dynamic-runtime"] then
+			flags { "StaticRuntime" } 
+		end
 	configuration "Debug"
 		defines {"_DEBUG=1"}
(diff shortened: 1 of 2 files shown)


erwincoumans:
Thanks Steve! Of course I remember those fun SIGGRAPH times! Hope you are doing well!
Yes, let's add the option. I think it is also available in the cmake build system (even though I prefer static linking usually).
Let me see about the new bat file, if it doesn't work for the example browser, that will be confusing, perhaps we can make it work?
Of, and if a project involves Python, you may give pybullet a try, it installs even simpler (pip install pybullet)

sinbad:
:) Ok actually the example browser *does* build, you just have to delete the pdb first if you'd previously built with /MT because of the fastlink option. So that's all fine. 

I prefer static linking too - even with this option Bullet itself is still a static library, it's just the CRT used is the DLL version. So from the perspective of user apps it's not really any different, except compatibility with the CRT they're already using. That's why I thought I'd call it `--dynamic-runtime` instead of the shorter `--dynamic` which might have implied building Bullet as DLLs.

pybullet sounds great but I'm linking this into Unreal - I need deterministic physics for something without making source changes to UE, so I want to run my own little sim for it. I like that Bullet is deterministic if used carefully and has a rolling friction model. Unreal requires /MD[d] though, hence this PR.

erwincoumans:
Sounds good. I started working on another physics engine. It can be a differentiable physics engine, but also a fully deterministic physics engine cross platform, by using the fixed point scalar. It is still limited of features, but it may be of interest. See 

https://github.com/google-research/tiny-differentiable-simulator

PR #2900, opened 2020-07-01
Fix processInitPoseCommand: must update Aabbs after resetting body pose

### LINKED ISSUE #2881: strange result of rayTest in pybullet
I am trying `rayTest` from pybullet, and having some strange results. It would be really nice if someone points out my mistakes, or takes this as a bug-report if it's a bug. The following python script performs a rayTest against a small_sphere. 

```python3
import pybullet as p
import pybullet_data
import numpy as np

CLIENT = p.connect(p.GUI)
p.configureDebugVisualizer(p.COV_ENABLE_GUI, 0)
p.configureDebugVisualizer(p.COV_ENABLE_SHADOWS, 0)
p.setAdditionalSearchPath(pybullet_data.getDataPath()) #used by loadURDF

def test_smallsphere(pos_): 
    pos = np.array(pos_)
    ss = p.loadURDF("sphere_small.urdf")
    p.resetBasePositionAndOrientation(ss, pos, [0, 0, 0, 1])
    margin = np.array([0.3, 0.0, 0.0])
    p_start = pos - margin
    p_end = pos + margin
    p.addUserDebugLine(p_start, p_end, [1, 0, 0])
    result = p.rayTest(p_start, p_end)
    return result

result = test_smallsphere([0.0, 0, 0.000])
print(result)
```
This script works and shows a reasonable result: 
```
((0, -1, 0.45000000000000007, (-0.02999999999999997, 0.0, 0.0), (-1.0, 0.0, 0.0)),)
```
Something strange is when I change the position of the sphere, for example when I change the argument to `result = test_smallsphere([0.0, 0, 0.1])`, no ray-collision is detected, In other words, I got result: 
```
((-1, -1, 1.0, (0.0, 0.0, 0.0), (0.0, 0.0, 0.0)),)
```
This result is strange. I don't change relative geometrical relation between the ray and the small sphere, so it suppose to collide. Just for double check, the above script visualizes the line between `p_start` and `p_end` and I confirmed that it actually penetrate the sphere. 
bullet

I use ubuntu 16.04 and the pybullet library is installed via pip(python 3.7.7). When I import pybullet inside python3, `pybullet build time: Jun 19 2020 04:02:40` appears. This version seems newer and I guess the topics discussed in https://github.com/bulletphysics/bullet3/issues/1947 are probably unrelated.

HiroIshida:
Closes #2881 
I found that the reason why this [probelm](https://github.com/bulletphysics/bullet3/issues/2881) occurs. This is becaues after resetting the pose of the body Aabbs are not updated. I tested the error case (collision is not detected thought it supposed to) mentioned in #2881, and checked that collision is detected after this commit. 

I found this bug when compare the behavior of inside `btDbvt::rayTestInternal` method when I change parameter of code in #2881 as  follows:
```python
result = test_smallsphere([0.0, 0, 0.000]) # case A
result = test_smallsphere([0.0, 0, 0.1]) # case B 
``` 
I noticed that when case B, `btRayAabb2`  [here](https://github.com/bulletphysics/bullet3/blob/63e1d7d84055532e5f29484fa34c1af604f595f1/src/BulletCollision/BroadphaseCollision/btDbvt.h#L1252
) returns false although it should not so. I further found that `node->volume.Mins()` and `node->volume.Maxs()` in [here](https://github.com/bulletphysics/bullet3/blob/63e1d7d84055532e5f29484fa34c1af604f595f1/src/BulletCollision/BroadphaseCollision/btDbvt.h#L1248) do not change after `    p.resetBasePositionAndOrientation(ss, pos, [0, 0, 0, 1])` is called. Then, I fixed this bug.

### DIFF
--- a/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
+++ b/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
@@ -10197,6 +10197,7 @@ bool PhysicsServerCommandProcessor::processInitPoseCommand(const struct SharedMe
 		scratch_m.resize(nLinks + 1);
 
 		mb->updateCollisionObjectWorldTransforms(scratch_q, scratch_m);
+		m_data->m_dynamicsWorld->updateAabbs();
 	}
 
 	if (body && body->m_rigidBody)


erwincoumans:
Thanks. It is best to just update a single AABB, using the 
~~~~~~~~~
world->updateSingleAabb(colObj); 
~~~~~~~~~

cadop:
@erwincoumans  is there any reason not to pull this, or is it being solved somewhere else? I ran into the same issue with pybullet today.

HiroIshida:
@cadop Probably, my not responding is the reason why this PR is not pulled. 
@erwincoumans 
I am sorry that I didn't respond to
``` 
It is best to just update a single AABB, using the
world->updateSingleAabb(colObj); 
```
6 month ago, I attempted to interpret this comment for couple of hours, but I couldn't. 
it would be nice you make a new PR to fix this problem

cadop:
so I wasn't sure if this is intended behavior for some reason, or actually a bug.  Since I was using pybullet:  for myself, I would rather spend the time to expose the C function to updateAABB in the python interface as a method to be called by users if they manually move an object, but this is also an interface design decision.

erwincoumans:
Here is a suggested fix that automatically updates the AABB:
https://github.com/bulletphysics/bullet3/pull/3311/commits/0196edd4d2e0cde2de5c8138d5f69df2cea9fce4

Exposing a 'performDiscreteCollisionDetection' would be useful too (it would update all aabb's as well as computing contact points, without running a full stepSimulation with forward dynamics, and collision response etc).

PR #2904, opened 2020-07-01
Add USE_MSVC_RELEASE_RUNTIME_ALWAYS to use Release CRT even in Debug

sinbad:
This is another result of my noodling with UE4 & Bullet. In Debug mode, UE is built to use the Release CRT - I guess this makes sense for performance and you rarely need to debug into the CRT, but it means I couldn't use the debug build of Bullet with it.

I tried to make this work with premake but it appears the `runtime` option is only supported in premake 5, and I don't have the familiarity or the time to update the premake scripts to v5. So I switched to CMake, where I found the /MD options you mentioned in my other PR, thanks, and added the option there. I can now debug UE4 -> Bullet quite happily using this option.


erwincoumans:
re-trigger by close/open

erwincoumans:
Thanks for the contribution!

PR #2912, opened 2020-07-06
DeepMimic: add COM reward

ManifoldFR:
This addresses part of #2868, the missing center-of-mass reward (also noted #2871).

This is done by implementing computation of CoM position and velocity, similar to [`cSimCharacter::CalcCOM`](https://github.com/xbpeng/DeepMimic/blob/fbd77f66f7e0ee9d7cd2c863636f296f3af45e04/DeepMimicCore/sim/SimCharacter.cpp#L398) in the original codebase.

Using the CoM reward is left as an optional flag in the `HumanoidStablePD` class constructor.

This implements the CoM reward using the squared norm of the CoM velocity error, just like in Jason Peng's code (and not the paper where the CoM position error is used).


ManifoldFR:
I launched a training job with the original TF code on my end to see what kind of policy is learned. I also checked with the Gym environment, and I get similar policy to before (except the sim character drifts from the reference character a bit later).

ManifoldFR:
I checked training a policy with Jason Peng's TF code. It still works as before, with consistent rewards of about 600. Should I upload the TF checkpoint?

erwincoumans:
Thanks for working on this and on the contribution!
Yes, can you upload some checkpoint? Did you try out the wallk and backflip motions?

ManifoldFR:
I only trained the walking motion. Do I just overwrite the `humanoid3d_walk` checkpoint in `pybullet_data/data/policies/humanoid3d` and commit ?

erwincoumans:
I'm mainly concerned if the backflip can still train with the modification. If not, we can disable the COM reward by default (or using a flag).

tfederico:
I tried to run the motion file walk using the checkpoint `humanoid3d_walk_COMenabled.ckpt`, but the character walks in an odd way. I re-downloaded the checkpoints from this repository, just to be sure that I had the most up to date files.

You can see the result in this [https://drive.google.com/file/d/1GkRoqJ3umrxVqFb1Upzh00tvx42xPxyz/view?usp=sharing](video).

Did you train it using `humanoid3d_walk.txt`?

PR #2916, opened 2020-07-07
Re-enable sphere-box collision algorithm.

logzero:
It was fixed by https://github.com/bulletphysics/bullet3/commit/394bb103f05cfa764c82cfb894c020d3826ae3a3#diff-e36529e40845d95a860d9d36024464c9 but never re-enabled.


erwincoumans:
Thanks, but I'll leave it disabled, since it changes behavior in subtle ways and leave it up to the user to enable and test corner cases.

PR #2931, opened 2020-07-15
Fix minitaur_logging proto and import

AiRuiChen:
Avoid proto conflicts by changing the package of `minitaur/envs/timestamp.proto` from `google.protobuf` to `robotics.messages`. Otherwise there is a duplicate proto defined in `google.protobuf` which causes an error.

Also updated `__init__.py` so gym registry can find missing class entry points. Some environments are registered but not imported. e.g. [`MinitaurExtendedEnv-v0`](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/__init__.py#L57)


erwincoumans:
Thanks for the contribution. I'm less familiar with the proto features. Where is robotics.messages timestamp defined?

AiRuiChen:
> Thanks for the contribution. I'm less familiar with the proto features. Where is robotics.messages timestamp defined?

The change of package name can be found [here](https://github.com/bulletphysics/bullet3/pull/2931/files#diff-86b12063bc313c87ff532bf4917cd1b0R30). 

`timestamp.proto` is almost identical as `google.protobuf`'s [timestamp.proto](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto). So I guess the author copied from there directly, but forgot to change the package name?

I chose `robotics.messages` because another proto file in this directory, [vector.proto](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/minitaur/envs/vector.proto#L3), is created under this package.

erwincoumans:
Thanks, but I still don't find it. Where is that robotics messages proto file defining timestamp?

AiRuiChen:
> Thanks, but I still don't find it. Where is that robotics messages proto file defining timestamp?

`timestamp.proto` is [here](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/minitaur/envs/timestamp.proto).

If that's not what you mean. Are you looking for something like `robotics.messages.proto`?

I can't find it in `bullet3/` either. I generated the python proto file via `protoc`, and apparently proto package in python is used more for the purpose of avoiding naming conflicts (as far as I can tell from this [answer](https://stackoverflow.com/a/60483464)). That said, I don't know if a "package" proto needs to be present.

Another thought:
Can robotics.messages live inside google's codebase (`google3`)? I found the same naming declaration [here](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/minitaur/agents/scripts/train_ppo_test.py#L25).

erwincoumans:
Thanks, let's rename the timestamp proto then.

There are some large changes in another files. For example, why was this removed? It allows locally running (without install):
~~~~~~~~~~~~~~~~~~
import os, inspect
currentdir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
parentdir = os.path.dirname(os.path.dirname(currentdir))
os.sys.path.insert(0, parentdir)
~~~~~~~~~~~~~~~~~~

AiRuiChen:
I generated the _pb2.py files with the `protoc` command. (e.g. `protoc -I=$SRC_DIR --python_out=$DST_DIR $SRC_DIR/timestep.proto`). And I just replaced the file contents. I am not sure why the generated files differ so much. Btw I have protobuf 3.12.3. 

Anything wrong with the way I create these files?  I am happy to re-generate these once we figure this out.

erwincoumans:
ok, thanks! LGTM.

PR #2933, opened 2020-07-15
Fix pybullet gym envs

AiRuiChen:
Some small fixes to make a few environments work with `gym.make()`. 

1. `KukaDiverseObjectGymEnv` doesn't have `observation_space` defined. This property is required since it inherits from `gym.Env`. 
2. `MinitaurStandGymEnv._transform_action_to_motor_command()` should return a list instead of a `map` object. Otherwise it won't work with [`minitaur.ApplyAction()`](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/minitaur/envs/minitaur.py#L577) in `minitaur.Step()`.


erwincoumans:
Thanks for the contribution!

PR #2937, opened 2020-07-16
prev keyword removed

addy1997:
The code executes without the prev keyword. I removed it and tried to run the code, it ran perfectly.


erwincoumans:
Thanks, LGTM.

PR #2938, opened 2020-07-16
applied const modifiers to some RigidBody getters

MomoDeve:
getPushVelocity and GetTurnVelocity should be marked as const, because they return vector by value and do not change object state


erwincoumans:
Thanks, LGTM.

Issue #2935, opened 2020-07-15
Possible memory leak in saveStateToMemory


aravindsiv:
Hi,

I am trying to use the C++ API to simulate the MIT racecar. My program was using up a lot of memory, and when I ran valgrind to debug, I found a large number of errors that looked like below:

```
==60092==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==60092==    by 0x871182: btAlignedAllocDefault(unsigned long, int) (in /path/to/my/executable)
==60092==    by 0x5C0534: btDefaultSerializer::finishSerialization() (in /path/to/my/executable)
==60092==    by 0x5B44D4: PhysicsServerCommandProcessor::processSaveStateCommand(SharedMemoryCommand const&, SharedMemoryStatus&, char*, int) (in /path/to/my/executable)
==60092==    by 0x5B5065: PhysicsServerCommandProcessor::processCommand(SharedMemoryCommand const&, SharedMemoryStatus&, char*, int) (in /path/to/my/executable)
==60092==    by 0x6850F0: PhysicsDirect::submitClientCommand(SharedMemoryCommand const&) (in /path/to/my/executable)
==60092==    by 0x5F2933: b3SubmitClientCommand (in /path/to/my/executable)
==60092==    by 0x5F29CF: b3SubmitClientCommandAndWaitStatus (in /path/to/my/executable)
==60092==    by 0x580E4F: b3RobotSimulatorClientAPI_NoDirect::saveStateToMemory() (in /path/to/my/executable)
```

This happens even if I am not doing anything overly complicated in my executable. I am just loading the URDF of the racecar and the plane, switching on gravity, stepping the simulation, and then making a call to `saveStateToMemory()`. The problem is when I perform many simulations and keep calling `saveStateToMemory()`, even if I remove the states from memory, the memory usage of the program still blows up.

aravindsiv:
I tried the following edits to `examples/SharedMemory/PhysicsServerCommandProcessor.cpp`  (similar to the fix merged in #2397 ):

Inside `PhysicsServerCommandProcessor::processSaveStateCommand( ... )`, I added the following lines before the return statement:

```
delete ser; 
delete bulletFile;
```

Unfortunately, the second line doesn't let me save states anymore, and only leaving the first line, i.e. deleting the `btDefaultSerializer` object still causes the memory leak described earlier.

erwincoumans:
[...]

PR #2940, opened 2020-07-17
add removeState in C++ API, see also Issue https://github.com/bulletp…

erwincoumans:
…hysics/bullet3/issues/2935

### DIFF
--- a/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
+++ b/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
@@ -2038,6 +2038,12 @@ PhysicsServerCommandProcessor::~PhysicsServerCommandProcessor()
 	if (m_data->m_threadPool)
 		delete m_data->m_threadPool;
 
+	for (int i = 0; i < m_data->m_savedStates.size(); i++)
+	{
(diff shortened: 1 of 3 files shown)


aravindsiv:
Hi, I've added a follow-up comment to #2935 . But I am reproducing it here:

Even after I call `removeState()`, the memory al is not cleaned up until the end of `PhysicsServerCommandProcessor`'s lifetime. Is there a way to clean up the `savedState`'s `m_bulletFile` and `m_serializer` objects when we call `removeState()`?

erwincoumans:
removeState IS deleting that memory, there is no leak. No need to wait for resetSimulation. Dis you apply my CL?

aravindsiv:
I did, but I am still seeing no reduction in memory usage after I call `removeState` on all the stateIds I have saved so far. The stateId counter gets reset, but the memory doesn't appear to be deallocated.

erwincoumans:
Me too, and there were no leaks. Please check https://github.com/bulletphysics/bullet3/issues/2935 and let me know if you see leaks/growing memory in that example. Let's discuss there, assuming you applied this pull request and no other modifications to the Bullet source code.

Issue #2922, opened 2020-07-09
Chain of bodies that looks like a loop, is this possible?


DarioUrbina:
I want to join a child link (i.e. Link #10), to the base link (i.e. link #-1). This is generating discrepancies when trying to run the simulation.  

Since I cannot accomplish this with declared joints in my .urdf files, I have opted to restrict the motion of a body with a artificial joint (I am relying on collisions to do so). This is apparently a good solution. But when working with very complex bodies the collisions are not respected and the models behaves like in a chaotic way. 

Here you can see my code: Pendulum_exercise/Inverted_pendulum_tendon_actuation.py

erwincoumans:
You can create a fixed constraint (or point-to-point constraint etc) to create loops.

DarioUrbina:
Thank you for the reply, 

In the zip file you can find the files, the first to the Python code, the second to the URDF code: 

[Inverted_pendulum_tendon_actuation.zip](https://github.com/bulletphysics/bullet3/files/4900085/Inverted_pendulum_tendon_actuation.zip)

PR #2950, opened 2020-07-22
Inverted Pendulum Actuated by Tendons

DarioUrbina:
This is an example of PID control of an inverted pendulum actuated by strings. The pendulum axis is located on a cart that can only move pulled by two strings attached to it.

To create the bodies of this simulation, Issue #2922 was solved. You will see how I use a chain of bodies connected in a "loop" fashion.

When you run inverted_pendulum_tendon_actuation.py you will see how first the pendulum stabilizes, then it will move towards the white base. You can hit the pendulum close to its tip to change the displacement direction of the cart. If you hit harder it will try to stabilize. 

As any PID system, depending on the gains, it will be more or less stable.


erwincoumans:
Thanks for the PR, that looks like an interesting example. Some change in constraint.py slipped in the PR, we can remove it, right?

PR #2968, opened 2020-08-04
Implicit picking

xhan0619:
Change picking force to always be integrated implicitly and tune parameters in pybullet demo for stable pickling.


erwincoumans:
Thanks for the help, LGTM.

PR #2971, opened 2020-08-07
Fix warning in examples/Utils/ChromeTraceUtil.cpp

WilliamTambellini:
Fix warning in examples/Utils/ChromeTraceUtil.cpp :
bullet3/examples/Utils/ChromeTraceUtil.cpp:122:62: warning:
format ‘%lu’ expects argument of type ‘long unsigned int’,
but argument 4 has type ‘long long unsigned int’ [-Wformat=]:
startTimeDiv1000
Tested:
- examples/ExampleBrowser/App_ExampleBrowser --tracing
- press the P key to write the profiler json file
- head /tmp/timings_0.json
- open /tmp/timings_0.json with chrome://tracing

### DIFF
--- a/examples/Utils/ChromeTraceUtil.cpp
+++ b/examples/Utils/ChromeTraceUtil.cpp
@@ -5,6 +5,7 @@
 #include "LinearMath/btAlignedObjectArray.h"
 #include "Bullet3Common/b3Logging.h"
 #include <stdio.h>
+#include <climits>
 
 struct btTiming
 {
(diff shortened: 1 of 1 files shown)


erwincoumans:
It is an odd warning, since we use PRIu64 and not %lu. We want the unsigned long long int, not casting to an unsigned long int.

WilliamTambellini:
Indeed, I am also considering odd for this gcc to define PRIu64 as "lu" but it really does:
$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609
I also agree that "ts" should better be printed as "llu".
Checking why gcc5 badly resolves PRIu64.

erwincoumans:
Thanks, LGTM.

PR #2980, opened 2020-08-11
Simplify logic for searching files in UrdfFindMeshFile and prioritize…

sippeyxp:
Simplify logic for searching files in UrdfFindMeshFile and prioritize path without prefix.


erwincoumans:
LGTM, thanks!

PR #2984, opened 2020-08-12
Cachebarycenter

fuchuyuan:
Add barycenter caching to static face constraints. Fix effective mass inverse not initialized for multibody node constraint.

### DIFF
--- a/examples/DeformableDemo/GraspDeformable.cpp
+++ b/examples/DeformableDemo/GraspDeformable.cpp
@@ -35,6 +35,9 @@
 #include "../CommonInterfaces/CommonFileIOInterface.h"
 #include "Bullet3Common/b3FileUtils.h"
 
+#include "../ThirdPartyLibs/Wavefront/tiny_obj_loader.h"
+#include "../src/LinearMath/btScalar.h"
+
 ///The GraspDeformable shows grasping a volumetric deformable objects with multibody gripper with moter constraints.
 static btScalar sGripperVerticalVelocity = 0.f;
 static btScalar sGripperClosingTargetVelocity = 0.f;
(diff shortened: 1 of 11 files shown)


erwincoumans:
Thanks for the report. I briefly tried out running the ExampleBrowser deformable examples. to check for regression. 
I think the ExampleBrowser "Deformables/Cloth Friction" example changed, once settled down, pick the red cloth, the blue should follow due to friction. Could you check it?

fuchuyuan:
I turned off the option by default. Can you check again?

erwincoumans:
I had a quick look, the deformable examples in the example browser seem to have degenerated before. It would be good to resolve this. You can see it by compiling/running the example browser and pick the "Cloth Friction" example. The cloth doesn't properly settle down and doesn't maintain a distance. (briefly perturb the blue cloth with the mouse). 

Do you know any parameters to change to make the sim more reasonable?

The "Deformable Self Collision" example seem to maintain some distance between the two cloths, but it has no friction. When enabling friction (psb->setFriction(3)), and carefully moving the bottom deformable/cloth doesn't reasonably move the top cloth.

erwincoumans:
thanks, LGTM

fuchuyuan:
Hi Erwin,

Sorry for the late reply.

The "Cloth Friction" example difference was due to tolerance of CR solver. You can restore the behavior by setting residual of btConjugateResidual to be 1e-4 instead of 1e-8. However, if you play around with 1e-4 a little bit, you will see the same not settling down behavior. This is because when the top cloth is in constraint with the ground, the cloth-cloth interaction will be ignored.


For The "Deformable Self Collision" example, I checked that the behavior is the same before and after the PR. You can set  psb->m_cfg.kDF = 10 and ground friction to be 0.1 to see top cloth being moved with the bottom one.

erwincoumans:
Thanks Kelly, will check it out. By the way, if you have a minute, could you check out this reported cloth issue?
https://github.com/bulletphysics/bullet3/issues/3005

I haven't asked Xuchen to look at it yet, perhaps you have some suggestions for the issue?

PR #3008, opened 2020-08-26
Fix openglwindows example build on netbsd.

devnexen:
not support for pthread_getconcurrency.


erwincoumans:
Thanks, good to hear (net)bsd is still around!

PR #3009, opened 2020-08-27
fix some warnings with GCC5

WilliamTambellini:
ImportBspExample.cpp:173:38: warning: ignoring return value of ‘size_t
fread(void*, size_t, size_t, FILE*)’, declared with attribute
warn_unused_result [-Wunused-result]

PhysicsServerCommandProcessor.cpp:4853:53: warning: name lookup of ‘i’
changed


WilliamTambellini:
@erwincoumans  any review ?

PR #3018, opened 2020-08-31
Add ability to get the original vertex index from the convex hull

DarkDefender:
This makes it possible to know which input vertices were picked to
to define the hull.

With this information it is much easier to map certain features from the
input mesh/point cloud to the output convex hull.

This code is ported from the bundled bullet library in blender.

I'm a Blender developer and we are planning to remove our mini fork of bullet and use the upstream version.

Currently, the only thing that seems to be preventing us from doing so is this small change to the convex hull generation.

If this is not good enough, I'm willing to solve any issues so that we can hopefully be in sync with upstream bullet.

### DIFF
--- a/src/LinearMath/btConvexHullComputer.cpp
+++ b/src/LinearMath/btConvexHullComputer.cpp
@@ -2673,6 +2673,7 @@ btScalar btConvexHullComputer::compute(const void* coords, bool doubleCoords, in
 	}
 
 	vertices.resize(0);
+	original_vertex_index.resize(0);
 	edges.resize(0);
 	faces.resize(0);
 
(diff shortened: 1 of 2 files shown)


DarkDefender:
Ah, I forgot.
https://github.com/bulletphysics/bullet3/pull/2695 also needs to be up streamed for us to be in sync with Blender.

DarkDefender:
The build failures seems to be unrelated to this patch. (?)

erwincoumans:
Thanks! Let's re-trigger (close/re-open issue). Didn't see the reason for build error yet.
By the way, separately it could also be nice to have PyBullet python bindings available (it is used for reinforcement learning, has URDF robot import and reduced coordinate models).

DarkDefender:
Just to double check:

You want me to see if I can include this in the python API as well in a new pull request?

PR #3020, opened 2020-08-31
Fixsegfault

fuchuyuan:
Fix segfault encountered after removing and adding softbody. Add pybullet example to verify the fix.


erwincoumans:
Thanks, will merge it asap.

Issue #2887, opened 2020-06-27
Inconsistency between rayTest and rayTestBatch of pybullet


HiroIshida:
I found the case when rayTest and rayTestBatch yield inconsistent results.  There is an [example](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/examples/batchRayCast.py) code  performing batch ray cast toward r2d2. Based on that, I modified code to test the results of `rayTest` and `rayTestBatch` are consistent. I conducted the test also for `sphere_small`, `sphere2`, `block`, and `humanoid` as well as `r2d2`.  (note that all these data is provided in `pybullet_data`, so one can just run the following code if has pybullet.) The script below does that..
``` python
import pybullet as p
import math
import pybullet_data

def single_test(urdf_file, isDIRECT=True):
    p.connect(p.DIRECT if isDIRECT else p.GUI)
    p.setAdditionalSearchPath(pybullet_data.getDataPath())
    p.loadURDF(urdf_file, [3, 3, 1])

    numRays = 1024; rayLen = 13

    info_lst = []; rayFrom_lst = []; rayTo_lst = []

    for i in range(numRays):
        rayFrom = [0, 0, 1]
        rayTo = [rayLen * math.sin(2. * math.pi * float(i) / numRays),
                rayLen * math.cos(2. * math.pi * float(i) / numRays), 1]

        info = p.rayTest(rayFrom, rayTo)
        info_lst.append(info[0])

        p.addUserDebugLine(rayFrom, rayTo, [1, 0, 0])
        rayFrom_lst.append(rayFrom)
        rayTo_lst.append(rayTo)

    info_lst_batch = p.rayTestBatch(rayFrom_lst, rayTo_lst)

    collide_predicate = lambda info:info[0]!=-1
    info_filtered = list(filter(collide_predicate, info_lst))
    info_batch_filtered = list(filter(collide_predicate, info_lst_batch))

    p.disconnect()

    return info_filtered, info_batch_filtered

def isConsistent(model):
    info_filtered, info_batch_filtered = single_test(model)
    # number of collide ray must be equal 
    return (len(info_filtered) == len(info_batch_filtered)) 

models = ["sphere_small.urdf", "sphere2.urdf", "block.urdf",  # composed of single object
        "r2d2.urdf", "humanoid/humanoid.urdf" # composed of multiple objects 
        ]

lst_consistensy = [isConsistent(model) for model in models]
```
[...]

PR #3035, opened 2020-09-03
fix the issue "Inconsistency between rayTest and rayTestBatch of pybu…

Steven89Liu:
…llet #2887"

we should initialize these variables otherwise these will be random.


erwincoumans:
Thanks, LGTM!

PR #3037, opened 2020-09-06
CWE-190 integer overflow in btConvexHullComputer

issakomi:
Fixed integer overflow in btConvexHullComputer.cpp

Example to reproduce similar issue

```
#include <iostream>
#include <limits.h>

int main(int, char **)
{
	int x = INT_MAX;
	long long y0 = x * 5;
	long long y1 = (long long)x * 5;
	std::cout << "y0=" << y0 << " y1=" << y1 << std::endl;
	return 0;
}
```

Output: y0=2147483643 y1=10737418235

Coverity report:

cwe190-1

### DIFF
--- a/src/LinearMath/btConvexHullComputer.cpp
+++ b/src/LinearMath/btConvexHullComputer.cpp
@@ -105,7 +105,7 @@ public:
 
 		Point64 cross(const Point32& b) const
 		{
-			return Point64(y * b.z - z * b.y, z * b.x - x * b.z, x * b.y - y * b.x);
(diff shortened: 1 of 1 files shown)


issakomi:
Note:
in 
`int64_t dot(const Point32& b) const`
it is so far very clear, otherwise _int64_t_ doesn't make sence

in
`Point64 cross(const Point32& b) const`
... hmm, highly likely too, but if are think it was an intent - i shall update the PR

Thank you

WilliamTambellini (approved):

erwincoumans:
Thanks for the contribution, looks good to me.

PR #3038, opened 2020-09-07
EGL plugin: added EGL_VISIBLE_DEVICE and OpenGL 3.3 selection

BlGene:
This version has the following two changes:

1. you can select the EGL device you want to use by setting the environment variable EGL_RENDER_DEVICE=X. Just start at 0 and iterate up until it tells you "Invalid render_device choice"
2. It requests an EGL context with OpenGL version > 3.3, which is needed for the EGL plugin. On the machines I was using so far EGL was nice and gave me up to date OpenGL versions, but this was needed for my Intel HD / Mesa laptop.


erwincoumans:
Applied here: https://github.com/bulletphysics/bullet3/pull/4024

Issue #2782, opened 2020-05-05
Segfault in changeVisualShape while using eglRenderer


colin-r-schultz:
I have found a segfault that occurs consistently under the following conditions:

1. The eglRenderer plugin is loaded
2. A body is removed
3. A body is created (likely using the same id as the previously removed body)
4. `changeVisualShape` is called on the newly created object

The following code shows the simplest case where I could consistently reproduce this. 
According to `faulthandler`, the segfault occurs at `changeVisualShape`.
```
import pybullet as p
import pkgutil
import pybullet_data

p.connect(p.DIRECT)
egl = pkgutil.get_loader('eglRenderer')
plugin = p.loadPlugin(egl.get_filename(), "_eglRendererPlugin")
print("plugin=", plugin)

p.setAdditionalSearchPath(pybullet_data.getDataPath())
obj1 = p.loadURDF("plane.urdf")
p.removeBody(obj1)
obj2 = p.loadURDF("plane.urdf")
p.changeVisualShape(obj2, -1, rgbaColor=[1, 0, 0, 1])

p.unloadPlugin(plugin)
p.disconnect()
```
My uneducated guess is that this occurs because the removed body and new body share the same id. I'm wondering if, as a workaround, there is a way to stop id pooling and force each new body to have a new id?

erwincoumans:
Thanks for the report, will check it out.

BlGene:
@colin-r-schultz have a look if this works for you: https://github.com/BlGene/bullet3/tree/egl_remove_fix

erwincoumans:
I think this should be fixed in my recent commit:
https://github.com/bulletphysics/bullet3/commit/1bd82e7d81d3a265563538b7882eef80f215f0e8

Please reply/re-open if not.

PR #3039, opened 2020-09-07
EGL fix remove body

BlGene:
This PR aims to address the issue https://github.com/bulletphysics/bullet3/issues/2782.

It also contain the commit from PR https://github.com/bulletphysics/bullet3/pull/3038. So, in case you are planning to look at both PRs, it would be best to look at this one first. Both PRs are however independent.

The main point of this PR is so that I can ask a question about the API.
This code still contains debug messages, to help with development.

### DIFF
--- a/examples/SharedMemory/plugins/eglPlugin/eglRendererVisualShapeConverter.cpp
+++ b/examples/SharedMemory/plugins/eglPlugin/eglRendererVisualShapeConverter.cpp
@@ -1787,7 +1787,8 @@ void EGLRendererVisualShapeConverter::removeVisualShape(int collisionObjectUniqu
 			// recover data from swRenderInstance and use to index m_visualShapes
 			int bodyUniqueId = ptr->m_objectUniqueId;
 			int linkIndex = ptr->m_linkIndex;
-			for (int i = 0; i < m_data->m_visualShapes.size(); i++)
+			// backwards iteration seems safer here, is it needed?
+			for (int i = m_data->m_visualShapes.size()-1; i >= 0 ; i--)
 			{
 				if (m_data->m_visualShapes[i].m_objectUniqueId == bodyUniqueId && m_data->m_visualShapes[i].m_linkIndex == linkIndex)
 				{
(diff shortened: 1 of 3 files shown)


examples/SharedMemory/PhysicsServerCommandProcessor.cpp
```diff
@@ -11209,7 +11209,8 @@ bool PhysicsServerCommandProcessor::processRemoveBodyCommand(const struct Shared
 				{
 					if (m_data->m_pluginManager.getRenderInterface())
 					{
-						m_data->m_pluginManager.getRenderInterface()->removeVisualShape(bodyHandle->m_multiBody->getBaseCollider()->getUserIndex3());
+                                                m_data->m_pluginManager.getRenderInterface()->removeVisualShape(bodyUniqueId);
```
BlGene, line 11212: By my understanding this function call ends up with `EGLRendererVisualShapeConverter::removeVisualShape`. This function wants to have `collisionObjectUniqueId`, which I would assume is the `bodyUniqueId`.

In the master version, when running the simple test script, I was getting values of 1 from `getUserIndex3`, in cases where `collisionObjectUniqueId`,`m_internalInstanceIndex`, and `m_publicGraphicsInstances` are 0.

BlGene: In the code block below, which I can't place a comment next to, it iterates over all of the links, and removes the visual shapes (line 11226). This makes it look like `removeVisualShape` should not be getting `collisionObjectUniqueId` but an index of `m_visualShapes` instead.

examples/SharedMemory/plugins/eglPlugin/eglRendererVisualShapeConverter.cpp
```diff
@@ -1565,6 +1572,24 @@ void EGLRendererVisualShapeConverter::removeVisualShape(int collisionObjectUniqu
 		delete ptr;
 		m_data->m_swRenderInstances.remove(collisionObjectUniqueId);
 	}
+
```
BlGene, line 1575: In the case that we give `removeVisualShape` an `collisionObjectUniqueId` we would then need to iterate over the links and remove them. This would be instead of doing it in `PhysicsServerCommandProcessor.cpp:L11226`

BlGene:
On my current branch, with the modified testrender_egl.py script I get the following output: (see image)

From this it looks like there are a few problems:
1. The ground plane is deleted, it shouldn't be.
2. There should be only one r2d2 at a time
3. Also I'm not sure why its yellow.

image

BlGene:
@colin-r-schultz: does this work for you? #2782

BlGene:
Hello @erwincoumans ,
Could you please have a look at this? With this fix the test script provided in issue #2782 no fails for me.
The modified testrender_egl.py was also running correctly for me.

Tests will fail because the master version of`tinyRendererPlugin/TinyRendererVisualShapeConverter.cpp` was modified.
I can rebase one they are fixed.

erwincoumans:
Thanks, I made various changes to allow textured rendering of deformables in EGL/Tiny and regular OpenGL3 viewport.
Will look into the PR.

BlGene:
@erwincoumans ok, let me know if there are any questions or things that I should change.

erwincoumans:
This should have been fixed in a recent refactor, please reply if not.

PR #3041, opened 2020-09-08
we should clear the jointTorque that is an external force when 

Steven89Liu:
try to calculate the reaction force caused by the constraint force
the second pass.


erwincoumans:
Thanks for the contribution. I need to check and deal with some (broken) unit tests, since the behaviour changes.

chaowei-flexiv:
Hi @erwincoumans , what is the status of this pr? I encountered issues when using jointReactionForces and this pr can fix the issue. Thanks

PR #3048, opened 2020-09-11
enable remove visual shape for softbody

fuchuyuan:


erwincoumans:
Thanks for the PR.

I have a much larger CL upcoming that allow rendering of deformables with textures, in TinyRenderer, EGL and regular debug visualizer. I applied your patch in there, so we can close this one.

erwincoumans:
See https://www.youtube.com/watch?v=_EOvLbEEZHk

PR #3071, opened 2020-09-19
Fix for CMake 3.18.2 + VS2019

niello:
Otherwise the following error appears:

```
CMake Error at engine/Deps/bullet/CMakeLists.txt:77 (STRING):
  STRING sub-command REPLACE requires at least four arguments.


CMAKE_EXE_LINKER_FLAGS_DEBUG=/INCREMENTAL:NO
CMake Error at engine/Deps/bullet/CMakeLists.txt:81 (STRING):
  STRING sub-command REPLACE requires at least four arguments.
```


erwincoumans:
Thanks, LGTM.

PR #3089, opened 2020-10-01
we should use the m_jointMotorForceMultiDof which is indexed by degree

Steven89Liu:
of freedom.


Steven89Liu:
fix the issue of https://pybullet.org/Bullet/phpBB3/viewtopic.php?f=24&t=13074.

erwincoumans:
Thanks for tracking this down and the fix!

PR #3097, opened 2020-10-06
Add <inertia> for missing links in a1.urdf

yxyang:
Currently a few collision frames in a1.urdf ("{FL/FR/RL/RR}_upper_shoulder") do not have inertia defined, and pybullet loads as if they have 1kg mass. This leads to inaccurate simulation results.


erwincoumans:
Thanks a lot, will merge it and upload to pypi into pybullet 3.0.5 soon.

PR #3098, opened 2020-10-07
Improve the kinematic mode

WenlongLu:
- Added dynamics = 0 to collision object flags.
- btRigidBody uses its collision object flags to know if it is static, kinematic, or dynamic. If mass is zero, it has to be static.
- btMultiBody's base and child links also use its collision object flags for dynamic types. If it does not have a collision object, it uses its local member.
- Kinematic link does not have any integration.
- Kinematic object can collide with kinematic and static objects, but the collision will not be processed. (Kinematic object and dynamic object both use "dynamic" group in the broad phase.)


erwincoumans:
Thanks, will have a look soon!

erwincoumans:
Thanks for the PR! It is a tricky PR, since it can potentially break existing behavior. Generally we maintain backward compatibility, so not breaking the API. (unfortunately breaking changes or bugs slip in, but that is usually not intended).

1) new args to setMassProps argument should be optional. In that case, perhaps the arg would be set to 'uninitialized' and behave exactly as before (mass>0 -> dynamic, mass==0 -> static).
2) We don't use C++11, so no override.
3) The BasicExample cannot be modified to keep it basic. Since you already added a new example, that could be enough?

Checking the changes to files in Bullet/src will take a bit more time, to avoid regressions.

WenlongLu:
1. setMassProps has a default parameter to be dynamic. In this way, the original behavior is preserved.
2. override removed
3. BasicExample reverted

4. I merged the recent commit of the master to resolve the conflicts. (I am not sure how to not display other people's changes in this pull request. If I create a new pull request, commits merged from master won't show up.)

PR #3115, opened 2020-10-19
Compilation warning fixes

niello:
* new btSdfCollisionShapeInternalData() may not be allocated aligned 16
* Two cases of signed/unsigned mismatch


erwincoumans:
Looks good to me, thanks!

PR #3117, opened 2020-10-21
Improve the kinematic mode

WenlongLu:
- Added dynamics = 0 to collision object flags.
- btRigidBody uses its collision object flags to know if it is static, kinematic, or dynamic. If mass is zero, it has to be static.
- btMultiBody's base and child links also use its collision object flags for dynamic types. If it does not have a collision object, it uses its local member.
- Kinematic link does not have any integration.
- Kinematic object can collide with kinematic and static objects, but the collision will not be processed. (Kinematic object and dynamic object both use "dynamic" group in the broad phase.)


erwincoumans:
The PR is too ambitious to be accepted. Two features seem requested, it will be better to create a separate PR for each
1) allow btMultiBody to be kinematic, for one-way contact interaction with friction, similar to the [KinematicRigidBodyExample](https://github.com/bulletphysics/bullet3/pull/3120)
2) allow a contact query for kinematic btMultiBody versus other objects.

Only changes that are strictly required, so no additional dynamicsType etc.

erwincoumans:
When running the KinematicMultiBody demo, the multibody gains energy, moves uncontrollably and doesn't simulate properly, is that intended?

PR #3126, opened 2020-10-28
Made a small fix for MCST lcc compiler and e2k (Elbrus 2000) architecture

r-a-sattarov:
- GwenOpenGLTest/UnitTest.cpp: strip UTF-8 BOM for compatibility with MCST lcc compiler < 1.24
- X11/Xmd.h: added define for MCST e2k (Elbrus 2000) architecture


erwincoumans:
Sure, LGTM. Thanks!

PR #3129, opened 2020-10-29
Kinematic Multibody

WenlongLu:


WenlongLu:
@erwincoumans Created this PR for only kinematic multibody changes. You may take a look first. There are two commits coming: 
1) Remove the dynamic type field from btMultiBody and btMultiBodyLink, assuming that they always have a collision body. I tried this but it crashed. I need to debug it.
2) Update the kinematic example to be similar to the rigid body example, so we can see effects of the kinematic velocity. (Previously you said this kinematic multibody example was behaving crazily. It was possible I made a mistake in one of the revisions. The current state is good. You may take a look.)

erwincoumans:
Thanks for the PR! btMultiBody always has a collision body in PyBullet and using the Bullet Robotics/SharedMemory C-API, not in general when using the regular C++ API (directly creating btMultiBody etc).
I'll have a look.

PR #3158, opened 2020-11-20
Fix the bug where forces are ignore at a static/kinematic link 

WenlongLu:
... while one of its ancestors is not.


erwincoumans:
thanks for the fix!

PR #3162, opened 2020-11-24
Restrict kinematic multibody use cases to only a full chain ...

WenlongLu:
of kinematic joints.


erwincoumans:
Thanks, LGTM. The CI is locking on one Docker config, let's ignore it.

PR #3163, opened 2020-11-25
Debug patch for btGeneric6DofSpring2Constraint

jhwang7628:
This problem was preventing our code to work in debug build due to the btAssert check here (https://github.com/bulletphysics/bullet3/blob/master/src/LinearMath/btVector3.h#L305). I am aware of the `safeNormalize` function (https://github.com/bulletphysics/bullet3/blob/master/src/LinearMath/btVector3.h#L286), however, it does not seem to have the vectorization capability as the regular `normalization` and thus the proposed change.

### DIFF
--- a/src/BulletDynamics/ConstraintSolver/btGeneric6DofSpring2Constraint.cpp
+++ b/src/BulletDynamics/ConstraintSolver/btGeneric6DofSpring2Constraint.cpp
@@ -403,9 +403,13 @@ void btGeneric6DofSpring2Constraint::calculateAngleInfo()
 			btAssert(false);
 	}
 
-	m_calculatedAxis[0].normalize();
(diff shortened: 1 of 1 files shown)


jhwang7628:
Not sure how to fix the travis build error. Following is the build error log
```
Worker information
There was an error in the .travis.yml file from which we could not recover.
Unfortunately, we do not know much about this error.
Please review https://docs.travis-ci.com, or contact us at support@travis-ci.com with the error ID: 8d252f69a4cf46ca8682f4a64e4f41de
```
Seems like the failing builds are from `BUILD_NAME=XENIAL_CLANG DOCKER_FILE="ubuntu-xenial"` and `BUILD_NAME=TRUSTY_CLANG SUDO=sudo`

Please advise.

PR #3165, opened 2020-11-25
safely normalize in 6dof springs and small refactoring in vector3 class

jhwang7628:
This problem was preventing our code to work in debug build due to the btAssert check [here](https://github.com/bulletphysics/bullet3/blob/master/src/LinearMath/btVector3.h#L305).


jhwang7628:
Can we get some action on this?

PR #3182, opened 2020-12-04
Sharedmemory example build fix.

devnexen:
dependencies on prebuild openvr without BSD versions.


erwincoumans:
I just disabled OpenVR compilation by default, so this PR is not needed anymore.

erwincoumans:
Thanks for the help!

PR #3195, opened 2020-12-15
Disable velocity calculation in kinematic multibody mode by default

WenlongLu:
We haven't implemented this for the multi body joints. To be consistent, let's disable it for the base as well, and let user to set the velocities.


erwincoumans:
Thanks, LGTM.

PR #3206, opened 2020-12-23
bump up version to 3.08

erwincoumans:


akien-mga:
You forgot to bump the version in `btScalar.h`. Version bumps were also missing for 3.06 and 3.07 until fixed in `master` with #3168, maybe it would be good to document how version bumps should be done / write a script that validates it before tagging a release?

PR #3219, opened 2021-01-09
Fixed sphere sliders in Extended Tutorial: Incline Plane

derek-mcblane:
Sphere slider's are incorrect. Rolling/Spinning friction sliders both set restitution. Restitution slider is inactive.

Added callbacks for sphere's spinning and rolling friction sliders.
Set sphere restitution callback for slider.


derek-mcblane:
Please rebuild

derek-mcblane:
Linking to Issue #4397

PR #3220, opened 2021-01-10
btRaycastVehicle.cpp: forward impulse way too strong when sideImpulse==0

stephengold:
This pull request fixes a recently discovered bug in `btRaycastVehicle`. When a vehicle at rest starts moving, sometimes its forward impulse is much too large, causing it to immediately reach its steady-state speed.

In my application (a particular vehicle configuration on a level surface) the impulse was 140x too large about 18% of the time. The bug was highly reproducible: it happened for some starting locations on the surface but not for others. I tracked it back to line 612 of btRaycastVehicle.cpp: `m_sideImpulse[wheel]` was either `-0` (for bad starting locations) or roughly `-2e-09` (for good ones).

For such a tiny side impulse to reduce the forward impulse by 140x makes no sense, physically. I suspect the test in line 612 was originally a hand optimization from at time when the loop was scaling `m_sideImpulse[wheel]` only. Removing the test fixed the bug.


stephengold:
After 8 months, I'm curious why this PR hasn't been integrated. If the Bullet project is no longer accepting PRs from outsiders, or if it's no longer accepting PRs to "BulletDynamics/Vehicle", that would be nice to know.

stephengold:
@erwincoumans What's the next step for this PR?

stephengold:
Contact me by e-mail if you want the diff for this bugfix.

PR #3225, opened 2021-01-12
128 bits multiply optional optimisations for ARM64 proposal

devnexen:


glebm:
Does it improve the performance? 
Any reason not to `#define USE_ARM_64_ASM __aarch64__`?

devnexen:
Was more a WIP but finally does not bring much.

PR #3234, opened 2021-01-19
Fix TARGET_LINK_LIBRARIES for static linking

### LINKED ISSUE #3233: CMake does not TARGET_LINK_LIBRARIES if BUILD_SHARED_LIBS=OFF
`TARGET_LINK_LIBRARIES` should also be present for static linking

Example:

```cmake
IF (BUILD_SHARED_LIBS)
  TARGET_LINK_LIBRARIES(BulletCollision LinearMath)
ENDIF (BUILD_SHARED_LIBS)
```

This should just be:

```cmake
TARGET_LINK_LIBRARIES(BulletCollision LinearMath)
```

Otherwise, it's difficult to use Bullet statically from another project, the user must manually specify the correct order and all the dependencies. This also means when installing static `.a` libraries, the generated pkg-config files are wrong.

glebm:
Fixes link errors when using bullet built for static linking.

Fixes #3233


glebm:
@erwincoumans The travis build failed through no fault of this PR btw

PR #3237, opened 2021-01-20
CMake: Use Threads instead of pthread directly

glebm:
On some systems, such as Android, threads are not provided via `-lpthread`.

CMake comes with a built-in module for finding the correct threads library.

Use it to fix the linking issue on Android and possibly other systems.


jvoisin (approved):

erwincoumans:
Thanks, wasn't aware of Threads, thanks for the contribution. LGTM.

PR #3238, opened 2021-01-20
Add linkNames argument to CreateMultibody

yycho0108:
**Problem**

In `pybullet.createMultiBody`, there is no way to specify the link names that can later be retrieved via `pybullet.getJointInfo`. Since bullet does not necessarily preserve the exact order in which the arrays of link properties are specified in the arguments when assigning link indices, for programmatically defined robots (i.e. not loaded from a URDF/SDF) it can be difficult to exactly figure out which joint corresponds to which joint index.

There are workarounds for this, but I think something like this is the cleanest way to deal with the underlying problem. Furthermore, I think the ability to specify link names in `createMultibody` is useful in and of itself.

**Solution**
This PR adds a `linkNames` argument to `createMultiBody()` and other places in the code to where the argument propagates, so that it would be possible to specify the joint names at the time of creation of the robot and query them afterwards.

**Minimal example**
```python
#!/usr/bin/env python3
import pybullet as p

# Connect to physics.
physics_client_id = p.connect(p.GUI)

# Create dummy collision/visual shapes.
col_id = p.createCollisionShape(p.GEOM_SPHERE, radius=1.0)
vis_id = p.createVisualShape(p.GEOM_SPHERE, radius=1.0)

# Create a minimal configuration that illustrates
# the usefulness of `linkNames`.
link_names = ['a', 'b', 'c']
parent_indices = [2, 0, 1]
num_links = len(link_names)

body_id = p.createMultiBody(
    baseMass=1,
    baseCollisionShapeIndex=col_id,
    baseVisualShapeIndex=vis_id,
    basePosition=[0, 0, 0],
    baseOrientation=[0, 0, 0, 1],
    baseInertialFramePosition=[0, 0, 0],
    baseInertialFrameOrientation=[0, 0, 0, 1],

    linkMasses=num_links * [1],
    linkCollisionShapeIndices=num_links * [col_id],
    linkVisualShapeIndices=num_links * [vis_id],
    linkPositions=num_links * [[0, 0, 0]],
    linkOrientations=num_links * [[0, 0, 0, 1]],
    linkInertialFramePositions=num_links * [[0, 0, 0]],
    linkInertialFrameOrientations=num_links * [[0, 0, 0, 1]],
    linkParentIndices=parent_indices,
    linkJointTypes=num_links * [p.JOINT_REVOLUTE],
    linkJointAxis=num_links * [[0, 0, 1]],
    linkNames=link_names,
    physicsClientId=physics_client_id)

# In this example, the indices happen to be exactly the same as the input.
for i in range(num_links):
    joint_info = p.getJointInfo(
        body_id, jointIndex=i, physicsClientId=physics_client_id)
    print('LINK #{} = {}'.format(i, joint_info[1]))

# OUTPUT:
# LINK #0 = b'b'
# LINK #1 = b'a'
# LINK #2 = b'c'
```
Thank you for this amazing library! I hope this can be a positive contribution to pybullet.
Let me know if there are things that I can do to make this PR more useful!

### DIFF
--- a/examples/SharedMemory/PhysicsClientC_API.cpp
+++ b/examples/SharedMemory/PhysicsClientC_API.cpp
@@ -2052,7 +2052,8 @@ B3_SHARED_API int b3CreateMultiBodyLink(b3SharedMemoryCommandHandle commandHandl
 										const double linkInertialFrameOrientation[4],
 										int linkParentIndex,
 										int linkJointType,
-										const double linkJointAxis[3])
+										const double linkJointAxis[3],
+										const char* const linkName)
 {
 	struct SharedMemoryCommand* command = (struct SharedMemoryCommand*)commandHandle;
 	b3Assert(command);
@@ -2097,6 +2098,7 @@ B3_SHARED_API int b3CreateMultiBodyLink(b3SharedMemoryCommandHandle commandHandl
 			command->m_createMultiBodyArgs.m_linkJointAxis[3 * linkIndex + 2] = linkJointAxis[2];
 
 			command->m_createMultiBodyArgs.m_linkMasses[linkIndex] = linkMass;
+			command->m_createMultiBodyArgs.m_linkNames[linkIndex] = linkName;
 			command->m_createMultiBodyArgs.m_numLinks++;
 			return numLinks;
 		}
--- a/examples/SharedMemory/PhysicsClientC_API.h
+++ b/examples/SharedMemory/PhysicsClientC_API.h
@@ -552,7 +552,8 @@ extern "C"
 											const double linkInertialFrameOrientation[/*4*/],
 											int linkParentIndex,
 											int linkJointType,
-											const double linkJointAxis[/*3*/]);
+											const double linkJointAxis[/*3*/],
+											const char* const linkName);
 
 	//batch creation is an performance feature to create a large number of multi bodies in one command
 	B3_SHARED_API int b3CreateMultiBodySetBatchPositions(b3PhysicsClientHandle physClient, b3SharedMemoryCommandHandle commandHandle, double* batchPositions, int numBatchObjects);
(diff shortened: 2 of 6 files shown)


erwincoumans:
Thanks a lot for the contribution! Will check it out soon, and merge it if all is fine.

erwincoumans:
Thanks, looks good to me.

erwincoumans:
Note that you can use a flag, URDF_MAINTAIN_LINK_ORDER, to maintain the link order. This only works if the links and joints in the URDF are added in the proper order: parents go before childs, for both links and joints.

erwincoumans:
Have to revert the PR, since we cannot store pointers to local memory in the shared memory structures.

erwincoumans:
We could have a link uid instead, storing integers.

yycho0108:
Thanks for letting me know of the issue @erwincoumans - sorry for the late response. Can you point me to the exact use-case that would result in an error, so that I could address the bug if possible? Furthermore, can you clarify what you're referring to as pointers to `local` memory? As far as my understanding goes, the pointer to `linkName` isn't ever a local stack pointer, and it should remain in scope at least up until its usage (in `getLinkName()`...) during loading, so I'd like to know exactly when it becomes problematic.

I had considered alternatives like `m_linkNames[MAX_CREATE_MULTI_BODY_LINKS][MAX_FILENAME_LENGTH]` which felt excessively wasteful in memory, so that's probably never going to work.
We could certainly consider storing link uids instead, but wouldn't that add inconsistencies in the API where all other loaders (besides `createMultibody...` functions) would also need to support setting and querying uids, as opposed to the `linkName` and `jointName` parameters which are already present in the current API?
Otherwise, are you suggesting that the uids will be passed in as `integer`s would then be converted as `string`s would become the default `linkName` instead of the previous default (sequentially labeled `link%d`)? So that `createMultibody...(linkUids=[8,10,15])` would result in `link8, link10, link15`?

It would be ideal if the `linkNames` parameter can stay, but if that would not be possible due to fundamental limitations (and it results in a bug / undesirable behavior) then we should definitely revert the PR and consider alternatives. In order to assess that, I'd be grateful if you could share the workflow that would result in an error.

PR #3239, opened 2021-01-20
Separate kp and kd channels for spherical motor control

WenlongLu:


examples/pybullet/pybullet.c
```diff
@@ -3199,17 +3199,29 @@ static PyObject* pybullet_setJointMotorControlMultiDof(PyObject* self, PyObject*
...
 	int physicsClientId = 0;
-	static char* kwlist[] = {"bodyUniqueId", "jointIndex", "controlMode", "targetPosition", "targetVelocity", "force", "positionGain", "velocityGain", "maxVelocity", "physicsClientId", NULL};
+	static char* kwlist[] = {"bodyUniqueId", "jointIndex", "controlMode", "targetPosition", "targetVelocity", "force", "positionGain", "velocityGain", "maxVelocity", "damping", "physicsClientId", NULL};
 	if (!PyArg_ParseTupleAndKeywords(args, keywds, "iii|OOOdddi", kwlist, &bodyUniqueId, &jointIndex, &controlMode,
-									 &targetPositionObj, &targetVelocityObj, &targetForceObj, &kp, &kd, &maxVelocity, &physicsClientId))
+								 	&targetPositionObj, &targetVelocityObj, &targetForceObj, &kpArray[0], &kdArray[0], &maxVelocity, &physicsClientId))
 	{
-		return NULL;
+		PyErr_Clear();
+		if (!PyArg_ParseTupleAndKeywords(args, keywds, "iii|OOOOOOOdi", kwlist, &bodyUniqueId, &jointIndex, &controlMode,
```
erwincoumans, line 3220: kwlist has 11 entries, but the string "iii|OOOOOOOdi" has 12, that won't work well.

examples/pybullet/pybullet.c
```diff
@@ -3199,17 +3199,29 @@ static PyObject* pybullet_setJointMotorControlMultiDof(PyObject* self, PyObject*
...
 	double maxVelocity = -1;
+	double dampingArray[3] = {1.0, 1.0, 1.0};
+	int dampingSize = 0;
+	PyObject* dampingObj = 0;
 	b3PhysicsClientHandle sm = 0;
 
 	int physicsClientId = 0;
-	static char* kwlist[] = {"bodyUniqueId", "jointIndex", "controlMode", "targetPosition", "targetVelocity", "force", "positionGain", "velocityGain", "maxVelocity", "physicsClientId", NULL};
+	static char* kwlist[] = {"bodyUniqueId", "jointIndex", "controlMode", "targetPosition", "targetVelocity", "force", "positionGain", "velocityGain", "maxVelocity", "damping", "physicsClientId", NULL};
 	if (!PyArg_ParseTupleAndKeywords(args, keywds, "iii|OOOdddi", kwlist, &bodyUniqueId, &jointIndex, &controlMode,
```
erwincoumans, line 3216: kwlist has 11 entries, but the string "iii|OOOdddi" has only 10.

src/BulletDynamics/Featherstone/btMultiBodySphericalJointMotor.cpp
```diff
@@ -152,12 +156,16 @@ btQuaternion relRot = currentQuat.inverse() * desiredQuat;
...
 				{
 					btVector3 constraintNormalAng = frameAworld.getColumn(row % 3);
-					posError = m_kp*angleDiff[row % 3];
+					double kp = m_use_multi_dof_params ? m_kp[row % 3] : m_kp[0];
+					posError = kp*angleDiff[row % 3];
+					double max_applied_impulse = m_use_multi_dof_params ? m_maxAppliedImpulseMultiDof[row % 3] : m_maxAppliedImpulse;
 					fillMultiBodyConstraint(constraintRow, data, 0, 0, constraintNormalAng,
 						btVector3(0,0,0), dummy, dummy,
-						posError,
+						posError + velocityError,
```
erwincoumans, line 164: adding a velocityError here is a change in behavior. Also, there is a separate argument to add a desired relative velocity. If this change is required, it is best to create a separate PR/CL for that, so we can easily undo it (if it causes issues anywhere).

WenlongLu:
Thanks for the comments!

1. The velocityError added to fillMultiBodyConstraint call has been removed/restored.
2. The original pybullet function should have been fixed (with one more optional parameter for damping.)

erwincoumans:
This CL breaks the spherical motor. See for example this example: humanoidMotionCapture.py
https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/examples/humanoidMotionCapture.py

erwincoumans:
I had a look, and m_use_multi_dof_params is not properly initialized (only for the first 7 values).
The 'm_hasDesiredStateFlags' is probably a better way, add a flag for that (see EnumSimDesiredStateUpdateFlags enums in SharedMemoryCommands.h)

erwincoumans:
The example is still broken (https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/examples/humanoidMotionCapture.py).

Damping is only initialized for the first component, so 2 components are not initialized.
It is not clear if SIM_DESIRED_STATE_USE_MULTI_DOF refers to kd, kp or damping, and multidof damping is set, but not multidof kp/kd, then logic can easily break. So better to remove the SIM_DESIRED_STATE_USE_MULTI_DOF flag and check for each individual dof if the appropriate flag is set.

erwincoumans:
For spherical joints, kd, kp and damping have either 1 or 3 components (0 or 0,1,2).
So dofCount check should be >=0  && dofCount < 3, not <=4.
~~~
if ((dofIndex >= 0) && (dofIndex < MAX_DEGREE_OF_FREEDOM ) && dofCount >= 0 && dofCount <= 4)
~~~

erwincoumans:
Somehow the merging cause too many diff files in this PR.
I applied the patch with some changes, so it doesn't break some of the existing examples.
See https://github.com/erwincoumans/bullet3/commit/8b8c1af6a4b9db8b596af57da1a7073a3db862d9
Best to reset your fork and update to latest. Please check if the features work are as requested.
Thanks!

PR #3246, opened 2021-01-28
Fix Stable-PD Control bug on First Order Taylor approximation of next `q` state

Danfoa:
Good Day,

I was exploring the example of the Stable PD and its comparison to the other PD and P controllers when I found a small bug on the implementation. I assume the stable PD implementation is based on "[Stable Proportional-Derivative Controllers](https://www.cc.gatech.edu/~turk/my_papers/stable_pd.pdf)". If it is, then you have a mistake because you are not using the first-order Taylor approximation of the next position step: 

equation

This mistake was located in the original repository precisely in the following lines, specially line 130: 

https://github.com/bulletphysics/bullet3/blob/afa4fb54505fd071103b8e2e8793c38fd40f6fb6/examples/pybullet/examples/pdControllerStable.py#L128-L133

You see the original equations are based on the use of the Taylor approximation in order to "look ahead" to the evolution of the dynamics of the system: 

equation

With the correction, the new lines of code look like this:

https://github.com/bulletphysics/bullet3/blob/f719c27b04ed754f61e3a356eaf18bb11fb9abbc/examples/pybullet/examples/pdControllerStable.py#L121-L143

I am almost sure that was the only bug in the code. However, I took the liberty of adding some comments, cleaning the code, and simplifying it a little bit the code in hopes of making it more understandable. 

I modified both the multi DOF version and the normal version and tested both the `testHumanoid.py`, `testQuadruped.py` for the multi DOF version, and the `pdControl.py` example. Everything seems to be working appropriately!

### DIFF
--- a/examples/pybullet/examples/pdControllerStable.py
+++ b/examples/pybullet/examples/pdControllerStable.py
@@ -70,43 +70,34 @@ class PDControllerStableMultiDof(object):
     Kp = np.diagflat(kps)
     Kd = np.diagflat(kds)
 
-    p = Kp.dot(qError)
+    # Compute -Kp(q + qdot - qdes)
+    p_term = Kp.dot(qError - qdot*timeStep)
+    # Compute -Kd(qdot - qdotdes)
+    d_term = Kd.dot(qdoterr)
 
-    #np.savetxt("pb_qError.csv", qError, delimiter=",")
-    #np.savetxt("pb_p.csv", p, delimiter=",")
-
-    d = Kd.dot(qdoterr)
-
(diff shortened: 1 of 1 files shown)


erwincoumans:
Thanks for making the effort of fixing this! LGTM. Note that by default, we use a C++ implementation, not the Python version. I implemented the Python version for clarity.

Danfoa:
Good Day @erwincoumans, happy it was accepted.

I did not know there was a C++ implementation, could you share a link to it for clearer access to it on the repo (or API docs or on an example, the only references that I found of stable PD in the repo directed me to the python implementation). I believe people could use clearer access to this information.

PR #3288, opened 2021-03-01
Fix a null dereference in UrdfParser

vabr-g:
When the parser sees no <inertial> tag inside a <deformable>, it logs an error but does not return false. So in the next step, a null pointer is dereferenced.

This can be tested with loading the following URDF:
```
<?xml version="1.0"?>
<robot name="rhythm_carrot_sticks">
  <deformable name="bag">
  </deformable>
</robot>
```


erwincoumans:
Thanks, looks good to me!

PR #3289, opened 2021-03-02
Fix stack overflow in parsing URDF files in Bullet

vabr-g:
When reading response from the Bullet server after a "load URDF" command, the client code allocated and zeroed buffers for the data chunks with a "+1" to account for the terminating zero. This was when the buffer pointer was interpreted as a `char *`, so it meant "+ 1 byte".

When reading those buffers, however, the associated pointer was a `void *`, so reading `sizeof(void*)` (8 on my machine) bytes at a time. Therefore it was reading up to 7 bytes past the allocated (and zeroed) memory.

The change fixes that by changing the "+ 1" to "+ sizeof(void*)". At one place it also extends the zeroing to the final buffer position (missing "+ 1" in the original).


erwincoumans:
Thanks for tracking this down, lgtm.

PR #3291, opened 2021-03-02
Optimize calls to find*() for a single char.

vabr-g:
The character literal overload is more efficient.


erwincoumans:
Thanks, LGTM!

PR #3293, opened 2021-03-03
heightfield: Fix PHY_FLOAT data handling in double-precision mode

glebm:
Previously, the heightfield constructor always assumed `PHY_FLOAT` meant `btScalar`, contrary to the documentation and inconsistent with other usages of `PHY_FLOAT`.

Adds new constructors that distinguish between `float` and `double`. The legacy constructor is still there for backwards compatibility.

This allows the client to use `float` or `double` heightfield data regardless of `BT_USE_DOUBLE_PRECISION`.

Example use-case is to avoid data conversion when the source data is `float` but `btScalar` is `double`: https://gitlab.com/OpenMW/openmw/-/merge_requests/647/diffs

/cc @Zylann @psi29a @ddrone @erwincoumans


erwincoumans:
Good point, thanks for the contribution!

PR #3300, opened 2021-03-05
one cmake option per one extra package

al-tu:
This feature allows to avoid building long chain of dependant modules (BUILD_EXTRAS implies BUILD_CPU_DEMOS that implies BUILD_BULLET2_DEMOS) when only a small subset is desirable. Actual use case: i needed to build core module and hacd extra module.


al-tu:
fixed linker errors by setting relevant cmake options default values to ON, this way new functionality won't break existing build routines

erwincoumans:
Thanks, LGTM.

PR #3311, opened 2021-03-16
allow to compile pybullet using btDiscreteDynamicsWorld (no btMultiBody, no btSoftBody, no btDeformableBody)

erwincoumans:


setup.py
```diff
@@ -501,7 +501,7 @@ def _single_compile(obj):
 
 setup(
     name='pybullet',
-    version='3.0.9',
+    version='3.1.0',
```
glebm, line 504: Can you please also bump the C++ version throughout?

Please bump it here as well, as it still says 3.0.7

https://github.com/bulletphysics/bullet3/blob/87e668f6b2a883b4ef63db8a07c8e9283916e9d9/src/LinearMath/btScalar.h#L27-L28

PR #3330, opened 2021-04-01
Setup functions to expose python discrete collision

cadop:
Work towards exposing the `performDiscreteCollisionDetection` in `btCollisionWorld.cpp` to pybullet.  This should allow users to call the discrete collision detection without running a simulation step.  

Implementation followed the style/locations of `stepSimulation()` since it seems to be a similar command, just different results. 

This implementation just calls the method, doesn't return anything to the python user. So the idea would be to:
```
# do something
p.discreteCollision()
p.getContactPoints(...)
```

@erwincoumans  this was to implement the comment you made in: https://github.com/bulletphysics/bullet3/pull/2900 .  

It's set as draft now, so please let me know if this is what you expected or any pointers to fix it up.


erwincoumans:
Thanks! There is already a PR in flight, https://github.com/bulletphysics/bullet3/pull/3331 probably similar.

erwincoumans:
The feature has been added separately. Thanks for the PR!

PR #3341, opened 2021-04-06
Replace xrange with range

jaraujo98:
The `xrange` function was deprecated in Python 3. Removing it from `minitaur_env_randomizer.py` for compatibility.


erwincoumans:
Thanks, LGTM.

PR #3344, opened 2021-04-08
Magic spring

maphysart:


erwincoumans:
Thanks. Can you provide some description what the PR is trying to do? Does the magic spring act on rigid or multi bodies? What are its feature(s)?

maphysart:
Hi,
Thanks for your reply. Sorry, I made a mistake. Please ignore the pull
request.
BTW, would you like to give me some advice? I tried to use multibody and
btMultiBodySphericalJointMotor to create a chain so that it will behave
like a rope but also can restore to its initial poses/angles when static.
Eg, the initial pose may be the rope winds up 45 degrees.  I use the
following config:
link number = 16
base link mass = 1.0
other links' mass = 1.0
I setup the multibody with floating set to be true, and use
the btMultiBodyPoint2Point to drive the base link oscilatoring around  zero
point in axis.
I bind btMultiBodySphericalJointMotor to each link, in setPositionTarget to
btQuaternion(0,0,0,1) for other 15 links, and set the first link to rorate
around x axis 45 degree.
I wish to use the m_kp to control how the rope will restore back to initial
pose. But it seems like that the rope behave slowly even I set m_kp to
0.0f.

Another problem is that btMultiBodySphericalJointMotor does not work when
the rope is static and floating set to true. I don't know why.

Any suggestions are realy appreciated. Thanks a lot.

erwincoumans:
ok, closing then. I would recommend prototyping in PyBullet using Python, hard to tell what is going wrong in C++ code.

PR #3349, opened 2021-04-17
fix soft body triangular area calculation

jeffrey-cochran:
I'm not sure if this was deliberate, but if `area` is supposed to be the area of a triangle with edges `x1 - x0` and `x2 - x0`, it's missing a factor of `0.5`.

### DIFF
--- a/src/BulletSoftBody/btSoftBodyInternals.h
+++ b/src/BulletSoftBody/btSoftBodyInternals.h
@@ -1285,7 +1285,7 @@ static inline btScalar AreaOf(const btVector3& x0,
 	const btVector3 a = x1 - x0;
 	const btVector3 b = x2 - x0;
 	const btVector3 cr = btCross(a, b);
-	const btScalar area = cr.length();
+	const btScalar area = cr.length() / 2.;
 	return (area);
 }
 


jeffrey-cochran:
It looks like maybe this *is* deliberate?

```			const btScalar twicearea = AreaOf(f.m_n[0]->m_x,
											  f.m_n[1]->m_x,
											  f.m_n[2]->m_x);```

I guess I don't see why you wouldn't compute the correct area and multiply by two if you want `twicearea`

erwincoumans:
Area should indeed be half ('volume' also seems to miss a factor). I'm not very familiar with all the details of the soft body code, and I'd rather not start refactoring the soft body code, some comment would help indeed, but a fix is more involved than just adding a 0.5 factor, since some other parts may assume twice area indeed. If you need the API better to write your own version.

jeffrey-cochran:
> Area should indeed be half ('volume' also seems to miss a factor). I'm not very familiar with all the details of the soft body code, and I'd rather not start refactoring the soft body code, some comment would help indeed, but a fix is more involved than just adding a 0.5 factor, since some other parts may assume twice area indeed. If you need the API better to write your own version.

Yes, I agree. A refactor would be a pain in the neck. I mostly created the PR to hear your thoughts and discover if you knew of some logic behind it that would build my confidence in the soft body code's correctness. Thanks. I'll review the rest of the code.

PR #3350, opened 2021-04-19
examples/RobotSimulator/CMake: fixed warning (mis-matching arguments)

r-a-sattarov:
Fixed CMake Warning in examples/RobotSimulator/CMakeLists.txt

`A logical block opening on the line`
`/bullet3/examples/RobotSimulator/CMakeLists.txt:149 (IF)`
`closes on the line`
`/bullet3/examples/RobotSimulator/CMakeLists.txt:152 (ENDIF)`
`with mis-matching arguments.`


erwincoumans:
Good point, thanks for the PR. Looks good to me.

PR #3354, opened 2021-04-22
Fix memory leak. Contact infomation should be cleared regardless if d…

fuchuyuan:
Contact information should be cleared regardless if deformable is sleeping or not.


erwincoumans:
Thanks. There is a similar PR that was a bit earlier, so I pick that one (assuming they are the same):
https://github.com/bulletphysics/bullet3/pull/3353

PR #3848, opened 2021-05-14
Fix early exit in heightfield raycast

pouleyKetchoupp:
When deciding the ray is too short to use the accelerated structure and a simple raycast is performed, the method now returns to avoid processing it a second time using the accelerated structure.


glebm (approved):

erwincoumans:
Thanks, LGTM.

PR #3865, opened 2021-05-26
cmake fix for users already having /GR- 

datgame:
they ended up having a '-' compile warning per file


erwincoumans:
Thanks, LGTM.

PR #3901, opened 2021-06-30
Fixed installation issue of headers with hpp extension.

nash169:
Minor fix in the CMakeLists.txt to install headers with .hpp extension.


ZanovelloAlberto:
not sure why this is still pending

PR #3920, opened 2021-07-29
Migrate to Stable-Baselines3

araffin:
As Stable-Baselines (SB2) is now deprecated, this PR updates example code to the actively maintained SB3 (where time feature is not needed anymore as timeout are properly handled).
I also reformatted the code using black and used f-strings (python 3.6+) for prints.


erwincoumans:
Thanks for the update! Is there a list of pybullet environments that are tested with SB-3?

araffin:
> Is there a list of pybullet environments that are tested with SB-3?

If by tested you mean tuned hyperparams + pre-trained models, you can take a look at the [RL Zoo](https://github.com/DLR-RM/rl-baselines3-zoo), the pre-trained models (with associated performance) are listed in https://github.com/DLR-RM/rl-baselines3-zoo/blob/master/benchmark.md ;) (the hyperparameters are in the `hyperparams/` folder)

PR #3921, opened 2021-07-29
Create URDF for white xarm

ayzaan:


erwincoumans:
Thanks for the contribution, sorry for the huge delay, haven't taken the time to check out pull requests.

PR #3925, opened 2021-08-04
Always install .pc

Ace314159:
It would be useful to always install the .pc file even with MSVC since there are Windows alternatives to pkg-config like pkg-conf


Ace314159:
@erwincoumans I'd be happy to add a flag as well. Please, let me know your decision.

PR #3931, opened 2021-08-07
Fix crash when btSoftBody collides with btCompoundShape

notrabs:
Fix crash when `btSoftBody` collides with a `btBvhTriangleMeshShape` inside a `btCompoundShape`

- `triBody->getCollisionObject()` returns the `btCompoundShape` in that situation, which does not implement the `btConcaveShape` interface required for the `processAllTriangles()` call later
- `triBody->getCollisionShape()` already returns the shape to check, so it can be used instead

Reproduction of crash: 
1. Apply https://github.com/notrabs/bullet3/commit/0009830cf1be0440286ec320336f73836b1109d2 to SoftDemo.cpp
2. Start `SoftBody>Cluster Deform` Demo
3. The crash happens as soon as the torus gets close to the ground


erwincoumans:
Thanks, LGTM

PR #3945, opened 2021-08-25
Fix: remove cmake policy 0115 errors

craftablescience:
As of CMake 3.20+, a new policy was introduced, which outputs a warning when a source file does not have a file extension. I see no reason why these files should not have an extension, so I added an extension (`.h`) to remove the warnings. If the extension was purposely omitted, I can instead disable the policy, but it appears to have no effect besides removing the warnings.

By making this PR I assert I am legally allowed to contribute this modification under the zlib license.

The warning:
image


erwincoumans:
Thanks! LGTM

PR #3950, opened 2021-09-02
Fix: reference frame issue of external force command processor (Physics Server)

bttner:
**Problem Description:**
The _applyExternalForce/Torque_ API in PyBullet offers the option to select either the world (WORLD_FRAME) or the inertial frame (LINK_FRAME) as the reference frame. But only in the case of a multibody (useMaximalCoordinates = False) and only when force is applied to the body, the respective command processor function (_processApplyExternalForceCommand_) works as expected, in all other cases the actions are flipped or do not distinguish between different reference frames (e.g., Multibody -> _applyExternalTorque_ to a link not base). E.g., `p.applyExternalTorque(body, -1, [0, 0, 100], flags=p.LINK_FRAME)` is actually `                p.applyExternalTorque(body, -1, [0, 0, 100], flags=p.WORLD_FRAME)` and vice versa.

To visualize the problem I added a Python script to the example folder called _draw_frames.py_, which allows drawing all frames associated with a link (or base), i.e., the link, inertial, collision, and visual frame. Since the option _useMaximalCoordinates_ in _createMultiBody_ seems experimental*, the mentioned script offers the creation of a single cube or two cubes connected by a fixed joint to test different scenarios.

The overall goal of this PR is that _applyExternalForce_ and _applyExternalTorque_ work the same no matter if the body is a rigid body, multibody or soft body, and that the reference frame is actually the one specified.

***(Slightly offtopic) Problems I encountered when _useMaximalCoordinates = True_:**
- When a link is added to the base, _getLinkState_ returns _None_ for all links (not base).
- When no link is added to the base, _getCollisionShapeData_ fails (pybullet.error: Error receiving collision shape info).
- As depicted below, the return values of _getBasePositionAndOrientation_, _getVisualShapeData_, and _getCollisionShapeData_ appear to be expressed in a different reference frame than specified in the documentation.

Multibody (useMaximalCoordinates = False) |  Rigidbody (useMaximalCoordinates = True)
| ------------- | ------------- |
 |

### DIFF
--- a/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
+++ b/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
@@ -11821,7 +11821,6 @@ bool PhysicsServerCommandProcessor::processApplyExternalForceCommand(const struc
 				else
 				{
 					int link = clientCmd.m_externalForceArguments.m_linkIds[i];
-
 					btVector3 forceWorld = isLinkFrame ? mb->getLink(link).m_cachedWorldTransform.getBasis() * tmpForce : tmpForce;
 					btVector3 relPosWorld = isLinkFrame ? mb->getLink(link).m_cachedWorldTransform.getBasis() * tmpPosition : tmpPosition - mb->getLink(link).m_cachedWorldTransform.getOrigin();
 					mb->addLinkForce(link, forceWorld);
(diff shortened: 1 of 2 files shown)


erwincoumans:
Thanks a lot for looking into this, it is indeed sloppy that this API is not working consistently (we don't use that API in our work much). Also, maximal coordinates doesn't implement various APIs indeed. Will check the PR, hope it doesn't break too much.

erwincoumans:
Thanks for the contribution!

PR #3979, opened 2021-09-28
Add deformable contact forces info

johbrust:
- add contact information for collisions between rigid and soft bodies
- collisions between different soft bodies are not supported
- uses impulse acting on tetrahedral nodes for calculation of forces
- contact points are approximated by node positions
- multiple forces acting on identical nodes are accumulated

### DIFF
--- a/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
+++ b/examples/SharedMemory/PhysicsServerCommandProcessor.cpp
@@ -8078,74 +8078,74 @@ bool PhysicsServerCommandProcessor::processRequestDeformableContactpointHelper(c
 			btSoftBody::Node* node = contact->m_face->m_n[contactNodeIdx];
 			// check if node is already in the list
 			int idx = nodesInContact.findLinearSearch2(node);
+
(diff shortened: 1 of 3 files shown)


erwincoumans:
The pull request makes our internal tests fail, so reverting.

line 8077 int contactNodeIdx = contact->m_bary.maxAxis(); m_bary has weights not node index. line 8152 reversed the sign of distance.
That offset is distance calculated directly from collision detection algorithms http://google3/third_party/bullet/src/BulletSoftBody/btSoftBody.cpp;l=546;rcl=429192192

johbrust:
Hi,
regarding `int contactNodeIdx = contact->m_bary.maxAxis();`: This line is getting the index of the maximum weight (which corresponds to the wanted node) through `.maxAxis()` and not the weight itself.
Regarding the reversed distance sign: I checked it multiple times to be correct by letting a deformable body drop in simulation and checking the distance over time. If not inverted, the distance starts with negative values and approaches zero afterwards.
(I can't open the link provided in your comment.)

PR #4016, opened 2021-10-28
Fix bug in server handling of inverse dynamics

stephentu:
The current implementation of processInverseDynamicsCommand has two issues:

1. For floating bases, the handler is not properly copying over the position and orientation from the client inputs.
2. For floating bases, the handler is ignoring the position of the joint angles from the client inputs. 

This CL fixes both issues. While the fix for (2) is straightforward, I am not sure whether or not the right pybullet API convention is to have (x, y, z, orn) passed in, or instead (orn, x, y, z). The former makes more sense to me and is what I implemented.


erwincoumans:
Ah, thanks for the pull request (didn't get the notification, working some other projects, including Tiny Differentiable Simulator etc). Will check it out.

PR #4024, opened 2021-11-03
EGL device selection by environment variable

BlGene:
This PR enables selection of the EGL devices via an environment variable named `EGL_VISIBLE_DEVICES`. The functionality is designed to be similar to `CUDA_VISIBLE_DEVICES`. On desktop machines the default device settings usually pick the right EGL device, however on server machines with multiple graphics cards this will not always be the case. This can be especially problematic when different people are sharing a machine, which is the reason why I wrote the code.

It should also be noted that the EGL device id is not always the same as the CUDA device id, a script showing how the two can be matched can be found here: https://github.com/BlGene/egl_check under `egl_cuda.py.` It also includes the `list_egl_options.py` script to show you what the different EGL device options are.

### DIFF
--- a/examples/OpenGLWindow/EGLOpenGLWindow.cpp
+++ b/examples/OpenGLWindow/EGLOpenGLWindow.cpp
@@ -33,6 +33,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <cstdlib>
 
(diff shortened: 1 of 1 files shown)


erwincoumans:
Nice, thanks for the PR (and good to see you again 👍 )

PR #4046, opened 2021-11-23
Add Aliengo URDF to pybullet_data

priority-kew:


erwincoumans:
Thanks for the PR. The STL meshes are a bit large, and this repo is hitting close to the maximum for pypi uploads. Need to use some mesh simplification to reduce the file size (I can help with that, will check so that the large commit won't get into the git history).

erwincoumans:
Applied here: https://github.com/bulletphysics/bullet3/pull/4048

PR #4048, opened 2021-11-24
Import Aliengo URDF

priority-kew:
from github.com/unitreerobotics/unitree_pybullet. Reduced all meshes by
50% due to concerns about the size of pybullet_data. Also renamed some
links and joints in the URDF to match conventions and added an example
script.
image


erwincoumans:
Thanks! LGTM

PR #4069, opened 2021-12-03
Reduced Deformable Model

jyc-n:
~~**Issues**~~
~~The reduced deformable object has some strange bounces when there is another moving rigid body in the simulation world. When [lines 58-61](https://github.com/jyc-n/bullet3/blob/4634077733992f5d492a91bfcd663574978086e6/examples/RobotSimulator/RobotSimulatorMain.cpp#L58) are commented, i.e., no other object in the simulation world, the bouncing disappear. This issue is only seen in RobotSimulatorMain or PyBullet.~~

~~I tested it with reduced modes disabled (setting [this flag](https://github.com/jyc-n/bullet3/blob/4634077733992f5d492a91bfcd663574978086e6/src/BulletSoftBody/BulletReducedSoftBody/btReducedSoftBody.cpp#L11) to false). For the reduced cube only case, after the initial collision with the ground, the cube bounce back, and the residual remains 0 for the next few steps. For the two cubes case, after the initial collision, there are two consecutive steps that have about the same number of iteration and residuals. This might be an indicator that it solved twice.~~

~~**Reproduce this issue**~~
~~Build and run App_RobotSimulatorMain.~~

Update: the bouncing issue has been fixed by setting m_minimumSolverBatchSize=128 in PhysicsServerCommandProcessor.cpp (like the one in Bullet).

Reduced deformable model codes are ready to check in.


build_cmake_pybullet_double.sh
```diff
@@ -5,7 +5,7 @@ if [ -e CMakeCache.txt ]; then
 fi
 mkdir -p build_cmake
 cd build_cmake
-cmake -DBUILD_PYBULLET=ON -DBUILD_PYBULLET_NUMPY=ON -DUSE_DOUBLE_PRECISION=ON -DBT_USE_EGL=ON -DCMAKE_BUILD_TYPE=Release .. || exit 1
+cmake -DBUILD_PYBULLET=ON -DBUILD_PYBULLET_NUMPY=ON -DUSE_DOUBLE_PRECISION=ON -DBT_USE_EGL=ON -DCMAKE_BUILD_TYPE=Release -DPYTHON_EXECUTABLE:FILEPATH=/opt/homebrew/Caskroom/miniforge/base/bin/python3 -DPYTHON_INCLUDE_DIR:PATH=/opt/homebrew/Caskroom/miniforge/base/include/python3.9 -DPYTHON_LIBRARY:FILEPATH=/opt/homebrew/Caskroom/miniforge/base/lib/libpython3.9.dylib .. || exit 1
```
erwincoumans, line 8: please revert this file or exclude it from your PR.

examples/RobotSimulator/HelloBulletRobotics.cpp
```diff
@@ -18,19 +18,21 @@ int main(int argc, char* argv[])
...
+	sim->resetSimulation(RESET_USE_REDUCED_DEFORMABLE_WORLD);
 	sim->setGravity(btVector3(0, 0, -9.8));
 	sim->setNumSolverIterations(100);
 	b3RobotSimulatorSetPhysicsEngineParameters args;
 	sim->getPhysicsEngineParameters(args);
 
 	int planeUid = sim->loadURDF("plane.urdf");
 	printf("planeUid = %d\n", planeUid);
+	
+	int r2d2Uid = sim->loadURDF("reduced_cube/reduced_cube.urdf");
```
erwincoumans, line 31: please revert this file or exclude it from this PR.

src/BulletSoftBody/btDeformableBackwardEulerObjective.cpp
```diff
@@ -136,6 +136,8 @@ void btDeformableBackwardEulerObjective::updateVelocity(const TVStack& dv)
 
 void btDeformableBackwardEulerObjective::applyForce(TVStack& force, bool setZero)
 {
+	static btScalar sim_time = 0;
```
erwincoumans, line 139: is this needed or can be removed?

src/BulletSoftBody/btDeformableBodySolver.cpp
```diff
@@ -401,7 +402,7 @@ void btDeformableBodySolver::predictMotion(btScalar solverdt)
 			}
 		}
 	}
-	m_objective->applyExplicitForce(m_residual);
+	applyExplicitForce();
```
erwincoumans, line 405: are the changes to the original btDeformableBodySolver needed?

build_cmake_pybullet_double.sh
jyc-n: will do

examples/RobotSimulator/HelloBulletRobotics.cpp
jyc-n: will do

src/BulletSoftBody/btDeformableBackwardEulerObjective.cpp
jyc-n: I'll remove this line

src/BulletSoftBody/btDeformableBodySolver.cpp
jyc-n: This came from some debugging code, and it is no longer needed

erwincoumans:
Thanks for the modifications. Let's merge it and fix what is needed (build files for setup.py and premake for example).

erwincoumans:
I tracked down a regression due to this PR.


In PhysicsServerCommandProcessor::processDeformable, right after the line psb->initializeFaceTree();
those lines are added, and break the deformable_ball.py and deformable_torus.py

deformWorld->setImplicit(true);
deformWorld->setLineSearch(false);
deformWorld->setUseProjection(true);

Will remove those lines in a follow up PR.

Issue #4115, opened 2022-01-02
pointCloudFromCameraImage.py doesn't run: forgotten conversion to numpy array


jamesbraza:
From https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/examples/pointCloudFromCameraImage.py#L77

```python
img = p.getCameraImage(imgW, imgH, renderer=p.ER_BULLET_HARDWARE_OPENGL)
rgbBuffer = img[2]
depthBuffer = img[3]
print("rgbBuffer.shape=", rgbBuffer.shape)
print("depthBuffer.shape=", depthBuffer.shape)
```

Both `img[2]` and `img[3]` are 1-D tuples when they come out of `p.getCameraImage`.  Then, a few lines later they are printed with `.shape` as if they're numpy arrays.  This causes this example to crash.

In the similar example file [`getCameraImageTest.py`](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/examples/getCameraImageTest.py#L34) there are conversions to numpy arrays using `np.reshape`.  It seems these conversions were left out of `pointCloudFromCameraImage.py`.

PR #4116, opened 2022-01-02
Fixing pybullet example on getting a point cloud

jamesbraza:
I am a newcomer to `bullet` and will be using it + Python to sample point clouds.

This PR resolves https://github.com/bulletphysics/bullet3/issues/4115.

After acquiring images via `getCameraImage`, sometimes the images are reshaped based on `w x h`, and other times based on `h x w`.  Understanding which to use, and when, is something I've found confusing.  So, I also added comments to two of the Python starter files, providing more information.


erwincoumans:
Thanks for the contributions, looks good to me!

PR #4118, opened 2022-01-02
Fix indentation

dbartolini:


akien-mga (approved):
I was about to PR the same, the fix is correct. Here's an example warning with Emscripten:
```
error: field 'm_normalWorldOnB' will be initialized after field 'm_positionWorldOnB' [-Werror,-Wreorder-ctor]
```

I'd suggest also fixing indentation of `m_lateralFrictionDir{1,2}` while at it which were also added recently with space indentation instead of tabs.

dbartolini:
Updated with indentation fix.

dbartolini:
Rebased with indentation fix only.

erwincoumans:
thanks, LGTM.

PR #4134, opened 2022-01-12
suppress printf in Win32 thread creation/destruction

erwincoumans:
allow to use Jolt GJK/EPA implementation (disabled by default, use dispatchInfo.m_useJoltGjkEpa = true, or pybullet.setInternalSimFlags(flags=32 )) 32=enable, 64=disable
implement culling based on triangle normal, for concave trimesh versus convex shapes (disabled by default, use dispatchInfo.m_concave_trimesh_normal_culling=false, or p.setInternalSimFlags(flags=8)) (8=enable, 16=disable)
and compile with cmake -DUSE_JOLT_GJK_EPA=ON or -DBT_ENABLE_JOLT_GJK_EPA
allow to compile Bullet and PyBullet using clang with SSE on Windows (in single precision mode), for example:
cmake -DBUILD_PYBULLET=ON -DCMAKE_BUILD_TYPE=Release -DPYTHON_INCLUDE_DIR=c:\python38\include -DPYTHON_LIBRARY=c:\python38\libs\python38.lib -DPYTHON_DEBUG_LIBRARY=c:\python38\libs\python38.lib -G "Visual Studio 16 2019" -DBULLET2_MULTITHREADING=ON  -T ClangCL -A x64 ..


bttner:
I worked myself through the implementation of Jolt and provided my findings here: https://github.com/jrouwe/JoltPhysics/discussions/628#discussioncomment-7331656. It might be helpful for someone who wants to get an overview before digging into the code.

PR #4141, opened 2022-01-18
Expose collision shape in RayCast callbacks

duckdoom5:
I'm using a btCompoundShape to combine a bunch of shapes. When I hit a ray I need the specific shapes. Unfortunately, this is currently impossible because btCollisionObject:getShape() doesn't return the individual shape. This can only be retrieved from the btCollisionObjectWrapper or when the result is constructed.

The proposed changes expose the shape directly, allowing me to get the actual shape that was inside of the compound


BluePrintRandom:
:D

duckdoom5:
Actually it seems I'm mistaken. I was looking through godot source and found this code:

`
// Triangle index is an odd name but contains the compound shape ID.
// A shape part of -1 indicates the index is a shape index and not a triangle index.
if (convexResult.m_localShapeInfo && convexResult.m_localShapeInfo->m_shapePart == -1) {
	m_shapeId = convexResult.m_localShapeInfo->m_triangleIndex;
} else {
	m_shapeId = 0;
}
`

With that shape ID you can retrieve the shape for the compound object.

Unfortunatly I can't confirm currently, so I'll come back with an update later

PR #4146, opened 2022-01-25
Add explicit int cast to avoid conversion warning

petershh:
On some platforms (e.g. e2k) size_t is declared as unsigned long, and this triggers spurious
Wconversion warnings in expressions using sizeof operator.
To supress this warning, add explicit cast of sizeof output to int. This is safe, since 
sizeof(btChunk) is very small compared to max value of int.


erwincoumans:
Thanks, looks good to me.

PR #4150, opened 2022-01-28
Delete constraints in CommonRigidBodyMTBase

QiaodongCui:
Delete constraints in CommonRigidBodyMTBase.h to avoid leaking memory.


erwincoumans:
The one who creates memory, should also delete it. I think btMLCPSolver doesn't allocate the memory, so it shouldn't delete it.

PR #4170, opened 2022-02-19
Reconstruct

fuchuyuan:
Add API example to 1) write deformable state(position and velocity to file) 2) read state and reconstruct deformabled object.
Add API to get/set mesh vertex velocity and position.


erwincoumans:
Thanks, LGTM

PR #4174, opened 2022-02-20
Fix incorrect application of disturbance push.

sippeyxp:
* Introduced a new randomize hook, randomize_sub_step()
* Adapt the minitaur_push_randomizer to use new hook for applying force.
* Update MinitaurPushRandomizer so that it is gin configurable.


erwincoumans:
Thanks! I suppose this may break older trained policies? Anyway, those can use a pinned version of pybullet.

PR #4178, opened 2022-02-23
Also adds user data specified in URDF files to rigid bodies.

RanTig:
Previously, only user data specified at the root node of the URDF file was applied to rigid bodies. Now user data specified at the links and visual shapes are also added.


erwincoumans:
Thanks, LGTM.

PR #4187, opened 2022-03-04
fix build with bionic c

asuka-mio:


erwincoumans:
Thanks, looks good to me.

PR #4209, opened 2022-03-16
quick fix

pmp-p:
https://github.com/bulletphysics/bullet3/commit/fb3b306af19aac4d7581702a7676c2452fb9658a introduced dependancy on example folder


erwincoumans:
Thanks for the report, I overlooked that indeed. Need to see how this can be fixed better (copy/paste of code will become a problem sooner or later).

erwincoumans:
resolved here: https://github.com/bulletphysics/bullet3/pull/4241

Thanks a lot for reporting it!

PR #4213, opened 2022-03-20
UWP: Fix potentially uninitialized local pointer variable error

bwrsandman:
When compiling uwp such as with vcpkg you get the following error:
```
     8>D:\buildtrees\bullet3\src\3.17-e90f5ef08b.clean\src\BulletSoftBody\btSoftBody.cpp(3933): error C4703: potentially uninitialized local pointer variable 'deltaV' used [D:\buildtrees\bullet3\arm-uwp-dbg\src\BulletSoftBody\BulletSoftBody.vcxproj]
```

Seems like the compiler is not resolving that both cases of the following are identical:
```
if (cti.m_colObj->getInternalType() == btCollisionObject::CO_FEATHERSTONE_LINK)
{
    if (multibodyLinkCol)
    {
    }
}
```


erwincoumans:
Thanks, LGTM!

PR #4217, opened 2022-03-24
Intermittent alignment issues fix

DVLP:
Not sure if `BT_DECLARE_ALIGNED_ALLOCATOR` is omitted at purpose in this file, but adding it like in most aligned classes/structs fixes some intermittent problems with alignment when reassigning transforms with new values. This is happening when building with SIMD support in Emscripten. There's also one `BT_DECLARE_ALIGNED_ALLOCATOR` missing in `btMatrix3x3.h` but with the fix above it doesn't seem to be necessary.

### DIFF
--- a/src/LinearMath/btTransform.h
+++ b/src/LinearMath/btTransform.h
@@ -34,6 +34,7 @@ btTransform
 	btVector3 m_origin;
 
 public:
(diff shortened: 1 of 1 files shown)


erwincoumans:
LGTM, seems reasonable, thanks for the fix!

PR #4228, opened 2022-04-01
Fix uninitialized read in rpyFromMatrix

Teemperor:
The computation of `rpy(1)` tries to take the sin of `rpy(0)` (which is only
initialized a line later).


erwincoumans:
Wow, that is well spotted! Thanks for the fix and report.

PR #4234, opened 2022-04-14
fix missing getTransformB in btMinkowskiSumShape

szabolcsdombi:
- also keep original method with different casing for backward compatibility reasons


erwincoumans:
Thanks, LGTM

szabolcsdombi:
Thank you

PR #4244, opened 2022-04-21
DebugDrawcallback: calculate center only if needed

1vanK:


erwincoumans:
Thanks, LGTM. All bits help!

PR #4248, opened 2022-04-22
Allow delivered classes to access unscaled points

CrosRoad95:
As in title, someone might want to do custom actions on points what currently is not possible.


erwincoumans:
sounds good, LGTM.

PR #3996, opened 2021-10-14
Add pybullet.addUserDebugPoints

wkentaro:
https://user-images.githubusercontent.com/4310419/137382677-11c55c8d-ac3a-443a-a5ca-3bc132396397.mp4

https://user-images.githubusercontent.com/4310419/137382681-2adbe48c-87df-499f-a2a9-49c6b0aa5a31.mp4

PR #4249, opened 2022-04-23
fix for debug lines / point rendering, introduced in PR #3996

erwincoumans:
https://github.com/bulletphysics/bullet3/pull/3996


wkentaro:
Thanks for the fix!

PR #4261, opened 2022-05-14
Example for standalone CMake project

briansoe66:
Build Bullet RobotSimulator example as a standalone CMake project, instead of integrating it into the bullet3 build process. This is useful for simulation projects based on Bullet, that want to use their own independent CMake project setup.


CrezyDud:
can you do one for c api without cmake, i really have no idea how to do it, c api has only 1 example with premake and no docs really..

PR #4264, opened 2022-05-16
Fix `-Wunused-but-set-variable` in debug code

akien-mga:
Another option is to delete this debug code, I can amend the PR if that's what you prefer.

Warning raised by Clang in optimized builds (actually Emscripten which is based on LLVM).


erwincoumans:
Thanks, LGTM.

PR #4283, opened 2022-06-15
fix(ppc64): don't include altivec headers on ppc64

darkbasic:
I've crafted a small patch which I carry in my portage tree since several releases in order to make bullet compile on ppc64le.

I'd like to get it merged.


darkbasic:
Replaced by https://github.com/bulletphysics/bullet3/pull/4371

PR #4284, opened 2022-06-15
Initialize vertexSource

vitalybuka:
Default constructor of btVector3 does not initialize the object.
This fixes Memory Sanitizer reports.


vitalybuka:
Do we need to @erwincoumans somehow?

erwincoumans:
Thanks, looks good to me.

PR #4309, opened 2022-07-22
Initialize 4th component of btVector3

vitalybuka:
It's used in btVector3::setMax


erwincoumans:
thanks, looks good to me.

PR #4314, opened 2022-08-07
Use OpenGL 3.3 core profile in eglCreateContext().

mahiuchun:
- Similar attributes are already used in GLFWOpenGLWindow.cpp.
- This could avoid setting environment variables when using old Mesa where OpenGL compatibility profile is limited to 3.1.


erwincoumans:
Looks good to me, thanks!

PR #4321, opened 2022-08-11
Update btSoftBody to handle close to degenerate triangles.

crewmatt:
In the case of deformable objects, the triangles in the mesh can get very close to degenerate, which can lead to division by zero and NaN propagating through the system. In this case if it's close to degenerate and going to result in NaN, it uses a vertex.


erwincoumans:
Thanks for the improvements, LGTM.

Issue #4331, opened 2022-09-05
pybullet_envs cannot render


markub3327:
Hi,

   I try to render the pyBullet but nothing to do. For more details please look here: https://github.com/openai/gym/issues/3073

# Example 1:
```python
import gym
import pybullet_envs

env = gym.make('HopperBulletEnv-v0')
env.render(mode="human")

env.reset()

for i in range(10000):
    obs, rewards, done, _ = env.step(env.action_space.sample())
```

# Example 1 gives me a error:
> /Users/martin/miniforge3/lib/python3.9/site-packages/gym/envs/registration.py:440: UserWarning: WARN: The `registry.env_specs` property along with `EnvSpecTree` is deprecated. Please use `registry` directly as a dictionary instead.
>   logger.warn(
> pybullet build time: Sep  2 2022 11:52:28
> /Users/martin/miniforge3/lib/python3.9/site-packages/gym/core.py:317: DeprecationWarning: WARN: Initializing wrapper in old step API which returns one bool instead of two. It is recommended to set `new_step_api=True` to use new step API. This will be the default behaviour in future.
>   deprecation(
> /Users/martin/miniforge3/lib/python3.9/site-packages/gym/wrappers/step_api_compatibility.py:39: DeprecationWarning: WARN: Initializing environment in old step API which returns one bool instead of two. It is recommended to set `new_step_api=True` to use new step API. This will be the default behaviour in future.
>   deprecation(
> /Users/martin/miniforge3/lib/python3.9/site-packages/gym/core.py:43: DeprecationWarning: WARN: The argument mode in render method is deprecated; use render_mode during environment initialization instead.
> See here for more information: https://www.gymlibrary.ml/content/api/
>   deprecation(
> Traceback (most recent call last):
>   File "/Users/martin/Documents/Projects/rl-toolkit/test.py", line 5, in <module>
>     env.render(mode="human")
>   File "/Users/martin/miniforge3/lib/python3.9/site-packages/gym/core.py", line 58, in render
>     return render_func(self, *args, **kwargs)
>   File "/Users/martin/miniforge3/lib/python3.9/site-packages/gym/core.py", line 421, in render
>     return self.env.render(*args, **kwargs)
>   File "/Users/martin/miniforge3/lib/python3.9/site-packages/gym/core.py", line 58, in render
>     return render_func(self, *args, **kwargs)
[...]

PR #4335, opened 2022-09-07
compatibility with Gym 0.26.0

markub3327:
Hi @erwincoumans,

   for fixing the issues https://github.com/bulletphysics/bullet3/issues/4331 and https://github.com/openai/gym/issues/3073, I must add `order_enforce=False` flag to all `register()` at __init__.py becouse new version of Gym doesn't allow call `env.render()` before calling `env.reset()`.

Thanks for your time.

### DIFF
--- a/examples/pybullet/examples/rendertest_sync.py
+++ b/examples/pybullet/examples/rendertest_sync.py
@@ -98,7 +98,7 @@ class TestEnv(gym.Env):
                                renderer=p.ER_BULLET_HARDWARE_OPENGL)
     #renderer=pybullet.ER_TINY_RENDERER)
     self._observation = img_arr[2]
-    return np.array(self._observation), 0, 0, {}
+    return np.array(self._observation), 0, 0, False, {}
 
   def seed(self, seed=None):
     pass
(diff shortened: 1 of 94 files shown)


markub3327:
For compatibility with Gym 0.26.0 is needed to change the:
`if id in registry.env_specs:`
To:
`if id in registry.keys():`

Thanks.

markub3327:
For compatibility with Gym 0.26.0 and newer, I must change all `step()` and `reset()`, too.

The new form of `step()` is:
```python
observation, reward, terminated, truncated, info = env.step(action)
```

The new form of `reset()` is:
```python
observation, info = env.reset()
```

The `render` or `renders` argument in `__init__()` in all envs is changed to `render_mode`.

@erwincoumans Please can you do tests of the correctness?

erwincoumans:
Would this change also make PyBullet incompatible with all older versions of Gym? Is there some gym version check possible? Or remove the order enforcing wrapper after make externally?

markub3327:
There are more changes than the wrapper, because Gym 0.26.0 use new form of `env.reset()` and `env.step()`.

It will be incompatible with the older Gym..... But checking the Gym version can solve this problem.

Thanks.

anht-nguyen:
When will these changes for compatibility with Gym 0.26 be merged to the master branch? Thanks!

araffin:
Fyi, I pushed and released on pypi a subset of pybullet envs compatible with gymnasium: https://github.com/araffin/pybullet_envs_gymnasium
`pip install pybullet_envs_gymnasium` (I will crosspost this message as it may interest several people)

Issue #4372, opened 2022-11-15
addUserDebugPoints are not removed even with lifeTime parameter.


ManuCorrea:
Added debug points are not removed, the lifeTime parameter seems to work in debug lines.

I created an example to see the behavior:
````
import pybullet as p
from time import sleep
import pybullet_data

physicsClient = p.connect(p.GUI)

p.setAdditionalSearchPath(pybullet_data.getDataPath())

planeId = p.loadURDF("plane.urdf", [0,0,-1])

x = 0

while p.isConnected():
  p.addUserDebugPoints(
      [[x,1,1]], [[0.1, 0, 0]], pointSize=3, lifeTime=0.5)
  
  p.addUserDebugLine([x, 0, 0], [x, 1, 1], lineColorRGB=[1,0,0], lifeTime=0.5)

  p.stepSimulation()

  x += 0.1

  sleep(0.5)


````

I am using pybullet 3.2.5 on Linux Mint 19.3 Cinnamon.
Python 3.7.13

I attach video and code example.

[pybullet_bug_points.mkv.zip](https://github.com/bulletphysics/bullet3/files/10008220/pybullet_bug_points.mkv.zip)

erwincoumans:
Thanks a lot for the reproduction and pull request, it has been merged.

PR #4382, opened 2022-12-13
Added fix for deleting debug points with life time.

ManuCorrea:
Debug points weren't deleted with having life time.

The issue was shown in the following issue, providing a video and an example.
https://github.com/bulletphysics/bullet3/issues/4372

The problem is solved with this pull request.


erwincoumans:
Thanks for the report, let's merge it.

PR #4413, opened 2023-02-24
re-enable softbody/rigidbody contact report

erwincoumans:


erwincoumans:
Thanks for the contribution, let's enable it again and see if it requires further fixes.

PR #4438, opened 2023-03-30
Fix a bug related to the law of action and reaction in btGeneric6DofSpring2Constraint.

yahei:
The relA and relB in calculateJacobi() are vectors used to calculate the torque that the linearAxis applies to the bodies.
Therefore, relA and relB should point to the same coordinate, but they were pointing to the connection points of bodyA and bodyB respectively.
This caused a bug where the positions where the action and reaction forces were applied were shifted when the relative positions of the bodies changed.

To fix this, relA was corrected to point to the connection point of bodyB.


yahei:
fig

PR #4441, opened 2023-04-04
Fix NaN due to division by zero

zhutq:
I encountered NaN when simulating with a kinematic rigid body, one dynamic rigid body, and one rope anchored between them.
It was solved by avoiding division-by-zero when computing `solverConstraint.m_jacDiagABInv`.


zhutq:
Hello @erwincoumans ! Would you kindly review this PR? If there are any concerns or comments, I would greatly appreciate your input and am happy to address them. Thank you very much for your time and consideration!

SirNate0:
I would also appreciate clarification on this, as I also ran into this NaN when trying to use a fixed point type for btScalar, and I'm not certain what the proper behavior is supposed to be.

whatchamacallem:
try something more like:    solverConstraint.m_jacDiagABInv = btScalar(1.) / (sum + FLT_MIN);

It won't affect normal operation and avoids adding a branch statement to an inner loop.  I'm sure there is more to it when considering different scalar types.

Issue #4451, opened 2023-04-19
Incorrect joint force/torque sensor values for a passive pendulum


azeey:
Might be related to #2966, but no motor torque is applied here.

We've run into an issue with implementing force/torque sensors for Bullet in gz-physics https://github.com/gazebosim/gz-physics/pull/434 where test expectations that pass with DART are failing with Bullet (`btMultiBody`). To narrow down the scope, I've set up a simple example that demonstrates the problem. The example loads a URDF of simple pendulum that passively swings due to gravity starting from an initial angle. It prints the "sensed" force as well as the expected reaction forces. The following is the last printout of the script below showing that the normal (centripetal) forces sensed by bullet are significantly different than the expected value.

```
joint pos: -0.0005009165033698082, vel: -4.62124490737915, accel: -0.0438690185546875
 sensor force: (0.0, 0.004692971706390381, 289.4410705566406) norm: 289.4410705946863
 expected force: [0, -0.2663465830861952, 174.1218738355234] norm: 174.12207754473943
```

However, applying the following patch seems to fix the issue (credit goes to @nkoenig)
```diff
diff --git a/src/BulletDynamics/Featherstone/btMultiBody.cpp b/src/BulletDynamics/Featherstone/btMultiBody.cpp
index d7588aedc..fcadfbb65 100644
--- a/src/BulletDynamics/Featherstone/btMultiBody.cpp
+++ b/src/BulletDynamics/Featherstone/btMultiBody.cpp
@@ -936,7 +936,10 @@ void btMultiBody::computeAccelerationsArticulatedBodyAlgorithmMultiDof(btScalar
 			if (m_useGyroTerm)
 				zeroAccSpatFrc[i + 1].addAngular(spatVel[i + 1].getAngular().cross(m_links[i].m_inertiaLocal * spatVel[i + 1].getAngular()));
 			//
-			zeroAccSpatFrc[i + 1].addLinear(m_links[i].m_mass * spatVel[i + 1].getAngular().cross(spatVel[i + 1].getLinear()));
+			if (!isConstraintPass)
+			{
+				zeroAccSpatFrc[i + 1].addLinear(m_links[i].m_mass * spatVel[i + 1].getAngular().cross(spatVel[i + 1].getLinear()));
+			}
 			//
 			//btVector3 temp = m_links[i].m_mass * spatVel[i+1].getAngular().cross(spatVel[i+1].getLinear());
 			////clamp parent's omega
```

[...]

PR #4462, opened 2023-04-26
Fix joint force/torque sensor values

nkoenig:
This PR resolves https://github.com/bulletphysics/bullet3/issues/4451 by skipping an update to `zeroAccSpatFrc` during the constraint pass.


erwincoumans:

PR #4497, opened 2023-07-13
Deduplicate getIslandId

szabolcsdombi:
Hi Erwin,

Thank you for this repository I enjoy working with it.

I need this fix to be able to build bullet from source as a single file. (this gets me 3x less build time)
With this patch applied I can have a **btBulletAll.cpp**:

```c++
#include "btBulletCollisionAll.cpp"
#include "btBulletDynamicsAll.cpp"
#include "btLinearMathAll.cpp"
```

PS: should I include a **btBulletAll.cpp** in this PR?


szabolcsdombi:
I just found that it breaks my webassembly build.

PR #4536, opened 2023-10-23
cmake: Add BULLET_BUILD_SHARED_LIBS to replace BUILD_SHARED_LIBS

Ryp:
When building with MSVC, bullet doesn't support using shared libs.

Projects might choose to build bullet as part of their CMake pipeline, and there's currently no good way to override BUILD_SHARED_LIBS to static just for bullet, and still support shared library outside of bullet.

This change adds an intermediate BULLET_BUILD_SHARED_LIBS that caches the initial value of BUILD_SHARED_LIBS, but can be freely overriden by parent projects. In my case, I want to be able to force bullet to static on windows, since the original developer didn't support this use case.

All calls to add_library() now have an explicit BULLET_BUILD_TYPE to this purpose. Previous call to add_library() that already were overriding the default build type were left alone. This should be enough to support all current use cases and add the ability for parent projects to force bullet to static.


Ryp:
See also: https://stackoverflow.com/questions/28840628/building-bullet-physics-as-shared-libraries
And https://github.com/Ryp/reaper/commit/5698c5b4b5f7780f481f23560ff45ec95c7bdb27

Ryp:
Gentle ping to have any feedback please

justincredible:
You may be waiting a while for feedback, per #4583 the project is suspended.

PR #4559, opened 2023-12-19
Made raycast vehicle work with collision groups

AndyGura:
A simple non-breakable change, which allows to use collision filters in the vehicle raycaster


AndyGura:
Hey @erwincoumans , anything on this?

PR #4566, opened 2024-01-09
Fix damping constraint index in ConstraintDemo

jad-hamza:
Fix minor index error in `setDamping`, maybe a copy/paste error?


jad-hamza:
I accidentally added a binary file modification that I have now removed from the pull request

PR #4580, opened 2024-02-21
add long double precision support

ZeunO8:
Adds support for long double. Everything seems to work fine and we get the added benefit of long double precision, allowing for larger masses and more refined precision.

The serialization may need to be revisited as I haven't tested that.


erwincoumans:
Wow, quite some work went into this. What kind of application requires this long double precision?

ZeunO8:
Use case lies within simulating masses of universal scale

Issue #2966, opened 2020-08-02
jointReactionForces torque two times the applied MotorTorque


ma-rv:
Hi @erwincoumans,

I would like to obtain joint reaction torques since they vary from the MotorTorque during collisions, but have some troubles understanding the sensor values.

Using `p.TORQUE_CONTROL` with `force=2` results in a torque of 4 in the `jointReactionForces (0.0, -6.8181816792362735, -8.070764462809916e-06, 4.0, 0.0, 0.0)`. Why is the sensor torque two times the motor torque? 

I would be grateful for any advice!

```
import pybullet as p

hinge = p.loadURDF("hinge.urdf")
hingeJointIndex = 0

p.setPhysicsEngineParameter(jointFeedbackMode=p.JOINT_FEEDBACK_IN_JOINT_FRAME)

p.setJointMotorControl2(hinge, hingeJointIndex,p.VELOCITY_CONTROL,0,0,0)
p.setGravity(0,0,0)

p.enableJointForceTorqueSensor(hinge,hingeJointIndex)

for _ in range(100):
    p.setJointMotorControl2(hinge,hingeJointIndex,p.TORQUE_CONTROL,force=10)
    p.stepSimulation()
    print(p.getJointState(hinge, hingeJointIndex)[2])
 
```

hinge.urdf

```
<?xml version="1.0" ?>
<robot name="urdf_robot">
  <link name="baseLink">
    <inertial>
      <origin rpy="0 0 0" xyz="0 0 -0.175"/>
       <mass value="0.0"/>
       <inertia ixx="0" ixy="0" ixz="0" iyy="0" iyz="0" izz="0"/>
    </inertial>
    <visual>
      <origin rpy="0 0 0" xyz="0 0 0.35"/>
      <geometry>
      <box size="0.08 0.16 0.2"/>
      </geometry>
    </visual>
    <collision>
      <origin rpy="0 0 0" xyz="0 0 0.35"/>
      <geometry>
	 <box size="0.08 0.16 0.2"/>
      </geometry>
    </collision>
  </link>
   <link name="childA">
    <inertial>
      <origin rpy="0 0 0" xyz="0 0 0.25"/>
      <mass value="100"/>
      <inertia ixx="0.0" ixy="0" ixz="0" iyy="0.0" iyz="0" izz="0.0"/>
    </inertial>
    <visual>
      <origin rpy="0 0 0" xyz="0 0 0.25"/>
      <geometry>
[...]

PR #4583, opened 2024-02-23
Fix joint force/torque feedback

iche033:
We ran into an [issue](https://github.com/gazebosim/gz-physics/issues/565) downstream in gz-physics (Gazebo) with large force / torque sensor feedback when using bullet featherstone (btMultibody) implementation. The problem is similar to what's described in https://github.com/bulletphysics/bullet3/issues/2966 where user reported larger than expected joint reaction force/torque values when using `TORQUE_CONTROL`. This bullet issue is closed as it seems like the original user switched to using `POSITION_CONTROL`. The patch in this PR is the code posted in https://github.com/bulletphysics/bullet3/issues/2966#issuecomment-688136359 by @Steven89Liu, which fixes the issue for us.

We have a simple test setup that checks the FT sensor output:

bullet_fixed_force_control_feedback

In the above setup, the payload (orange) is connected to the end link with a force/torque sensor attached. There is a joint controller at the revolute joint connecting the base link and the end link.
* Expected value of force: weight of the attached payload link in the sensor frame (`[0 0 -9.8]`).
* Actual value of force: different from above, and magnitude is larger than the weight of the payload link (`[0 -0.34 -13.85]`).

The force output value is correct when run with the physics engine (dartsim) in gz-physics. With this patch, the F/T sensor now also reports the expected value with bullet.

### DIFF
--- a/src/BulletDynamics/Featherstone/btMultiBody.cpp
+++ b/src/BulletDynamics/Featherstone/btMultiBody.cpp
@@ -984,7 +984,11 @@ void btMultiBody::computeAccelerationsArticulatedBodyAlgorithmMultiDof(btScalar 
 			//
 			hDof = spatInertia[i + 1] * m_links[i].m_axes[dof];
 			//
-			Y[m_links[i].m_dofOffset + dof] = m_links[i].m_jointTorque[dof] - m_links[i].m_axes[dof].dot(zeroAccSpatFrc[i + 1]) - spatCoriolisAcc[i].dot(hDof);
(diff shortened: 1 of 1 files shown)


azeey:
Hi @erwincoumans, we were hoping to resolve an issue in Gazebo with this PR. Would you be able to review it?

erwincoumans:
Thanks for bringing this to attention. I've suspended the PyBullet project but let's merge this change.

PR #4619, opened 2024-05-25
replace logging.warn with logging.warning

jelly:
logging.warn is an alias to logging.warning since Python 3.3 and will be removed in Python 3.13.


jelly:
Gentle ping

PR #4741, opened 2025-09-06
Add BUILD_BULLET3_OPENCL option allowing PSP build

jopadan:
For PSP compile use:

```sh
psp-cmake . -DBUILD_CLSOCKET=OFF -DBUILD_BULLET3_OPENCL=OFF
make install 
```


src/CMakeLists.txt
```diff
@@ -1,9 +1,11 @@
 
 IF(BUILD_BULLET3)
-	SUBDIRS(  Bullet3OpenCL Bullet3Serialize/Bullet2FileLoader Bullet3Dynamics Bullet3Collision Bullet3Geometry )
+	SUBDIRS(  Bullet3Serialize/Bullet2FileLoader Bullet3Dynamics Bullet3Collision Bullet3Geometry )
+	IF(BUILD_BULLET3_OPENCL)
+		add_subdirectory ( Bullet3OpenCL )
```
sharkwouter, line 5: The order here might matter, so I would put it above the SUBDIRS call. SUBDIRS is also deprecated since CMake 3.0, so I guess you could also convert the whole list to add_subdirectory, but I don't know the maintainers of this project, so I'm not sure if they would want this.

PR #4756, opened 2025-11-07
Implement demo reset on 'R' key press

rgosens2:
Added functionality to reset the demo when 'R' key is pressed.


rgosens2:
dirty

PR #4760, opened 2025-11-10
Fix useMCLPsolver to useMLCPsolver

rgosens2:
Fixing a small metathesis.


rgosens2:
Why keep using useMCLPsolver, plaese?

rgosens2:
dirty

PR #4761, opened 2025-11-10
change useMCLPSolver to useMLCPSolver

rgosens2:
Fix that metathesis.


metiscus (approved):
Straightforward change. Looks good to me.

PR #4771, opened 2026-02-01
Update CMake minimum version to 3.10

mosfet80:
Fix cmake deprecation 
Cmake version < then 3.10 is deprecated


AnyOldName3:
This will also change CMake policies `CMP0066` to `CMP0071` from old to new, but I don't think any of them are obviously relevant. However, there are a bunch of older policies that Bullet's explicitly setting to new even though they're already new by virtue of the existing `cmake_minimum_required` call, suggesting that whoever did the previous version bump didn't know that they needed to check policies, so it's fairly likely that toggling one of the 65 older policies broke something.

PR #4788, opened 2026-05-05
fix: disable Bullet rayTest stack temp memory on Apple Silicon (arm64)

vlordier:
## Summary

btDbvt::rayTest in btDbvt.h uses a stack-allocated char tempmemory[] buffer which may not be properly aligned for btDbvtNode* pointers on Apple Silicon (arm64).

## Fix

Define BT_DISABLE_STACK_TEMP_MEMORY for Apple platforms in CMakeLists.txt. This forces heap allocation via btAlignedObjectArray::resize(), which guarantees proper alignment.

## Root Cause

On arm64, stack-allocated char arrays have no alignment guarantee. btDbvtNode* requires 8-byte alignment. The misaligned buffer causes undefined behavior during BVH ray traversal.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,10 +54,11 @@ IF(NOT WIN32)
 			# Fix arm64 rayTest stack alignment:
 			# btDbvt::rayTest uses char tempmemory[] on the stack which may
 			# not be properly aligned for btDbvtNode* pointers on arm64.
-			# BT_DISABLE_STACK_TEMP_MEMORY uses heap-allocated
-			# btAlignedObjectArray::resize() instead.
+			# BT_DISABLE_STACK_TEMP_MEMORY disables use of that stack-backed
+			# buffer as btAlignedObjectArray backing storage, forcing
+			# btAlignedObjectArray::resize() to use heap-backed storage.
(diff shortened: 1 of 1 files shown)


vlordier:
@copilot apply changes based on the comments in [this thread](https://github.com/bulletphysics/bullet3/pull/4788#pullrequestreview-4231089999)

vlordier:
**Update**: The root cause of the arm64 hang was identified as an out-of-bounds vector access in ChBody::accumulators (Chrono C++ code), not the Bullet stack alignment. The fix is in Chrono's ChBody.cpp.

However, this PR's BT_DISABLE_STACK_TEMP_MEMORY change is still a valid safety improvement for arm64. Stack-allocated `char tempmemory[]` buffers in `btDbvt::rayTest` have no alignment guarantee on Apple Silicon. While not causing the specific hang we were debugging, it prevents potential future alignment issues.

vlordier:
Closing: the root cause was an OOB vector access in Chrono's ChBody.cpp, not a Bullet stack alignment issue. The accumulator bounds check fix in Chrono resolves the hang without any Bullet changes needed.

PR #4792, opened 2026-06-08
Add function to retrieve all transforms in one batch

scorpion81:
This function is useful in scenarios with larger numbers of rigidbodies ( > 500) in order to retrieve a numpy transforms array in one go.  Querying all transforms in a frame loop in order to update a render representation, e.g. a shattered blender mesh one by one is a significant performance bottleneck. 

Unfortunately, i had to deactivate  2 sections of different numpy related code via #if 0  due to compilation errors i couldnt easily fix. (I didnt change anything else of that code, I added only my C loop function)

If you may find this function useful, feel free to integrate it. Maybe i have overlooked a "Better" way tho, but this improved the performance significantly if you need to use a lot of transformation data.

### DIFF
--- a/examples/pybullet/pybullet.c
+++ b/examples/pybullet/pybullet.c
@@ -4067,6 +4067,82 @@ static PyObject* pybullet_getAABB(PyObject* self, PyObject* args, PyObject* keyw
 	return NULL;
 }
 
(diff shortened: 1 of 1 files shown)


scorpion81:
merged #4787 into my branch, since it properly fixes the numpy related compile errors

Issue #4748, opened 2025-10-15
Inconsistent Structure Return Convention


HanetakaChou:
When I am using Bullet2 on Linux, I have the following bug. The address of the same object in caller and callee is NOT the same.

- Address "convexA" Caller

<img width="1600" height="1007" alt="Image" src="https://github.com/user-attachments/assets/fa8767dc-f278-44b0-b866-c12bb303761c" />

- Address "this" Callee

<img width="1600" height="1007" alt="Image" src="https://github.com/user-attachments/assets/eec0201b-57a7-4a65-8243-6e1ca631899c" />

---

The reason is that, when we return the structure btVector3, the compiler will use "hidden pointer" (-fpcc-struct-return) or "register" (-freg-struct-return) based on whether the macro "BT_USE_SSE_IN_API" is defined.

However, this macro "BT_USE_SSE_IN_API" is NOT consistent between different files.

PR #4799, opened 2026-07-15
Fix inconsistent BT_USE_SSE_IN_API across translation units (#4748)

AniruthSuresh:
# Fix inconsistent `BT_USE_SSE_IN_API` across translation units

This PR addresses the inconsistent use of `BT_USE_SSE_IN_API` across translation units, which can lead to the ABI mismatch described in **#4748**.

## The issue

`btVector3` is returned **by value** from many functions throughout Bullet, including several widely used `virtual` functions.

Whether a compiler returns a struct by value in registers or through a hidden pointer argument depends on `BT_USE_SSE_IN_API`. Because of that, every translation unit that calls or implements these functions must agree on whether the macro is enabled. Otherwise, the caller and callee end up using different ABIs for the same function.

`btScalar.h` is clearly intended to be the single place that decides this macro. In fact, it explicitly documents that it should either be enabled there or through the build system.

However, I found that five `.cpp` files bypass that centralized decision by defining the macro themselves:

- `src/LinearMath/btVector3.cpp`
- `src/BulletCollision/CollisionShapes/btConvexHullShape.cpp`
- `src/BulletCollision/CollisionShapes/btConvexShape.cpp`
- `src/BulletCollision/CollisionShapes/btMultiSphereShape.cpp`
- `src/BulletCollision/CollisionShapes/btPolyhedralConvexShape.cpp`

```cpp
#if defined(_WIN32) || defined(__i386__)
#define BT_USE_SSE_IN_API
#endif
```

Meanwhile, every other translation unit simply includes `btScalar.h` and uses whatever it decides.

This means these five files can disagree with the rest of the project about the calling convention used for `btVector3` return values. If a caller and callee are compiled with different assumptions, they no longer agree on the ABI for the same function. That is the issue reported in **#4748**, where the caller and callee disagree on how a `btVector3` is returned, leading to different addresses being observed for what is logically the same returned object.

## Fix

Remove the five local `BT_USE_SSE_IN_API` definitions.

After this change, the macro is determined exclusively by `btScalar.h` (or by the build system), ensuring that every translation unit uses the same ABI. No behavior changes are intended beyond eliminating the inconsistent macro definition.

This is effectively a no-op for the common configurations:

- **64-bit Linux/Windows:** unchanged (`__i386__` is not defined).
- **macOS x86:** unchanged (`btScalar.h` already enables it).
- **32-bit builds:** projects can still enable `BT_USE_SSE_IN_API` globally, exactly as documented in `btScalar.h`.

## Testing

- [x] Verified that the only remaining definitions of `BT_USE_SSE_IN_API` are the centralized ones in `btScalar.h`.
- [x] Built the project in **Release** mode on x86_64 Linux.
- [x] Ran the `Test_Collision` unit tests (**5/5 passing**).
- [x] Recompiled all 111 source files under `src/LinearMath` and `src/BulletCollision` for 32-bit (`-m32`), both with and without `BT_USE_SSE`, to confirm every affected configuration still builds successfully.

### DIFF
--- a/src/BulletCollision/CollisionShapes/btConvexHullShape.cpp
+++ b/src/BulletCollision/CollisionShapes/btConvexHullShape.cpp
@@ -13,10 +13,6 @@ subject to the following restrictions:
 3. This notice may not be removed or altered from any source distribution.
 */
 
-#if defined(_WIN32) || defined(__i386__)
-#define BT_USE_SSE_IN_API
-#endif
-
 #include "btConvexHullShape.h"
 #include "BulletCollision/CollisionShapes/btCollisionMargin.h"
 
--- a/src/BulletCollision/CollisionShapes/btConvexShape.cpp
+++ b/src/BulletCollision/CollisionShapes/btConvexShape.cpp
@@ -13,10 +13,6 @@ subject to the following restrictions:
 3. This notice may not be removed or altered from any source distribution.
 */
 
-#if defined(_WIN32) || defined(__i386__)
-#define BT_USE_SSE_IN_API
-#endif
-
 #include "btConvexShape.h"
 #include "btTriangleShape.h"
 #include "btSphereShape.h"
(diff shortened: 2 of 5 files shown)


AniruthSuresh:
@erwincoumans Whenever you have some time, could you please take a look? It's a small fix to address the `BT_USE_SSE_IN_API` inconsistency across translation units. If everything looks good, please feel free to merge it. Happy to make any changes if you have feedback. Thanks!

PR #4804, opened 2026-08-24
Fix macOS OpenGL3 detection with CMake 4.1

### LINKED ISSUE #4733: Misidentified macOS version during build of OpenGL3 demos
After upgrading to CMake 4.1.0 I noticed that the OpenGL3 demos no longer build as the build system defines `-DNO_OPENGL3`.
This is a result of the `_CURRENT_OSX_VERSION` variable being removed in https://gitlab.kitware.com/cmake/cmake/-/commit/eb8facd39606e9472f9ff6de40d7caf76767d0da

However the CMakeLists file of bullet still relies on it:
```
IF(BUILD_BULLET3)
	 IF(APPLE)
		MESSAGE("Mac OSX Version is ${_CURRENT_OSX_VERSION}")
		IF(_CURRENT_OSX_VERSION  VERSION_LESS 10.9)
```
I think `CMAKE_SYSTEM_VERSION` should be used instead, though this one gives a the kernel version number (which is higher).
There is also `_CMAKE_HOST_VERSION` that replaces the old variable but it's private as well so may not be reliable.
Alternatively, something like
```
    execute_process(
        COMMAND sw_vers -productVersion
        OUTPUT_VARIABLE MACOSX_VERSION
        OUTPUT_STRIP_TRAILING_WHITESPACE
    )
```
could be used.

tandede:
## Problem

CMake 4.1 removed the private `_CURRENT_OSX_VERSION` variable used by Bullet's macOS OpenGL3 check. When the variable is undefined, its empty value compares as earlier than 10.9, so a default modern macOS configuration disables `BUILD_OPENGL3_DEMOS` and adds `NO_OPENGL3`.

This prevents the Example Browser from using OpenGL3 even though the active compiler and SDK support it.

## Approach

The relevant compatibility boundary is the minimum macOS version targeted by the generated binaries, rather than the host's Darwin kernel version or another private CMake implementation variable.

Use the public `CMAKE_OSX_DEPLOYMENT_TARGET` variable and enter the legacy path only when the user explicitly requests a target earlier than 10.9. When no deployment target is specified, leave the option at its normal default and let the selected compiler and SDK determine the effective target.

This also preserves the existing behavior for explicit legacy targets without introducing another dependency on private CMake state.

## Result

With CMake 4.4.2:

- the default empty deployment target keeps `BUILD_OPENGL3_DEMOS=ON` and does not define `NO_OPENGL3`;
- an explicit `10.8` target sets the option to `OFF` and defines `NO_OPENGL3`;
- an explicit `10.9` target keeps the option `ON` and does not define `NO_OPENGL3`.

The pre-fix baseline sets the option to `OFF` for the same default empty-target configuration.

## Validation

- Configured fresh default, macOS 10.8, and macOS 10.9 build trees with CMake 4.4.2.
- Completed the default Clang 21 build: 967/967 build steps passed.
- Confirmed that `examples/ExampleBrowser/App_ExampleBrowser` was linked successfully.
- Ran `git diff --check`.

Fixes #4733.


Snovve:
just download linux

